PR #42, opened 2019-05-16
Add ThreadId to LogEvents.

loyolman:


src/System/Logging/LogEventArgs.cs
```diff
@@ -24,6 +25,7 @@ public LogEventArgs(uint tagid, Category category, Level level, string message,
 			Category = category;
 			Level = level;
 			TagId = tagid;
+			ThreadId = Thread.CurrentThread.ManagedThreadId;
```
davidahs: >CurrentThread [](start = 21, length = 13)

While this should never be null, I'd suggest being defensive (Thread.CurrentThread?.ManagedThreadId)

PR #43, opened 2019-05-21
Add not null attributes for code analysis purposes in validation code

patrikmolsson:
In order to suppress [CA1062](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1062-validate-arguments-of-public-methods?view=vs-2019) when using external validation libraries, e.g. this library, you need to decorate the validated params with a ValidatedNotNullAttribute in order for the compiler to understand that the method in fact checks for null.

- Created the ValidatedNotNullAttribute 
- Added the ValidatedNotNullAttribute to the Validation methods that checks for null

### DIFF
--- a/src/System/Validation/ValidatedNotNullAttribute.cs
+++ b/src/System/Validation/ValidatedNotNullAttribute.cs
@@ -16,17 +16,16 @@ namespace Microsoft.Omex.System.Validation
 	/// </summary>
 	/// <example>
 	/// In this example the Guard.NotNull has the attribute present on its parameter. If it wasn't present the analyzer would not know
-	/// that s cannot be null following the execution of the method, resulting in a warning on the execution of the `.toUpper` method.
+	/// that s cannot be null following the execution of the method, resulting in a warning on the execution of the <c>toUpperInvariant</c> method.
 	/// <code>
 	/// string s = mightBeNullClass?.stringProperty;
 	///
 	/// Guard.NotNull(s);
 	///
 	/// // The analyzer now knows `s` will never be null here
-	/// return s.toUpper();
+	/// return s.toUpperInvariant();
 	/// </code>
 	/// </example>
-	/// 
 	[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
 	internal sealed class ValidatedNotNullAttribute : Attribute { }
 }
--- a/src/System/Validation/ValidatedNotNullAttribute.cs
+++ b/src/System/Validation/ValidatedNotNullAttribute.cs
@@ -6,17 +6,18 @@ using System;
 namespace Microsoft.Omex.System.Validation
 {
 	/// <summary>
-	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null.
+	/// Indicates to code analysis tools that a method validates a particular parameter to not be null.
 	/// 
 	/// This can for instance be used to suppress VS design warning CA1062 (validate public method parameters before use) by applying to
-	/// external validation libraries, such as the `Code` class in this project.
+	/// external validation libraries, such as the <see cref="Code" /> class in this project.
 	/// 
 	/// If an attribute with the name ValidatedNotNull is present on the parameter of the validation method it tells the analyzer that the 
 	/// parameter will never be null in the code path following the execution of that method.
 	/// </summary>
 	/// <example>
-	/// In this example the Guard.NotNull has the attribute present on its parameter. If it wasn't present the analyzer would not know
-	/// that s cannot be null following the execution of the method, resulting in a warning on the execution of the <c>toUpperInvariant</c> method.
+	/// In this example the <c>Guard.NotNull</c> has the attribute present on its parameter. If it wasn't present the analyzer would not know
+	/// that <c>s</c> cannot be <c>null</c> following the execution of the method, resulting in a warning on the execution of the
+	/// <c>toUpperInvariant</c> method.
 	/// <code>
 	/// string s = mightBeNullClass?.stringProperty;
 	///
--- a/src/System/Validation/ValidatedNotNullAttribute.cs
+++ b/src/System/Validation/ValidatedNotNullAttribute.cs
@@ -1,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.Omex.System.Validation
+{
+    /// <summary>
+    /// Indicates to Code analysis tools that a method validates a particular parameter to not be null
+    /// </summary>
+    internal sealed class ValidatedNotNullAttribute : Attribute {}
+}
--- a/src/System/Validation/ValidatedNotNullAttribute.cs
+++ b/src/System/Validation/ValidatedNotNullAttribute.cs
@@ -6,8 +6,27 @@ using System;
 namespace Microsoft.Omex.System.Validation
 {
 	/// <summary>
-	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null
+	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null.
+	/// 
+	/// This can for instance be used to suppress VS design warning CA1062 (validate public method parameters before use) by applying to
+	/// external validation libraries, such as the `Code` class in this project.
+	/// 
+	/// If an attribute with the name ValidatedNotNull is present on the parameter of the validation method it tells the analyzer that the 
+	/// parameter will never be null in the code path following the execution of that method.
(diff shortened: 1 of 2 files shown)


patrikmolsson:
Validated locally with SonarQube by using the following code as a guard:

```cs
using System;

namespace ValidateLib
{
    public class ValidatedNotNullAttribute: Attribute { }
    public static class Guard
    {
        public static void NotNull<T>([ValidatedNotNull] T value, string error) where T : class
        {
            if (value == null)
            {
                throw new ArgumentNullException(error);
            }
        }
    }
}
```

Compiled that to a .dll as a release build and copied said dll into the project. Used the code like so in a method, which correctly suppressed an error that the `Code.ExpectsArgument` didn't catch:


*No error:*
```cs
// Introduce a code path where m2 can be null
Model m2 = aBooleanThatMightBeFalse ? await GetModel() : null;

Guard.NotNull(m2, "m2 is null");

return m2.MyProperty;
```

*Error:*
```cs
// Introduce a code path where m2 can be null
Model m2 = aBooleanThatMightBeFalse ? await GetModel() : null;

Code.ExpectsArgument(m2, "m2 is null", 0);

return m2.MyProperty;
```

muiriswoulfe:
> 		public static T ExpectsArgument<T>(T argumentValue, string argumentName, Func<T, bool> predicate, string errorMessage, uint? tagId)

This should probably be marked as well. The only difficulty here is that T may or may not be a nullable type.

---
Refers to: src/System/Validation/Code.cs:108 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

muiriswoulfe:
> 		public static bool ValidateAllNotNull<T>(IEnumerable<T> argumentValue, string argumentName, uint? tagId)

I believe this should also be marked.

---
Refers to: src/System/Validation/Code.cs:210 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

muiriswoulfe:
> 		public static bool ValidateAny<T>(IEnumerable<T> argumentValue, string argumentName, uint? tagId)

Should also mark this.

---
Refers to: src/System/Validation/Code.cs:241 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

muiriswoulfe:
> 		public static bool ValidateArgument(object argumentValue, string argumentName, uint? tagId)

And this.

---
Refers to: src/System/Validation/Code.cs:270 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

muiriswoulfe:
> 		public static bool ValidateArgument<T>(T argumentValue, string argumentName, Func<T, bool> predicate, string errorMessage, uint? tagId)

And also this.

---
Refers to: src/System/Validation/Code.cs:297 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

muiriswoulfe:
> 		public static bool ValidateNotNullOrWhiteSpaceArgument(string argumentValue, string argumentName, uint? tagId = null)

And finally this.

---
Refers to: src/System/Validation/Code.cs:345 in d5f4e8f. [](commit_id = d5f4e8f34ea0c064183f3dffb36f80e1ffa64af9, deletion_comment = False)

src/System/Validation/ValidatedNotNullAttribute.cs
```diff
@@ -0,0 +1,13 @@
...
+
+using System;
+
+namespace Microsoft.Omex.System.Validation
+{
+	/// <summary>
+	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null
+	/// </summary>
+	[AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
+	internal sealed class ValidatedNotNullAttribute : Attribute { }
```
muiriswoulfe: >ValidatedNotNullAttribute [](start = 23, length = 25)

I understand the rationale for this, but the naming doesn't seem quite right. The elements marked with this are potentially null and you're validating that they're not null, whereas the naming here indicates that it has been validated as not being null. Can you use an alternative name?

src/System/Validation/ValidatedNotNullAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.Omex.System.Validation
+{
+	/// <summary>
+	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null
```
muiriswoulfe: There is some useful information in your PR description about the cases where this is beneficial. I'd encourage you to add those details into this comment so that they aren't lost over time.

patrikmolsson:
@muiriswoulfe I might have been too vague on my explanation here. What that attribute tells the code analyser is that "the method will validate that the parameter marked with the attribute will not be null after this method has been executed", i.e. no code path exists after the execution of the method where the parameter is null.

Only some of the `Code.Expects...` methods have that behaviour (the Code.ExpectsArgument with the predicate does not actually check the param for null, it's up to the predicate). None of the `Code.Validate...` methods do, they merely log and return true/false depending on whether the param is null or not.

That is why I have chosen to only decorate some of the methods.

src/System/Validation/ValidatedNotNullAttribute.cs
patrikmolsson: I will look into this!

src/System/Validation/ValidatedNotNullAttribute.cs
patrikmolsson: The naming is unfortunately nothing I can change. That is what the analysers look at; whether an attribute with the name `ValidatedNotNull` is present on the parameter.

src/System/Validation/ValidatedNotNullAttribute.cs
muiriswoulfe: Thanks for adding the details!

src/System/Validation/ValidatedNotNullAttribute.cs
muiriswoulfe: It might be worth briefly mentioning this in the comment as well.

src/System/Validation/ValidatedNotNullAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+
+namespace Microsoft.Omex.System.Validation
+{
+	/// <summary>
+	/// Indicates to Code analysis tools that a method validates a particular parameter to not be null.
+	/// 
+	/// This can for instance be used to suppress VS design warning CA1062 (validate public method parameters before use) by applying to
+	/// external validation libraries, such as the `Code` class in this project.
```
muiriswoulfe: >`Code` [](start = 48, length = 6)

Very small issue: it's preferable to include this as <see cref="Code"/> as the outputted documentation will be better formatted. Same for a few instances below. Where cref isn't appropriate, you can use <c>.toUpper</c> for instance.

src/System/Validation/ValidatedNotNullAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+	/// <example>
+	/// In this example the Guard.NotNull has the attribute present on its parameter. If it wasn't present the analyzer would not know
+	/// that s cannot be null following the execution of the method, resulting in a warning on the execution of the `.toUpper` method.
+	/// <code>
+	/// string s = mightBeNullClass?.stringProperty;
+	///
+	/// Guard.NotNull(s);
+	///
+	/// // The analyzer now knows `s` will never be null here
+	/// return s.toUpper();
```
muiriswoulfe: >toUpper [](start = 14, length = 7)

Very small issue: ToUpperInvariant() is preferable (which SonarJS enforces as well). I'd suggest updating the example here to avoid people copying non-ideal code.

muiriswoulfe:
Thanks for clarifying. That makes sense.

---
In reply to: [494549755](https://github.com/microsoft/Omex/pull/43#issuecomment-494549755) [](ancestors = 494549755)

src/System/Validation/ValidatedNotNullAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+	/// </code>
+	/// </example>
+	///
```
muiriswoulfe: Nitpick: Trailing line in comment, and some trailing whitespace in this comment :smile:

PR #49, opened 2019-07-22
Removed Newtonsoft.Json dependency

knstntn:
Removed Newtonsoft.Json dependency


knstntn:
Discussed in person. I do not agree with arguments I got. 

My position on this PR is: it should be merged - I think that for publicly released products we should narrow down the scope of the code
- if we do not use Newtonsoft.Json than we should remove it from dependencies.
- if we do not use some code than we should remove it. Each code removal introduces the chance of a breaking change, so we should be extra careful about introducing code which later we will need to maintain and support.

doguva-msft:
> Discussed in person. I do not agree with arguments I got.
> 
> My position on this PR is: it should be merged - I think that for publicly released products we should narrow down the scope of the code
> 
> * if we do not use Newtonsoft.Json than we should remove it from dependencies.
> * if we do not use some code than we should remove it. Each code removal introduces the chance of a breaking change, so we should be extra careful about introducing code which later we will need to maintain and support.

doguva-msft:
> > * if we do not use Newtonsoft.Json than we should remove it from dependencies.
Some Document Db research may help here. Document Db client .NET SDK which is this adapter layer built upon, already has a dependency on Newtonsoft.Json. As Document DB is a json document store. What you are proposing is on the same level as you, making a delete request to all methods in Document Db public client SDK for the methods that "your team" is not using :)

Issue #46, opened 2019-06-26
Gating configuration not taken into account


dakanev-msft:
Specified resource file names are ignored, and the default names OmexTip.xml and OmexGates.xml are always used instead. This manifests as gates being loaded as disabled (due to a data set load error) and the gating is effectively not working.

The workaround for this issue is to rename the resource files to OmexTip.xml and OmexGates.xml.

dakanev-msft:
This is worked on as part of https://github.com/microsoft/Omex/pull/66

dakanev-msft:
Closed in commit 66a70c78206cd094717f8fe19fff214da7cac563

dakanev-msft:
The fix caused unit tests in other components to fail. The corresponding commit will be reverted.

K-Cully:
Gating is no longer in this repo, or supported.

PR #66, opened 2019-07-24
Fixing "Gating configuration not taken into account #46" issue

dakanev-msft:
Added a data set factory to fix the issue with gating configuration not taken into account.


src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
```diff
@@ -347,7 +366,13 @@ private TDataSet LoadDataSet(ResourceUpdatedEventArgs arguments)
...
 					"Loading data set for '{0}'.", typeof(TDataSet).Name);
 
-				TDataSet dataSet = m_dataSetDefault ?? new TDataSet();
+				TDataSet dataSet = m_dataSetDefault ?? m_dataSetFactory.Create();
+
+				if (dataSet == null)
+				{
+					ULSLogging.LogTraceTag(0, Categories.ConfigurationDataSet, Levels.Error,
+						"Unable to create data set of type '{0}'", typeof(TDataSet).Name);
+				}
```
dakanev-msft: >} [](start = 4, length = 1)

This should return

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
```diff
@@ -275,19 +283,30 @@ private CachedConfigurationDataSet<T> DataSets
...
+			/// </summary>
+			private readonly IDataSetFactory<TDataSet> m_dataSetFactory;
+
+
 			/// <summary>
 			/// Initializes a new instance of the <see cref="CachedConfigurationDataSet{TDataSet}" /> class.
 			/// </summary>
 			/// <param name="dataSetDefault">The data set default.</param>
-			public CachedConfigurationDataSet(TDataSet dataSetDefault) => m_dataSetDefault = dataSetDefault;
+			/// /// <param name="dataSetFactory">Data set factory.</param>
```
vlad-ion:
```suggestion
			/// <param name="dataSetFactory">Data set factory.</param>
```

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
```diff
@@ -275,19 +283,30 @@ private CachedConfigurationDataSet<T> DataSets
...
 			/// <summary>
 			/// Initializes a new instance of the <see cref="CachedConfigurationDataSet{TDataSet}" /> class.
 			/// </summary>
 			/// <param name="dataSetDefault">The data set default.</param>
-			public CachedConfigurationDataSet(TDataSet dataSetDefault) => m_dataSetDefault = dataSetDefault;
+			/// <param name="dataSetFactory">Data set factory.</param>
+			public CachedConfigurationDataSet(TDataSet dataSetDefault, IDataSetFactory<TDataSet> dataSetFactory)
+			{
+				m_dataSetDefault = dataSetDefault;
+				m_dataSetFactory = Code.ExpectsArgument(dataSetFactory, nameof(dataSetFactory), TaggingUtilities.ReserveTag(0));
```
aszczepanski: Would it make sense to make dataSetFactory optional here too? Especially if m_dataSetDefault is set

src/System/Configuration/DataSets/IDataSetFactory.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.Omex.System.Configuration.DataSets
+{
+	/// <summary>
+	/// Data set factory interface
+	/// </summary>
+	/// <typeparam name="T">Type of the data set to create</typeparam>
+	public interface IDataSetFactory<T>
```
aszczepanski: It might be worth adding `where T : class, IConfigurationDataSet`

src/Gating/GateDataSetFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+	public class GateDataSetFactory : IDataSetFactory<GateDataSet>
+	{
+		/// <summary>
+		/// Constructor
+		/// </summary>
+		/// <param name="gatesResourceName">The resource name for loading gates</param>
+		/// <param name="testGroupsResourceName">The resource name for loading test groups</param>
+		public GateDataSetFactory(string gatesResourceName, string testGroupsResourceName)
+		{
+			GatesResourceName = Code.ExpectsNotNullOrWhiteSpaceArgument(gatesResourceName, nameof(gatesResourceName), TaggingUtilities.ReserveTag(0));
```
aszczepanski: Might be worth adding unit tests

src/Gating/GateDataSetLoader.cs
```diff
@@ -22,7 +22,7 @@ public class GateDataSetLoader : ConfigurationDataSetLoader<GateDataSet>
 		/// <param name="gates">The gates.</param>
 		/// <param name="testGroups">The test groups.</param>
 		public GateDataSetLoader(ICache cache, IResourceMonitor resourceMonitor, IResource gates, IResource testGroups)
-			: base(cache, resourceMonitor)
+			: base(cache, resourceMonitor, new GateDataSetFactory(gates.Name, testGroups.Name))
```
K-Cully: suggest using gates?.Name and testGroups?.Name to allow the Code.Expects in GateDataSetFactory to log the missing values instead of throwing here.

src/System/Configuration/DataSets/IDataSetFactory.cs
K-Cully: Can T be restricted to IConfigurationDataSet?

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
```diff
@@ -44,11 +50,13 @@ public abstract class ConfigurationDataSetLoader<T> : ConfigurationDataSetLoader
...
 		/// <param name="cache">The cache object.</param>
 		/// <param name="resourceMonitor">Resource Watcher instance</param>
+		/// <param name="dataSetFactory">The data set factory.</param>
 		/// <param name="dataSetOverride">The data set override.</param>
-		protected ConfigurationDataSetLoader(ICache cache, IResourceMonitor resourceMonitor, T dataSetOverride = null)
+		protected ConfigurationDataSetLoader(ICache cache, IResourceMonitor resourceMonitor, IDataSetFactory<T> dataSetFactory, T dataSetOverride = null)
 		{
 			Cache = Code.ExpectsArgument(cache, nameof(cache), TaggingUtilities.ReserveTag(0x238208d9 /* tag_9669z */));
 			m_resourceMonitor = Code.ExpectsArgument(resourceMonitor, nameof(resourceMonitor), TaggingUtilities.ReserveTag(0x23821000 /* tag_967aa */));
+			DataSetFactory = Code.ExpectsArgument(dataSetFactory, nameof(dataSetFactory), TaggingUtilities.ReserveTag(0));
```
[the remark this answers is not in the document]
dakanev-msft: That shouldn't be possible because of the NuGet dependencies, or am I missing something?

---
In reply to: [306890022](https://github.com/microsoft/Omex/pull/66#discussion_r306890022) [](ancestors = 306890022)

src/Gating/GateDataSet.cs
```diff
@@ -1856,13 +1856,13 @@ private TestGroupsDataSet<T> LoadTestGroupsDataSet(IDictionary<string, IResource
 		/// <summary>
 		/// The default gates resource name.
 		/// </summary>
-		public const string DefaultGatesResourceName = "OmexGates.xml";
+		public const string DefaultGatesResourceName = "DefaultGates.xml";
```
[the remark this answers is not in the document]
dakanev-msft: Don't we have the file names configured or we rely on the default ones? 

---
In reply to: [306890976](https://github.com/microsoft/Omex/pull/66#discussion_r306890976) [](ancestors = 306890976)

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
dakanev-msft: The risk of the optional parameter is that it could hide potential errors in the future, in this case default values would be used and in most cases the gates won't loaded. 

I even think that as a next step we should remove the default values of OmexTip.xml and OmexGates.xml, I think it is better that they are specified explicitly.

What do you think?

---
In reply to: [306892478](https://github.com/microsoft/Omex/pull/66#discussion_r306892478) [](ancestors = 306892478)

src/Gating/GateDataSet.cs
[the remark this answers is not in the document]
dakanev-msft: Discussed offline. The names are not configured everywhere in the code, for backward compatibility these two should be reverted to the previous defaults.

---
In reply to: [307201214](https://github.com/microsoft/Omex/pull/66#discussion_r307201214) [](ancestors = 307201214,306890976)

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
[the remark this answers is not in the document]
dakanev-msft: Discussed offline, the NuGet package dependencies will handle this.

---
In reply to: [307200376](https://github.com/microsoft/Omex/pull/66#discussion_r307200376) [](ancestors = 307200376,306890022)

src/System/Configuration/DataSets/ConfigurationDataSetLoader.cs
dakanev-msft: Discussed offline. We will keep it like that from maintainability perspective.

---
In reply to: [307202321](https://github.com/microsoft/Omex/pull/66#discussion_r307202321) [](ancestors = 307202321,306892478)

PR #91, opened 2019-08-13
Modify levels enum order

mastac-msft:


src/System/Logging/Levels.cs
```diff
@@ -43,16 +43,16 @@ public static class Levels
...
-            Warning,
-            /// <summary> The Info log level </summary>
-            Info,
-            /// <summary> The Verbose log level </summary>
-            Verbose,
-            /// <summary> The Spam log level </summary>
-            Spam,
+			/// <summary> The Error log level </summary>
+			Error = 0,
+			/// <summary> The Warning log level </summary>
```
dakanev-msft: >Warning [](start = 21, length = 7)

comments do not match

PR #96, opened 2019-08-15
Add timed scopes stack manager.

mastac-msft:


src/System/TimedScopes/ITimedScopeStackManager.cs
```diff
@@ -0,0 +1,24 @@
...
+	/// <summary>
+	/// Interface for handling active scopes stack
+	/// </summary>
+	public interface ITimedScopeStackManager
+	{
+		/// <summary>
+		/// Get a stack of active scopes, creating a new stack if one does not exist
+		/// </summary>
+		/// <returns>stack of scopes</returns>
+		TimedScopeStack GetTimedScopeStack();
```
dakanev-msft: >GetTimedScopeStack [](start = 18, length = 18)

Can this be replaced with 





public interface ITimedScopeStackManager
 {
  
  TimedScopeStack {get; set; };
}

?

src/System/TimedScopes/TimedScope.cs
```diff
@@ -231,7 +234,7 @@ public void Start()
 				}
 			}
 
-			Parent = TimedScope.Current;
+			Parent = TimedScopeStackManager.GetTimedScopeStack()?.Peek();
```
dakanev-msft: >TimedScopeStackManager [](start = 12, length = 22)

This can be null, right?

src/System/TimedScopes/TimedScope.cs
```diff
@@ -28,20 +27,23 @@ public class TimedScope : IDisposable
...
-		private TimedScope(TimedScopeDefinition scopeDefinition, CorrelationData correlationData, ITimedScopeLogger scopeLogger, 
-			IReplayEventConfigurator replayEventConfigurator, IMachineInformation machineInformation)
+		private TimedScope(TimedScopeDefinition scopeDefinition, CorrelationData correlationData, ITimedScopeLogger scopeLogger, IReplayEventConfigurator replayEventConfigurator,
+			IMachineInformation machineInformation, ITimedScopeStackManager timedScopeStackManager)
 		{
 
 			Code.ExpectsArgument(scopeDefinition, nameof(scopeDefinition), TaggingUtilities.ReserveTag(0x238174da /* tag_96xt0 */));
 			Code.ExpectsArgument(scopeLogger, nameof(scopeLogger), TaggingUtilities.ReserveTag(0x238174db /* tag_96xt1 */));
 			Code.ExpectsArgument(replayEventConfigurator, nameof(replayEventConfigurator), TaggingUtilities.ReserveTag(0x238174dc /* tag_96xt2 */));
+			Code.ExpectsArgument(timedScopeStackManager, nameof(timedScopeStackManager), TaggingUtilities.ReserveTag(0));
```
dakanev-msft: >timedScopeStackManager [](start = 55, length = 22)

The timed scope manager is null by default, wouldn't that fail the constructor here?

src/System/TimedScopes/TimedScopeStack.cs
```diff
@@ -0,0 +1,89 @@
...
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.Omex.System.Logging;
+using Microsoft.Omex.System.Validation;
+
+namespace Microsoft.Omex.System.TimedScopes
+{
+	/// <summary>
+	/// LinkedStack
```
dakanev-msft: >LinkedStack [](start = 5, length = 11)

Please elaborate more what the purpose of the class is

src/System/TimedScopes/TimedScopeStack.cs
```diff
@@ -0,0 +1,89 @@
...
+	/// </summary>
+	[Serializable]
+	public class TimedScopeStack
+	{
+		/// <summary>
+		/// Root item for all stacks
+		/// </summary>
+		public static TimedScopeStack Root { get; } = new TimedScopeStack();
```
dakanev-msft: >static [](start = 9, length = 6)

Is it intended to be static?

src/System/TimedScopes/TimedScopeStack.cs
```diff
@@ -0,0 +1,89 @@
...
+
+
+		/// <summary>
+		/// Root node
```
dakanev-msft: >/// Root node [](start = 2, length = 13)

The comment does not correspond to the property.

Shouldn't it be ReferenceEquals(this,Root)?

src/System/TimedScopes/TimedScope.cs
mastac-msft: I kept the old logic in constructor; scopeLoger which was the original parameter also is by default null and still is checked; the same I applied to the other parameters (however it does not really make much sense to me).

src/System/TimedScopes/ITimedScopeStackManager.cs
mastac-msft: I can change this, however, if I am not mistaken it was @AndreyTretyak suggestion to have separate methods instead of property.

src/System/TimedScopes/TimedScopeStack.cs
mastac-msft: I am not sure, it was done like this in OmexSystem code, Root method returns new TimedScopeStack().

src/System/TimedScopes/TimedScopeStack.cs
mastac-msft: yes, it is used in TimedScopeStackManager.GetTimedScopeStack() method.

src/System/TimedScopes/ITimedScopeStackManager.cs
dakanev-msft: Can you please ask him what the reason for that was?

---
In reply to: [314391362](https://github.com/microsoft/Omex/pull/96#discussion_r314391362) [](ancestors = 314391362)

src/System/TimedScopes/TimedScope.cs
dakanev-msft: Can you please make sure it works as expected.

---
In reply to: [314390077](https://github.com/microsoft/Omex/pull/96#discussion_r314390077) [](ancestors = 314390077)

src/System/TimedScopes/ITimedScopeStackManager.cs
mastac-msft: Replaced

src/System/TimedScopes/TimedScope.cs
mastac-msft: As the timed scope manager is obligatory for times scope to work, I made it an obligatory parameter. The same applies to other parameters: scopeLogger, replayEventConfigurator.

src/System/TimedScopes/TimedScope.cs
```diff
@@ -231,7 +268,7 @@ public void Start()
 				}
 			}
 
-			Parent = TimedScope.Current;
+			Parent = TimedScopeStackManager.Scopes?.Peek();
```
dakanev-msft: >imedScopeStackManager.Scopes? [](start = 13, length = 29)

If TimedScopestackManager is null,

TimedScopeStackManager.Scopes will throw

PR #115, opened 2019-10-09
Change pipeline to use netcore 3.0

AndreyTretyak:


azure-pipelines.yml
```diff
@@ -12,21 +12,21 @@ pool:
...
 
 steps:
-- task: NuGetToolInstaller@0
-  displayName: 'Use NuGet 4.7.1'
+- task: UseDotNet@2
+  displayName: 'Use .NET Core sdk 2.2.104'
   inputs:
-    versionSpec: 4.7.1
-  enabled: true
+    version: 2.2.104
```
aszczepanski: Could you use 2.2.x?

azure-pipelines.yml
```diff
@@ -12,21 +12,21 @@ pool:
...
 
-- task: NuGetCommand@2
-  displayName: 'NuGet restore'
-  enabled: true
+- task: UseDotNet@2
+  displayName: 'Use .NET Core sdk 3.0.100'
+  inputs:
+    version: 3.0.100
```
aszczepanski: Could you use 3.0.x?

azure-pipelines.yml
AndreyTretyak: Thanks, changed in both cases

azure-pipelines.yml
AndreyTretyak: Fixed, thanks

PR #121, opened 2019-10-13
Add new validation and expects method: ValidateAnyAndAllNotNull

dan-hanlon:
We have methods to validate an enumerable is not empty, and another to validate that the enumerable does not contain nulls, but we don't have one to validate that it is both not empty AND does not contain nulls. I've added proposed methods to cover this validation case.

I've also added some unit test for cases I noticed weren't covered.


src/System/Validation/Code.cs
```diff
@@ -198,6 +221,38 @@ public static bool Validate(bool state, uint? tagId, string errorMessage, params
...
+		/// <returns>True if the argument <paramref name="argumentValue"/> is not null, is not empty, and does not contain nulls; false otherwise.</returns>
+		public static bool ValidateAnyAndAllNotNull<T>(IEnumerable<T> argumentValue, string argumentName, uint? tagId) 
+			where T : class
+		{
+			if (!ValidateArgument(argumentValue, argumentName, tagId))
+			{
+				return false;
+			}
+
+			if (!argumentValue.Any() || argumentValue.Any(x => x == null))
```
AndreyTretyak: I don't like double iteration over `IEnumerable` here, strictly speaking, we should've used `ICollection<T>` as a parameter type here, but that that would require changes in downstream repositories.

What do you think about replacing it with good old `foreach` to avoid double iteration?

dan-hanlon: Agree the double iteration isn't good. For this method we could use an `ICollection<T>` as there won't be any dependencies on it yet? Or if we want to keep it consistent using `IEnumerable<T>` I'll replace it with a `foreach`.

AndreyTretyak: I would probably go with `ICollection<T>` since after calling this validation method we would probably need iterate over a collection again to get values.

src/System/Validation/Code.cs
```diff
@@ -55,6 +55,29 @@ public static IEnumerable<T> ExpectsAllNotNull<T>([ValidatedNotNull] IEnumerable
 		}
 
 
+		/// <summary>
+		/// Checks the enumeable argument and throws an exception if it is null, contains no values, or contains any null values
```
vlad-ion: typo: enumerable

src/System/Validation/Code.cs
```diff
@@ -55,6 +55,29 @@ public static IEnumerable<T> ExpectsAllNotNull<T>([ValidatedNotNull] IEnumerable
...
+		/// </summary>
+		/// <remarks>Be careful to not pass enumerables that can be enumerated only once</remarks>
+		/// <typeparam name="T">Type of the enumerable</typeparam>
+		/// <param name="argumentValue">The argument value.</param>
+		/// <param name="argumentName">Name of the argument.</param>
+		/// <param name="tagId">Tag Id to log, leave null if no logging is needed.</param>
+		/// <exception cref="ArgumentException">Thrown if any argument  <paramref name="argumentValue"/> element is null.</exception>
+		/// <exception cref="ArgumentNullException">Thrown if the supplied argument <paramref name="argumentValue"/> is null.</exception>
+		public static IEnumerable<T> ExpectsAnyAndAllNotNull<T>([ValidatedNotNull] IEnumerable<T> argumentValue, string argumentName, uint? tagId) where T : class
```
vlad-ion: I feel that the name might be a bit confusing, I was reading it like (any and all) not null, what? It might be clear enough to call it ExpectsAllNotNull?

src/System/Validation/Code.cs
dan-hanlon: Updated to use `ICollection<T>`

PR #132, opened 2019-10-16
Enable nullability for Validation methods

ekorobchenko:
Added attributes to validation methods to show compiler that they make null checks. `Expects` methods now require generic argument to be constrained to reference type, which should be fine since it's casted to object internally but it's a potential breaking change.

### DIFF
--- a/src/DocumentDb/DocumentDbDocumentAdapter.cs
+++ b/src/DocumentDb/DocumentDbDocumentAdapter.cs
@@ -395,7 +395,7 @@ namespace Microsoft.Omex.DocumentDb
 		/// <returns>Replaced document.</returns>
 		public async Task<ResourceResponse<Document>> ReplaceDocumentAsync<T>(
 			string dbId, string collectionId, string docId, T entity, RequestOptions requestOptions = null)
-            where T:class
+			where T:class
 		{
 			Code.ExpectsNotNullOrWhiteSpaceArgument(dbId, nameof(dbId), TaggingUtilities.ReserveTag(0x2381b19a /* tag_961g0 */));
 			Code.ExpectsNotNullOrWhiteSpaceArgument(collectionId, nameof(collectionId), TaggingUtilities.ReserveTag(0x2381b19b /* tag_961g1 */));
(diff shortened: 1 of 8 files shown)


src/DocumentDb/DocumentDbDocumentAdapter.cs
```diff
@@ -395,6 +395,7 @@ public partial class DocumentDbAdapter : IDocumentDbAdapter
 		/// <returns>Replaced document.</returns>
 		public async Task<ResourceResponse<Document>> ReplaceDocumentAsync<T>(
 			string dbId, string collectionId, string docId, T entity, RequestOptions requestOptions = null)
+            where T:class
```
AndreyTretyak: Please use tabs here and in other places

src/System/Validation/Code.cs
```diff
@@ -107,11 +109,12 @@ public static IEnumerable<T> ExpectsAny<T>([ValidatedNotNull] IEnumerable<T> arg
...
 		/// <typeparam name="T">Type of argument to validate</typeparam>
 		/// <exception cref="ArgumentNullException">Thrown if the supplied argument <paramref name="argumentValue"/> is null.</exception>
-		public static T ExpectsArgument<T>([ValidatedNotNull] T argumentValue, string argumentName, uint? tagId)
+		public static T ExpectsArgument<T>([ValidatedNotNull] T? argumentValue, string argumentName, uint? tagId) 
+            where T:class
 		{
 			if (!ValidateArgument(argumentValue, argumentName, tagId))
 			{
-				ReportArgumentNull(argumentName);
+				return ReportArgumentNull<T>(argumentName);
```
AndreyTretyak: Nice trick

src/DocumentDb/DocumentDbDocumentAdapter.cs
ekorobchenko: Fixed, thanks.

AndreyTretyak: Thanks, it looks like we lost .editorconfing during the extraction of this repo. I'll create PR to add one.

AndreyTretyak: #133

doguva-msft: There is no such constraint on the underlying DocumentDb client SDK and so the wrapper should try to avoid introducing the constraints unless it is required for the purpose of the method. In this case and in many methods in this api, T can be a struct type as well as a class and this should still work.

---
In reply to: [335635225](https://github.com/microsoft/Omex/pull/132#discussion_r335635225) [](ancestors = 335635225)

ekorobchenko: I've removed constraints and created a separate method for the null check with type validation, thanks. Could you please check again?

PR #148, opened 2019-11-22
Add net472 as another target framework; update packages

aszczepanski:


src/DocumentDb/Microsoft.Omex.DocumentDb.csproj
```diff
@@ -16,13 +16,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="$(Microsoft_Azure_DocumentDB_Version)" />
-  </ItemGroup>
```
muiriswoulfe: You still have the core version defined within the props file. Is removing these if statements deliberate?

Dependencies.props
```diff
@@ -9,12 +9,13 @@
     <Microsoft_SourceLink_Vsts_Git_Version>1.0.0-beta2-19367-01</Microsoft_SourceLink_Vsts_Git_Version>
     <MSTest_TestAdapter_Version>2.0.0</MSTest_TestAdapter_Version>
     <MSTest_TestFramework_Version>2.0.0</MSTest_TestFramework_Version>
+    <System_Configuration_ConfigurationManager_Version>4.6.0</System_Configuration_ConfigurationManager_Version>
```
muiriswoulfe: Just to confirm - moving this from Pinned to AutoUpdate was deliberate?

aszczepanski: Yes, currently there's no need for it not to be autoupdated.

src/DocumentDb/Microsoft.Omex.DocumentDb.csproj
aszczepanski: Yes we don't need to have versions of this package.

muiriswoulfe: If that's the case, can you remove the second definition of it from the props file, i.e. the `System_Configuration_ConfigurationManager_Version_Core` one? If we're not using it anywhere, it doesn't seem like it would be relevant to have this anymore.

Dependencies.props
muiriswoulfe: Great that we can move this to autoupdate :smile:

src/DocumentDb/Microsoft.Omex.DocumentDb.csproj
muiriswoulfe: @aszczepanski: I think you might have missed my response here

aszczepanski: Thanks, I missed that unused variable. It's removed now.

muiriswoulfe: Thanks Adam

PR #151, opened 2019-12-02
Add missing parts for TimedScopes

AndreyTretyak:
Add missing parts for TimedScopes


src/System/TimedScopes/TimeScopeFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Omex.System.Logging;
+using Microsoft.Omex.System.TimedScopes.ReplayEventLogging;
+using Microsoft.Omex.System.Validation;
+
+namespace Microsoft.Omex.System.TimedScopes
+{
+	/// <summary>
+	/// Correlation Data
+	/// </summary>
+	public class TimedScopeFactory
```
armhil: Intentional to not have an interface?

src/System.AspNetCore/TimedScopesExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+		/// <summary>
+		/// Configures TimedScopes
+		/// </summary>
+		/// <param name="app">Application builder</param>
+		/// <returns>Application builder</returns>
+		public static IWebHostBuilder UseTimedScopes(this IWebHostBuilder app)
+		{
+			app.ConfigureServices(
+				service => service
+					.AddSingleton<CorrelationData>()
```
armhil: `AddSingleton` for correlation? Aren't we looking to have this `Transient` or `Scoped`?

src/System/TimedScopes/TimeScopeFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+		/// <summary>
+		/// Creates a scope (and starts by default)
+		/// </summary>
+		/// <param name="definition">TimeScope definition</param>
+		/// <param name="initialResult">Initial result to use</param>
+		/// <param name="startScope">Should the scope be automatically started (for use in e.g. 'using' statement)</param>
+		public TimedScope Create(TimedScopeDefinition definition, TimedScopeResult initialResult, bool startScope = true)
+		{
+			TimedScope scope = definition.Create(m_correlation.CurrentCorrelation, m_machineInformation, m_scopeLogger, m_replayEventConfigurator, m_timedScopeStackManager, initialResult, startScope);
+			if (startScope == true)
```
armhil: `if (startScope)`

tests/System.AspNetCore.UnitTests/ReplyEventConfiguratorUnitTests.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.Omex.System.TimedScopes;
```
aszczepanski: Not directly related to this PR, but there is a typo in file name - it should start with Replay

src/System/TimedScopes/ITimedScopeLogger.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+using Microsoft.Omex.System.Logging;
```
muiriswoulfe: >Microsoft.Omex.System.Logging [](start = 6, length = 29)

Is this required? I don't see any new dependencies in the class, so I'm not sure why this would be necessary (unless something moved to a different namespace)

src/System/TimedScopes/ITimedScopeProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+	/// </summary>
+	public interface ITimedScopeProvider
+	{
+		/// <summary>
+		/// Create timed scope
+		/// </summary>
+		/// <param name="definition">Timed scope definition</param>
+		/// <param name="initialResult">Intial result of timed scope</param>
+		/// <param name="startScope">Should create timed scope be started after creation</param>
+		/// <returns></returns>
```
muiriswoulfe: >returns [](start = 7, length = 7)

The `<returns>` comment is missing

src/System/TimedScopes/TimedScope.cs
```diff
@@ -58,9 +58,9 @@ public class TimedScope : IDisposable
 		/// <param name="correlationData">Correlation data</param>
 		/// <param name="machineInformation">Machine Information</param>
 		/// <returns>Newly created scope</returns>
-		public static TimedScope Create(TimedScopeDefinition scopeDefinition, CorrelationData correlationData, IMachineInformation machineInformation,
+		internal static TimedScope Create(TimedScopeDefinition scopeDefinition, CorrelationData correlationData, IMachineInformation machineInformation,
```
muiriswoulfe: >internal [](start = 2, length = 8)

An alternative in many cases is to make the class `internal` and leave this `public`. Would this work here, given the existing use cases? The advantage is that it better hides the implementation if you don't have external dependencies

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+		public static TimedScopeEventSource Instance => s_instance.Value;
+
+
+		/// <summary>
+		/// Static constructor
+		/// </summary>
+		static TimedScopeEventSource()
+		{
+			// A workaround for the problem where ETW activities do not get tracked until Tasks infrastructure is initialized.
+			// This problem will be fixed in .NET Framework 4.6.2.
```
muiriswoulfe: Do we support these old versions of .NET?

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+		public void LogEvent(
+				Category logCategory,
+				string name,
+				string subtype,
+				string metadata,
+				string userHash,
+				string serviceName,
+
+				TimedScopeResult result,
+				string correlationId,
```
muiriswoulfe: >string [](start = 4, length = 6)

Could this be a `Guid`?

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+		[NonEvent]
+		public void LogEvent(
+				Category logCategory,
+				string name,
+				string subtype,
+				string metadata,
+				string userHash,
+				string serviceName,
+
+				TimedScopeResult result,
```
muiriswoulfe: Nitpick: remove blank line

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+			string correlationIdAsString = SanitizeMdmString(correlationId, nameof(correlationId), name, logCategory);
+
+			WriteTimedScopeEvent(
+				nameAsString,
+				subTypeAsString,
+				metaDataAsString,
+				userHashAsString,
+				serviceNameAsString,
+
+				logCategory.Name,
```
muiriswoulfe: Nitpick: remove blank line

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+				result.ToString(),
+				correlationIdAsString,
+				Convert.ToInt64(durationMs));
```
muiriswoulfe: >ToInt64 [](start = 12, length = 7)

Please pass `CultureInfo.InvariantCulture` as the second parameter of `ToInt64` so that it behaves the same on every system

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+		}
+
+
```
muiriswoulfe: Nitpick: remove one of the 3 blank lines.

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+		private static string SanitizeMdmString(string mdmString, string name, string activityName, Category logCategory)
+		{
+			string validatedString = Convert.ToString(mdmString) ?? string.Empty;
```
muiriswoulfe: >ToString [](start = 36, length = 8)

Add `CultureInfo.InvariantCulture` as the second param to this method

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+			string result,
+			string correlationId,
+			long durationMs)
+		{
+			WriteEvent((int)EventIds.LogTimedScopeEventId, name ?? string.Empty, subType ?? string.Empty, metadata ?? string.Empty, userHash ?? string.Empty, serviceName ?? string.Empty, logCategory, result, correlationId, durationMs);
+		}
+
+
+		/// <summary>
+		/// Writes a Timed scope trace event with Level=Informational.
```
muiriswoulfe: >trace [](start = 27, length = 5)

What exactly does trace mean in this context? The method name includes `Trx` which potentially suggests a transaction. It would be good to clarify this for external consumers

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+			WriteTimedScopeTrxEvent(
+				nameAsString,
+				subTypeAsString,
+				metaDataAsString,
+				serviceNameAsString,
+
+				logCategory.Name,
+				result.ToString(),
+				correlationIdAsString,
+				Convert.ToInt64(durationMs));
```
muiriswoulfe: >ToInt64 [](start = 12, length = 7)

Add `CultureInfo.InvariantCulture` as the second param to this method

src/System/TimedScopes/TimedScopeProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+	/// <summary>
+	/// Correlation Data
+	/// </summary>
+	public class TimedScopeProvider : ITimedScopeProvider
+	{
+		/// <summary>
+		/// Create a timed scope
+		/// </summary>
+		/// <param name="customLogger">Use a custom logger for the timed scope</param>
```
muiriswoulfe: >customLogger [](start = 19, length = 12)

The parameters should be in the same order in the comments and in the constructor

src/System/TimedScopes/TimedScope.cs
AndreyTretyak: TimedScoped instance need to be returned in public method, so it can't be internal

src/System.AspNetCore/CustomActivityTimedScopeLogger.cs
```diff
@@ -0,0 +1,99 @@
...
+
+			if (scope.IsTransaction)
+			{
+				m_eventSource.LogEvent(Categories.TimingGeneral,
+					name: scope.Name,
+					subtype: scope.SubType ?? NullPlaceholder,
+					metadata: scope.MetaData ?? NullPlaceholder,
+					serviceName: ServiceName ?? NullPlaceholder,
+					result: scope.Result,
+					correlationId: data.VisibleId.ToString("D"),
```
muiriswoulfe: >ToString [](start = 35, length = 8)

Here and below, please pass `CultureInfo.InvariantCulture` to each of these `ToString` methods as the second param

src/System.AspNetCore/TimedScopesExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+	/// </summary>
+	public static class TimedScopesExtensions
+	{
+		/// <summary>
+		/// Register types required for timedscopes
+		/// </summary>
+		public static IServiceCollection AddTimedScopes(this IServiceCollection serviceCollection)
+		{
+			return serviceCollection
+				.AddSingleton<IMachineInformation, BasicMachineInformation>()
```
muiriswoulfe: >IMachineInformation [](start = 18, length = 19)

Is there a logic to the ordering? If not, I'd advise sorting alphabetically as there it avoids someone accidentally adding the same dependency twice, which can happen surprisingly often - particularly if two people are modifying the code at the same time

tests/System.UnitTests/TimedScopes/TimedScopeUnitTests.cs
```diff
@@ -37,9 +37,10 @@ public void Scope_TimedScopeLogger_IsCalled()
 			TimedScope scope;
 			CorrelationData data = new CorrelationData();
 
-			using (scope = new TimedScopeDefinition("TestScope")
-				.Create(data, machineInformation, customLogger: timedScopeLoggerMock.Object, replayEventConfigurator: replyEventConfiguratorMock.Object,
-				timedScopeStackManager: timedScopeStackManager, initialResult: TimedScopeResult.SystemError))
+
```
muiriswoulfe: Nitpick: remove multiple blank lines, including one with trailing whitespace

tests/System.UnitTests/TimedScopes/TimedScopeUnitTests.cs
```diff
@@ -59,8 +60,14 @@ public void Start_ShouldConstructTimedScope_WithTimerActive()
 			IMachineInformation machineInformation = new UnitTestMachineInformation();
 			ITimedScopeStackManager timedScopeStackManager = new TimedScopeStackManager(callContextManagerMock.Object, machineInformation);
 
-			using (TimedScope scope = TestHooks.CreateTestCountersUnitTestTimedScope(false, true, machineInformation, timedScopeLoggerMock.Object,
-				replyEventConfiguratorMock.Object, timedScopeStackManager))
+
```
muiriswoulfe: Nitpick: multiple blank lines with trailing whitespace

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+				Category logCategory,
+				string name,
+				string subtype,
+				string metadata,
+				string userHash,
+				string serviceName,
+
+				TimedScopeResult result,
+				string correlationId,
+				double durationMs
```
muiriswoulfe: >durationMs [](start = 11, length = 10)

In other places, this seems to be called `durationInMilliseconds`. I think this longer name is clearer to the consumers as `Ms` could potentially stand for many things

src/System/TimedScopes/TimedScopeEventSource.cs
```diff
@@ -0,0 +1,233 @@
...
+			)
```
muiriswoulfe: >) [](start = 3, length = 1)

Nitpick: The `)` is normally placed at the end of the last parameter. Applies throughout

src/System/TimedScopes/TimedScope.cs
muiriswoulfe: Thanks for confirming

src/System/TimedScopes/TimedScopeEventSource.cs
muiriswoulfe: The README.md says we only support .NET Core 3.0+ now

AndreyTretyak: If I'm understanding correctly README.md talking about SDK version required to build the project, but `Microsoft.Omex.System` string targets lower frameworks and used in projects that target net46

src/System/TimedScopes/TimedScopeEventSource.cs
AndreyTretyak: Fixed

src/System.AspNetCore/TimedScopesExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+	/// </summary>
+	public static class TimedScopesExtensions
+	{
+		/// <summary>
+		/// Register types required for timedscopes
+		/// </summary>
+		public static IServiceCollection AddTimedScopes(this IServiceCollection serviceCollection)
+		{
+			return serviceCollection
+			    .AddSingleton<ICallContextManager, CallContextManager>()
```
muiriswoulfe: >     [](start = 3, length = 4)

Nitpick: mix of spaces and tabs on this line

src/System/TimedScopes/TimedScopeEventSource.cs
AndreyTretyak: Fixed, thanks

src/System/TimedScopes/TimedScopeEventSource.cs
AndreyTretyak: Fixed

src/System/TimedScopes/TimedScopeProvider.cs
AndreyTretyak: Ordered, thanks you

src/System/TimedScopes/TimedScopeEventSource.cs
muiriswoulfe: Thanks for clarifying.

src/System/TimedScopes/TimedScopeEventSource.cs
AndreyTretyak: It might be possible, but I'm reluctant to change the contract in this class at this stage due to two reasons:
- I'm not sure that ETW able to work with GUID type properly and if it does not we will see it only after the event won't be available.
- From what I understand our log aggregation system configured to expect them using this contract, and changing it might require reconfiguring it.
Since it's a new implementation of TimedScopes I would prefer to minimize possible failure points and ensure that it's working with existing configuration and only after that attempt to change it

src/System/TimedScopes/TimedScopeEventSource.cs
AndreyTretyak: I think trace event here refers to ETW (Event Tracing for Windows) and it's just the name of log event. I agree that `Trx` could be expanded, but I don't want to change the contract due to the reasons stated [here](https://github.com/microsoft/Omex/pull/151/files#r356214797)

src/System.AspNetCore/TimedScopesExtensions.cs
AndreyTretyak: Sorry, fixed

src/System/TimedScopes/TimedScopeEventSource.cs
muiriswoulfe: That makes sense. I didn't realise that the `EventSource` class was an external dependency, but since it is, it is understandable that we are constrained.

src/System/TimedScopes/TimedScopeEventSource.cs
muiriswoulfe: If this is the name of the parameter in the class you're inheriting from, it's probably best to leave it as is. The comment is quite clear I believe.

AndreyTretyak: Thanks, I'll try to look into renaming this after we will verify that TimedScopes are working

PR #161, opened 2020-02-06
Start using GitHub actions

AndreyTretyak:
Attempt to use GitHub actions for PR validation

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -26,6 +26,6 @@ jobs:
     - name: Build with dotnet
       run: dotnet build --configuration ${{ matrix.configuration }}
     - name: Run Unit Tests
-      run: dotnet test --no-build
+      run: dotnet test --no-build --configuration ${{ matrix.configuration }}
     - name: Create NuGet packages
-      run: dotnet pack --no-build
+      run: dotnet pack --no-build --configuration ${{ matrix.configuration }}
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -1,4 +1,4 @@
-name: .NET Core
+name: Build and run unit tests 
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/pullrequest.yml
```diff
@@ -0,0 +1,30 @@
...
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: '3.1.100'
+    - name: Build with dotnet
+      run: dotnet build --configuration Release
```
muiriswoulfe: >-configuration Release [](start = 25, length = 22)

Is there any value in building Debug? If we use things like `#if DEBUG` in the code, the two can differ so building both ensures both modes work.

Also, would we want to build both x64 and x86? Or do we just build x64 now?

.github/workflows/pullrequest.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and run unit tests
```
muiriswoulfe: >  [](start = 30, length = 1)

Nitpick: trailing whitespace

muiriswoulfe:
Nice to see this moving to code for tracking 🙂

.github/workflows/pullrequest.yml
neilr81: I think we start with just x64 and release, it's very rare for us to deploy debug on server side code these days.

.github/workflows/pullrequest.yml
AndreyTretyak: fixed, thanks

.github/workflows/pullrequest.yml
AndreyTretyak: I've updated the pipeline to use both Release and Debug. About platform, I don't think it's needed for now since it's a library repo and all projects should be built for AnyCPU or locked to their platform withing AnyCPU solution profile.

PR #163, opened 2020-03-04
Add publish step to pipeline

AndreyTretyak:
Add publish step that will run on pushes to master


Directory.Build.props
```diff
@@ -38,7 +38,7 @@
     <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
     <!-- The release version should get updated on every new release -->
     <Version>20.0.3</Version>
-    <PreReleaseSuffix Condition="'$(PreReleaseSuffix)' == ''">local</PreReleaseSuffix>
+    <PreReleaseSuffix Condition="'$(PreReleaseSuffix)' == ''">$([System.DateTime]::Now.ToString(yyyyMMdd-HHmmss))</PreReleaseSuffix>
```
muiriswoulfe, line 41: `UtcNow` would be better in my view that `Now` as it is consistent irrespective of the timezone in use. It's also about 30 times more efficient

NuGet.Config
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
muiriswoulfe, line 1: Is this change deliberate?

Directory.Build.props
muiriswoulfe, line 41: I believe the "s" format would be good here and avoids using custom formatting. See https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings for more information.

If you add this, you should also add `CultureInfo.InvariantCulture` as the second parameter.

NuGet.Config
AndreyTretyak: Yes, there was some invisible character at the beginning of the file

Directory.Build.props
AndreyTretyak: Good point, updated, thank you

AndreyTretyak: This part would be used as version suffix for NuGet package (ex. 20.0.3-20200709-151556), so it might be sensible to `:` in "s" format (YYYY-MM-DDTHH:mm:ss), and in general, to me, it feels like in this case it's better to have direct control considering the usage of variable.

NuGet.Config
muiriswoulfe: Makes sense. Probably the Unicode BOM.

Directory.Build.props
muiriswoulfe: Ah that makes sense, I agree skipping the `:` in the NuGet package version makes sense.

It might still be beneficial to add `CultureInfo.InvariantCulture` as the second parameter, as theoretically the serialisation can very per locale (although I don't believe it will for this format string in any current version of .NET).

AndreyTretyak: I've tried a couple of options but I was not able to provide InvariantCulture into this MSBuild method call, so I'll probably leave it like this for now

muiriswoulfe: Thanks for trying. My guess as to how this would work would be:
```
$([System.DateTime]::Now.ToString("yyyyMMdd-HHmmss", [System.Globalization.CultureInfo]::InvariantCulture))
```

If that doesn't work, should be fine as is.

PR #153, opened 2019-12-31
Create Omex extension projects to simplify usage of common logic

AndreyTretyak:
Create extension projects to simplify usage of common logic:

- Microsoft.Omex.Extensions.Abstractions for common interfaces
- Microsoft.Omex.Extensions.Compatability to support old logging and validation logic
- Microsoft.Omex.Extensions.Logging implementation of  LoggerProvider
- Microsoft.Omex.Extensions.TimedScopes abstraction to measure code execution using Activity
- Microsoft.Omex.Extensions.Hosting providers entry point with initialized logger timescopes and DI
- Microsoft.Omex.Extensions.Hosting.Services entry point for SF Services
- Microsoft.Omex.Extensions.Hosting.Services.Web entry point for SF Web Services

A sample usage would be:
```csharp
public async Task Main(string[] args)
{
  await Host
    .CreateDefaultBuilder(args)
    .AddOmexCompatabilityServices()  // Microsoft.Omex.Extensions.Compatability, needed only if you want to use ULSLogger or Code.Validate
    .BuildStatelessService(  // Microsoft.Omex.Extensions.Hosting.Services
      "MyAwesomeService",
      builder =>
      {
        builder
          .AddServiceAction(RunAsyncAction1)
          .AddServiceAction(RunAsyncAction2)
          .AddServiceListener("MyListener1", Listener1)
          .AddServiceListener("MyListener2", Listener2)
          .AddKestrelListener<Startup>(  // Microsoft.Omex.Extensions.Hosting.Services.Web
            "MyWebListener",
            ServiceFabricIntegrationOptions.UseReverseProxyIntegration);
      })
    .RunAsync();
}
```

PR #168, opened 2020-03-12
Omex Extension Follow-up

AndreyTretyak:
Omex Extension Follow-up changes since [extension pull request](https://github.com/microsoft/Omex/pull/153) became too big

### DIFF
--- a/src/Extensions/Abstractions/EventSources/EventSourcesEventIds.cs
+++ b/src/Extensions/Abstractions/EventSources/EventSourcesEventIds.cs
@@ -1,7 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-namespace Microsoft.Omex.Extensions.EventSources // diffirent namespace since this class has specific usage
+namespace Microsoft.Omex.Extensions.Abstractions.EventSources // diffirent namespace since this class has specific usage
 {
 	/// <summary>
 	/// Event ids enum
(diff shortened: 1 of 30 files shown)


src/Extensions/Abstractions/EventSourcesEventIds.cs
```diff
@@ -1,11 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-namespace Microsoft.Omex.Extensions.EventSources
+namespace Microsoft.Omex.Extensions.EventSources // diffirent namespace since this class has specific usage
```
muiriswoulfe: >EventSources [](start = 36, length = 12)

Seems like it would be better to just move this to the appropriate folder

AndreyTretyak: I was thinking about it, but it would be a single file and proper name then would be `Microsoft.Omex.Extensions.Abstractions.EventSources`, but you probably right about the separate folder.

AndreyTretyak: Renamed, thanks for pushing me to do a proper thing

src/Extensions/Abstractions/EventSources/EventSourcesEventIds.cs
```diff
@@ -1,11 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-namespace Microsoft.Omex.Extensions.EventSources
+namespace Microsoft.Omex.Extensions.Abstractions.EventSources // diffirent namespace since this class has specific usage
```
muiriswoulfe: >diffirent namespace since this [](start = 65, length = 30)

I don't think you want this comment anymore

AndreyTretyak: I've wanted to keep it to emphasize the importance of namespace

PR #169, opened 2020-03-17
Add guid correlation support and change activity processing

AndreyTretyak:
* Move all `Activty` processing into extensible activity observres
* Add logic for supporting guid correlation id
* Change activity API to correspond to future activity changes https://github.com/dotnet/designs/pull/98
* Switch to using BacgroungService for Omex SF services to reduce the amount of custom code
* Now all core for using activities inside Abstraction package, add TimedScopes package needed only to log activities


src/Extensions/Abstractions/Activities/TimedScope.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Omex.Extensions.Abstractions.Activities
+{
+	/// <summary>
+	/// Logs duration of activity
+	/// </summary>
+	/// <remarks>This class will be replaced with Activity after move to net 5, since it will be trackable and disposible</remarks>
```
vlad-ion: nit:
```suggestion
	/// <remarks>This class will be replaced with Activity after move to net 5, since it will be trackable and disposable</remarks>
```

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
+
+namespace Microsoft.Omex.Extensions.Abstractions.Activities
+{
+	/// <summary>
+	/// Extnesions to TimedScopes class to set additional information
```
vlad-ion: nit:
```suggestion
	/// Extensions to TimedScopes class to set additional information
```

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+	public static class TimedScopesExtensions
+	{
+		/// <summary>
+		/// Set result
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		public static TimedScope SetResult(this TimedScope timedScope, TimedScopeResult result)
+		{
+			timedScope.Activity.AddTag(ActivityTagKeys.Result, ActivityResultStrings.ResultToString(result));
+			return timedScope;
```
vlad-ion: What is the reason for also returning the timedscope here, as long as it is not immutable because the change is done on the existing instance anyway?

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
```diff
@@ -0,0 +1,34 @@
...
+		{
+			m_options = options;
+		}
+
+		public void OnStart(Activity activity, object? payload)
+		{
+			if (m_options.Value.AddObsoleteCorrelationToActivity)
+			{
+#pragma warning disable CS0618
+				if (activity.GetObsoleteCorrelationId() != null)
```
vlad-ion: I'm not sure how this is supposed to work. So if there is a correlation id already set a new one on start, but what happens if it's not, then with the current code it will remain null.

src/Extensions/Logging/ServiceCollectionExtensions.cs
```diff
@@ -48,13 +48,14 @@ public static IServiceCollection AddOmexLogging(this IServiceCollection serviceC
...
 			serviceCollection.TryAddTransient<IExecutionContext, BasicMachineInformation>();
 			serviceCollection.TryAddTransient<IExternalScopeProvider, LoggerExternalScopeProvider>();
-
 			serviceCollection.TryAddTransient<IActivityProvider, ReplayableActivityProvider>();
 
 			serviceCollection.TryAddSingleton(p => OmexLogEventSource.Instance);
 			serviceCollection.TryAddTransient<ILogEventReplayer, OmexLogEventSender>();
 			serviceCollection.TryAddTransient<ILogEventSender, OmexLogEventSender>();
 
+			serviceCollection.TryAddEnumerable(ServiceDescriptor.Transient<IActivityStartObserver, ActivityEnhensementObserver>());
```
vlad-ion: Typo:
```suggestion
			serviceCollection.TryAddEnumerable(ServiceDescriptor.Transient<IActivityStartObserver, ActivityEnhancementObserver>());
```

src/Extensions/TimedScopes/ITimedScopeEventSender.cs
```diff
@@ -11,7 +13,6 @@ public interface ITimedScopeEventSender
 		/// <summary>
 		/// Log timed scope end information
 		/// </summary>
-		/// <param name="scope">Ended timed scope</param>
-		void LogTimedScopeEndEvent(TimedScope scope);
+		void LogActivityStop(Activity activity);
```
vlad-ion: nit: missing param description

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
AndreyTretyak: No that's a mistake, thank you, I'll add some test to cover this case

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
```diff
@@ -20,31 +23,46 @@ public TimedScopeEventSender(TimedScopeEventSource eventSource, IHostEnvironment
...
-			string metadata = scope.Metadata;
-			TimedScopeResult result = scope.Result;
-			Activity activity = scope.Activity;
 			string name = activity.OperationName;
 			string correlationId = activity.Id;
 			double durationMs = activity.Duration.TotalMilliseconds;
 			string userHash = activity.GetUserHash(); //TODO: We need add middleware that will set userhash in compliant way and IsTransaction GitHub Issue #166
 			bool isTransaction = activity.IsTransaction();
 
+
```
vlad-ion: nit:
```suggestion
```

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
```diff
@@ -20,31 +23,46 @@ public TimedScopeEventSender(TimedScopeEventSource eventSource, IHostEnvironment
...
+			string subtype = NullPlaceholder;
```
vlad-ion: Is there a reason to set these to "null" by default instead of ""?

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+using System;
+using System.Diagnostics;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Omex.Extensions.TimedScopes.UnitTests
+{
+	[TestClass]
```
vlad-ion: You could add a TestCategory("") in all test files.

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
AndreyTretyak: It's done to follow the same fluent API style that `Activity.AddTag` or `Activity.AddBaggage` is using. It's just a convenience that allows changing configuration like:
```csharp
TimedScope = provider.Create(definition)
    .SetResult(Success)
    .SetSubType("ImportantSubType")
    .SetMethadata("SuperUsefulInfo");
```

tests/Extensions/Abstractions.UnitTests/Activities/TimedScopeTests.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
vlad-ion: missing header:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license.

﻿using System;
```

tests/Extensions/Hosting.Services.UnitTests/ServiceAccessorTests.cs
```diff
@@ -7,6 +7,8 @@
 
 namespace Hosting.Services.UnitTests
 {
+
+
```
vlad-ion: remove newlines.

src/Extensions/Abstractions/Activities/Processing/ActivityResultStrings.cs
```diff
@@ -0,0 +1,52 @@
...
+		/// <summary>
+		/// Returns coresponding to enum string value with creating new string
+		/// </summary>
+		public static string ResultToString(TimedScopeResult result) =>
+			result switch
+			{
+				TimedScopeResult.SystemError => SystemError,
+				TimedScopeResult.ExpectedError => ExpectedError,
+				TimedScopeResult.Success => Success,
+#pragma warning disable CS0618 // need to handle all enum values
```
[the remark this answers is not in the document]
AndreyTretyak: `_` will throw an exception and intended for a case where a new value was added to enum, and `Unknown` is a valid value that we should not use. To be honest I'm not sure why it exists. But I see your point it might be easier to just throw in case of its usage

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
AndreyTretyak: I've thought it's some kind of ETW requirement, I've actually wanted to ask you about it :)

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
vlad-ion: Ah, right.

src/Extensions/Abstractions/Activities/Processing/ActivityResultStrings.cs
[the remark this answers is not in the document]
vlad-ion: Ok, whatever works for you in this case.

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
vlad-ion: I have no idea. If it was done like this before then leave it like that.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+		public static Activity MarkAsTransaction(this Activity activity) =>
+			activity.AddBaggage(TransactionMarkerKey, TransactionMarkerValue);
+
+
+		/// <summary>
+		/// Get correlation guid that is used by old Omex services
+		/// </summary>
+		[Obsolete(CorrelationIdObsoleteMessage, false)]
+		public static Guid? GetObsoleteCorrelationId(this Activity activity) =>
+			Guid.TryParse(activity.GetBaggageItem(ObsoleteCorrelationId), out Guid correlation)
```
muiriswoulfe: >correlation [](start = 74, length = 11)

I assume this API returns `Guid.Empty` in the case of failures. Are we sure this isn't sufficient for the obsolete API? Do we definitely need to switch this to `null` in this case?

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+		public static Activity SetObsoleteCorrelationId(this Activity activity, Guid correlation) =>
+			activity.AddBaggage(ObsoleteCorrelationId, correlation.ToString());
+
+
+		/// <summary>
+		/// Get transaction id that is used by old Omex services
+		/// </summary>
+		[Obsolete(TransactionIdObsoleteMessage, false)]
+		public static uint? GetObsolteteTransactionId(this Activity activity) =>
+			uint.TryParse(activity.GetBaggageItem(ObsoleteTransactionId), out uint transactionId)
```
muiriswoulfe: >transactionId [](start = 74, length = 13)

Same comment about the `null` check as the one above

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+				: (uint?)null;
+
+
+		/// <summary>
+		/// Set transaction id that is used by old Omex services
+		/// </summary>
+		/// <remarks>This property would be transfered to child activity and via web requests</remarks>
+		[Obsolete(TransactionIdObsoleteMessage, false)]
+		public static Activity SetObsoleteTransactionId(this Activity activity, uint transactionId) =>
+			activity.AddBaggage(ObsoleteTransactionId, transactionId.ToString());
```
muiriswoulfe: >ToString [](start = 60, length = 8)

Add `CultureInfo.InvariantCulture` as a parameter to avoid differences in behaviour on different machines

src/Extensions/Abstractions/Activities/TimedScope.cs
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+
+		/// <summary>
+		/// Stop TimedScope and log informations about it
+		/// </summary>
+		public void Stop() => s_listener.StopActivity(Activity, this);
+
+
+		internal Activity Activity { get; }
```
muiriswoulfe: >internal [](start = 2, length = 8)

`internal` is a less usual access modifier. Would `protected` work?

src/Extensions/Abstractions/Activities/Processing/ActivityTagKeys.cs
```diff
@@ -0,0 +1,27 @@
...
+	/// Activity does not have methods to extract tags efficiently, so we need to expose keys to provide ability of extracting tags
+	/// </summary>
+	public static class ActivityTagKeys
+	{
+		/// <summary>
+		/// Activity result tag key
+		/// </summary>
+		public static string Result { get; } = "Result";
+
+		/// <summary>
```
muiriswoulfe: Nit: add another blank line for consistent formatting

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
```diff
@@ -0,0 +1,34 @@
...
+		public ActivityEnhensementObserver(IOptions<OmexLoggingOptions> options)
+		{
+			m_options = options;
+		}
+
+		public void OnStart(Activity activity, object? payload)
+		{
+			if (m_options.Value.AddObsoleteCorrelationToActivity)
+			{
+#pragma warning disable CS0618
```
muiriswoulfe: >pragma warning disable CS0618 [](start = 1, length = 29)

Could you add a comment clarifying why this needs to be suppressed?

src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
```diff
@@ -20,65 +20,90 @@ static OmexLogEventSender()
...
-#pragma warning disable 618 // need to be used for to process reserved tags from GitTagger
+#pragma warning disable CS0618 // Need to be used for to process reserved tags from GitTagger
 				? TagsExtensions.TagIdAsString(eventId.Id)
-#pragma warning restore 618
+#pragma warning restore CS0618
 				: eventId.Id.ToString(CultureInfo.InvariantCulture);
 
-			string traceIdAsString = traceId.ToHexString();
+
+			string activityId = string.Empty;
```
muiriswoulfe: Remove second blank line

muiriswoulfe:
> 			string applicationName,

The parameters here are replicated throughout and then passed directly down the hierarchy of methods as far as I can tell, which is typically considered a code smell. It would be preferable to create a property bag class and pass this around, to avoid all of the replication

---
Refers to: src/Extensions/Logging/Internal/EventSource/OmexLogEventSource.cs:15 in 1f380b6. [](commit_id = 1f380b621457018f1f77775f38763a2bcd460432, deletion_comment = False)

src/Extensions/Logging/Internal/Replayable/ReplayableActivityStopObserver.cs
```diff
@@ -0,0 +1,38 @@
...
+
+		public void OnStop(Activity activity, object? payload)
+		{
+			if (m_options.Value.ReplayLogsInCaseOfError && ShouldReplayEvents(activity))
+			{
+				m_logReplayer.ReplayLogs(activity);
+			}
+		}
+
+
```
muiriswoulfe: Nitpick: Inconsistent use of single and double line breaks between methods in this class.

src/Extensions/Logging/Internal/Replayable/ReplayableActivityStopObserver.cs
```diff
@@ -0,0 +1,38 @@
...
+		private bool ShouldReplayEvents(Activity activity)
+		{
+			string resultTagValue = activity.Tags.FirstOrDefault(p => string.Equals(p.Key, ActivityTagKeys.Result, StringComparison.Ordinal)).Value;
+			return string.Equals(ActivityResultStrings.SystemError, resultTagValue);
```
muiriswoulfe: >Equals [](start = 17, length = 6)

It would be better to specify `StringComparison.Ordinal[IgnoreCase]` here to make the intention clearer

src/Extensions/TimedScopes/ITimedScopeEventSender.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+using System.Diagnostics;
```
muiriswoulfe: >Diagnostics [](start = 13, length = 11)

Do you really need this dependency? It's just an interface

src/Extensions/TimedScopes/ServiceCollectionExtensions.cs
```diff
@@ -19,6 +20,9 @@ public static class ServiceCollectionExtensions
 		public static IServiceCollection AddTimedScopes(this IServiceCollection serviceCollection)
 		{
 			Activity.DefaultIdFormat = ActivityIdFormat.W3C;
+			serviceCollection.AddHostedService<ActivityObserversIntializer>();
+			serviceCollection.TryAddEnumerable(ServiceDescriptor.Transient<IActivityStopObserver, ActivityStopObserver>());
+
 			serviceCollection.TryAddTransient<IActivityProvider, SimpleActivityProvider>();
```
muiriswoulfe: Is there a reason for the line break? In my mind, it's indicating some difference but I'm not seeing one here

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
```diff
@@ -0,0 +1,94 @@
...
+			m_observerLifetime = DiagnosticListener.AllListeners.Subscribe(this);
+			return Task.CompletedTask;
+		}
+		
+		public Task StopAsync(CancellationToken cancellationToken)
+		{
+			foreach (IDisposable disposable in m_disposables)
+			{
+				disposable.Dispose();
+			}
```
muiriswoulfe: >	} [](start = 2, length = 2)

Nitpick: It's often considered best practice to skip a blank line after `}`

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
```diff
@@ -0,0 +1,94 @@
...
+			m_observerLifetime?.Dispose();
+			return Task.CompletedTask;
+		}
+
+		private bool IsActivityStart(string eventName) => m_activityStartObservers.Any() && eventName.EndsWith(".Start", StringComparison.Ordinal);
```
muiriswoulfe: >Any [](start = 77, length = 3)

`.Any()` uses LINQ which can be inefficient. Since we don't need LINQ here, it would be better to check `.Count != 0` (or similar).

Applies to the line below as well

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
```diff
@@ -0,0 +1,94 @@
...
+		void IObserver<KeyValuePair<string, object>>.OnNext(KeyValuePair<string, object> value)
+		{
+			Activity activity = Activity.Current;
+
+			if (IsActivityStart(value.Key))
+			{
+				OnActivityStarted(activity, value.Value);
+			}
+
+			if (IsActivityStop(value.Key))
```
muiriswoulfe: >if [](start = 3, length = 2)

As far as I can tell, this can't be both, so you can avoid the check in some cases, by changing this to an `else if`

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
```diff
@@ -20,31 +23,45 @@ public TimedScopeEventSender(TimedScopeEventSource eventSource, IHostEnvironment
...
+					resultAsString = pair.Value;
+				}
+				else if (string.Equals(ActivityTagKeys.SubType, pair.Key, StringComparison.Ordinal))
+				{
+					subtype = pair.Value;
+				}
+				else if (string.Equals(ActivityTagKeys.Metadata, pair.Key, StringComparison.Ordinal))
+				{
+					metadata = pair.Value;
+				}
```
muiriswoulfe: A `switch` statement would be nicer here, particularly since it's case insensitive. Just be warned that it will be less tolerant of `null`s

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+			string userHash = "some hash value";
+
+			activity1.SetUserHash(userHash);
+
+			Assert.AreEqual(userHash, activity1.GetUserHash());
+			Assert.AreNotEqual(userHash, activity2.GetUserHash());
+		}
+
+
```
muiriswoulfe: Nitpick: Inconsistent number of lines between methods

tests/Extensions/Abstractions.UnitTests/Activities/TimedScopeTests.cs
```diff
@@ -0,0 +1,71 @@
...
+			Assert.IsNull(scope.Activity.Id);
+
+			scope.Start();
+
+			Assert.IsNotNull(scope.Activity.Id);
+		}
+
+
+		[TestMethod]
+		public void Stop_MultipleCallsIgnored()
```
muiriswoulfe: >Stop_MultipleCallsIgnored [](start = 14, length = 25)

I'd format this method as `Stop_WhenCalledMultipleTimes_IgnoresSuperfluousCalls()` or something similar, to adhere to the tripartite naming scheme we favour. This can also be applied below and to other methods

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
muiriswoulfe: You may be able to remove the LINQ import if you remove this.

---
In reply to: [394383572](https://github.com/microsoft/Omex/pull/169#discussion_r394383572) [](ancestors = 394383572)

tests/Extensions/Hosting.Services.UnitTests/OmexHostedServiceTests.cs
```diff
@@ -44,24 +40,22 @@ public async Task StartAsync_ProperlyCanceled()
 		[TestMethod]
 		public async Task StartAsync_HandlesExceptions()
 		{
-			MockRunner runnerMock = new MockRunner(t => Task.Run(() => throw new Exception("Totaly valid exeption")));
-			MockLifetime lifetimeMock = new MockLifetime();
+			MockRunner runnerMock = new MockRunner(t => Task.Run(async () =>
+				{
+					await Task.Delay(5);
```
muiriswoulfe: >Delay [](start = 16, length = 5)

I'd suggest adding `.ConfigureAwait(false)` to here and the outer loop for better correspondence between this and the product code.

Same for another few `await`s below

tests/Extensions/Hosting.Services.UnitTests/OmexHostedServiceTests.cs
```diff
@@ -44,24 +40,22 @@ public async Task StartAsync_ProperlyCanceled()
...
+					throw new Exception("Totaly valid exeption");
```
muiriswoulfe: >Exception [](start = 15, length = 9)

If we add static analysers such as SonarQube to the project, they will complain about throwing a generic `Exception`. Therefore, it might be better to throw something more specific here now

tests/Extensions/Hosting.Services.UnitTests/ServiceAccessorTests.cs
```diff
@@ -7,6 +7,8 @@
...
 	[TestClass]
 	public class ServiceAccessorTests
```
muiriswoulfe: The changes to this file appear to be by mistake

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Hosting.Services.UnitTests
+{
+	[TestClass]
+	public class ActivityObserversIntializerTests
+	{
+		[TestMethod]
+		public void ActivityObserversIvokedProperly()
```
muiriswoulfe: >v [](start = 32, length = 1)

Typo in `Invoked`

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+			string name = nameof(ActivityObserversIvokedProperly);
+			Mock<IActivityStartObserver> startObserver = new Mock<IActivityStartObserver>();
+			Mock<IActivityStopObserver> stopObserver = new Mock<IActivityStopObserver>();
+			ActivityObserversIntializer initializer = new ActivityObserversIntializer(
+				new [] { startObserver.Object },
+				new [] { stopObserver.Object });
+
+			try
+			{
+				initializer.StartAsync(CancellationToken.None);
```
muiriswoulfe: >StartAsync [](start = 16, length = 10)

Do you wish to `await` this? I realise your code doesn't require it, but it seems like someone could potentially change the implementation in the future so that having an `await` here would make things simpler

tests/Extensions/TimedScopes.UnitTests/ServiceCollectionTests.cs
```diff
@@ -4,6 +4,7 @@
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
```
muiriswoulfe: >Omex [](start = 16, length = 4)

Sort the usings

src/Extensions/TimedScopes/ITimedScopeEventSender.cs
AndreyTretyak: yes, it's required for `Activity` class

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
AndreyTretyak: I have not managed to make switch work here because `ActivityTagKeys.Metadata` and other properties are read-only and not a constant, and it's a bit risky to make them constants since they are used in different assemblies.

src/Extensions/Abstractions/Activities/Processing/ActivityResultStrings.cs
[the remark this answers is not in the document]
AndreyTretyak: It thinks you are right, it would be proper to throw an exception in case of Unknown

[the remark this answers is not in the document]
AndreyTretyak: Fixed, thank you

src/Extensions/Logging/ServiceCollectionExtensions.cs
AndreyTretyak: Renamed, thank you for noticing

src/Extensions/TimedScopes/ITimedScopeEventSender.cs
AndreyTretyak: I've removed param intentionally, but I'll improve summary comment

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
AndreyTretyak: I'll leave it like this, for now, it's not allocating almost anything, but I'll try to investigate it in future

AndreyTretyak: From short googling ti looks like event source might fail if you pass null instead of a sting, but I'll create a test a bit later

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Makes sense, thank you, added

tests/Extensions/Hosting.Services.UnitTests/ServiceAccessorTests.cs
AndreyTretyak: Removed, thank you

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: I think `null` in this case better indicates that value not set, I'm not sure that default values are not used in other context, but you might be right

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: Fixed. Out of curiosity does it actually does anything for `int`? I'm not aware of different formats for it.

tests/Extensions/Abstractions.UnitTests/Activities/TimedScopeTests.cs
AndreyTretyak: Oh, thank you, totally missed that

src/Extensions/Abstractions/Activities/TimedScope.cs
AndreyTretyak: unfortunately not, it required in extensions methods, another option would be to make it public

src/Extensions/Abstractions/Activities/Processing/ActivityTagKeys.cs
AndreyTretyak: Added, thanks, I'll create pr for changing formating after merging this one

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
AndreyTretyak: Added, thank you

src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
AndreyTretyak: Removed, thank you

src/Extensions/Logging/Internal/Replayable/ReplayableActivityStopObserver.cs
AndreyTretyak: > > ```
> >   	string applicationName,
> > ```
> 
> The parameters here are replicated throughout and then passed directly down the hierarchy of methods as far as I can tell, which is typically considered a code smell. It would be preferable to create a property bag class and pass this around, to avoid all of the replication
> 
> Refers to: src/Extensions/Logging/Internal/EventSource/OmexLogEventSource.cs:15 in [1f380b6](https://github.com/microsoft/Omex/commit/1f380b621457018f1f77775f38763a2bcd460432). [](commit_id = [1f380b6](https://github.com/microsoft/Omex/commit/1f380b621457018f1f77775f38763a2bcd460432), deletion_comment = False)

Unfortunately parameters signature is a part of `EvetSource` contract, it uses parameters (that need to be simple types) and parameter names to create an event.

AndreyTretyak: Not sure why my reply got inserted here, sorry, but empty lines also fixed, thank you

src/Extensions/Logging/Internal/Replayable/ReplayableActivityStopObserver.cs
AndreyTretyak: Thank you, skipped that one

src/Extensions/TimedScopes/ServiceCollectionExtensions.cs
AndreyTretyak: I've added it split enumerable registration (and hosted that is also enumerable) from Transient, but I don't have a strong opinion and okay removing it if you insisting.

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
AndreyTretyak: Added, thank you

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
AndreyTretyak: Replaced, thanks, but I still need Linq to call `ToArray()` method

src/Extensions/TimedScopes/Internal/ActivityObserversIntializer.cs
AndreyTretyak: I've skipped it in case condition will change in the future, but I don't think it will change any time soon

AndreyTretyak: Changed, thanks

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Fixed, thank you

tests/Extensions/Abstractions.UnitTests/Activities/TimedScopeTests.cs
AndreyTretyak: Updated names for both tests, thanks

tests/Extensions/Hosting.Services.UnitTests/OmexHostedServiceTests.cs
AndreyTretyak: Makes sense, added for all awaits in this file, thank you

tests/Extensions/Hosting.Services.UnitTests/OmexHostedServiceTests.cs
AndreyTretyak: Interesting point, changed the type of exception, thank you

tests/Extensions/Hosting.Services.UnitTests/ServiceAccessorTests.cs
AndreyTretyak: Yes, removed, thank you

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
AndreyTretyak: Fixed, thank you

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
AndreyTretyak: Make sense, thank you, fixed

tests/Extensions/TimedScopes.UnitTests/ServiceCollectionTests.cs
AndreyTretyak: Sorted, thank you

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
AndreyTretyak: Fixed and added unit tests  for observers thanks a lot

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: Answered here https://github.com/microsoft/Omex/pull/169#discussion_r394707906

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
muiriswoulfe: There are differing number systems around the world, which this could potentially get converted into. See [https://en.wikipedia.org/wiki/Numeral_system#/media/File:Numeral_Systems_of_the_World.svg](https://en.wikipedia.org/wiki/Numeral_system#/media/File:Numeral_Systems_of_the_World.svg)

---
In reply to: [394708563](https://github.com/microsoft/Omex/pull/169#discussion_r394708563) [](ancestors = 394708563)

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
muiriswoulfe: That makes sense. I only called it out because often people use null when `Guid.Empty` or similar would be sufficient for their needs

---
In reply to: [394707906](https://github.com/microsoft/Omex/pull/169#discussion_r394707906) [](ancestors = 394707906)

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+			string name = nameof(ActivityObserversInvokedProperly);
+			Mock<IActivityStartObserver> startObserver = new Mock<IActivityStartObserver>();
+			Mock<IActivityStopObserver> stopObserver = new Mock<IActivityStopObserver>();
+			ActivityObserversIntializer initializer = new ActivityObserversIntializer(
+				new [] { startObserver.Object },
+				new [] { stopObserver.Object });
+
+			try
+			{
+				await initializer.StartAsync(CancellationToken.None);
```
muiriswoulfe: >StartAsync [](start = 22, length = 10)

I'd suggest adding `.ConfigureAwait(false)` here to the async call here and below for consistency with the product code

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
muiriswoulfe: I don't see the comment

---
In reply to: [394721545](https://github.com/microsoft/Omex/pull/169#discussion_r394721545) [](ancestors = 394721545)

src/Extensions/TimedScopes/ServiceCollectionExtensions.cs
muiriswoulfe: It might be good to add a comment clarifying this for future readers

---
In reply to: [394724148](https://github.com/microsoft/Omex/pull/169#discussion_r394724148) [](ancestors = 394724148)

tests/Extensions/TimedScopes.UnitTests/ActivityObserversIntializerTests.cs
AndreyTretyak: Sorry missed this one, thank you, fixed

src/Extensions/TimedScopes/Internal/TimedScopeEventSender.cs
vlad-ion: I was thinking "", so string.Empty, not null.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: Oh, I was hoping we have a single standard at least in numbers :) thanks

src/Extensions/Logging/Internal/EventSource/ActivityEnhensementObserver.cs
AndreyTretyak: I'm not sure why it's not displayed :) it actually there https://github.com/microsoft/Omex/blob/master/src/Extensions/Logging/Internal/EventSource/ActivityEnhancementObserver.cs

src/Extensions/TimedScopes/ServiceCollectionExtensions.cs
AndreyTretyak: No sure that it worth a comment, it's like splitting different blocks in method, if blank line creates confusion we can just remove it

PR #170, opened 2020-03-19
Remove double empty line

AndreyTretyak:
Remove double empty line due to code convention changes


src/CodeGenerators/TimedScopeGen/TimedScopeClassTemplate.cs
```diff
@@ -5,7 +5,7 @@
 // <auto-generated>
```
muiriswoulfe: >auto-generated [](start = 4, length = 14)

This is an autogenerated file. You'll need to modify the template to ensure these changes persist

muiriswoulfe:
> 		/// <summary>

Might be good to correctly split up these properties as you're making these changes

---
Refers to: src/CodeGenerators/TimedScopeGen/TimedScopeClassTemplate.cs:333 in b5f9de1. [](commit_id = b5f9de14df6c3398c49e0eda85b20c19072d3122, deletion_comment = False)

src/Gating/Configuration/GatingConfiguration.cs
```diff
@@ -8,14 +8,13 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
-// 
+//
 // This source code was auto-generated by xsd, Version=4.7.2046.0.
```
muiriswoulfe: >generated [](start = 29, length = 9)

This file is autogenered, so ensure the template is also updated to ensure this doesn't revert

muiriswoulfe: There could actually be other templates that need to be modified too. Please check for .txt files like this

---
In reply to: [395092251](https://github.com/microsoft/Omex/pull/170#discussion_r395092251) [](ancestors = 395092251)

AndreyTretyak:
> > ```
> >   /// <summary>
> > ```
> 
> Might be good to correctly split up these properties as you're making these changes
> 
> Refers to: src/CodeGenerators/TimedScopeGen/TimedScopeClassTemplate.cs:333 in [b5f9de1](https://github.com/microsoft/Omex/commit/b5f9de14df6c3398c49e0eda85b20c19072d3122). [](commit_id = [b5f9de1](https://github.com/microsoft/Omex/commit/b5f9de14df6c3398c49e0eda85b20c19072d3122), deletion_comment = False)

Next PR that I was planning is to delete of TimedScopeGen :) so I don't think it worth investing in it

src/Gating/Configuration/GatingConfiguration.cs
AndreyTretyak: Thank you I've forgotten about txt templates, just pushed updates for txt templates in gating

PR #172, opened 2020-03-20
Change interfaces to work with Service instead of Context

AndreyTretyak:
Change interfaces to work with Service instead of Context

### DIFF
--- a/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
@@ -5,6 +5,7 @@ using System;
 using System.Fabric;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.ServiceFabric.Services.Communication.AspNetCore;
+using Microsoft.ServiceFabric.Services.Runtime;
 
 namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 {
@@ -17,39 +18,49 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 		/// Add Kestrel service listener to stateless service
 		/// </summary>
 		/// <typeparam name="TStartup">The type containing the startup methods for the web listener</typeparam>
-		public static ServiceFabricHostBuilder<StatelessServiceContext> AddKestrelListener<TStartup>(
-			this ServiceFabricHostBuilder<StatelessServiceContext> builder,
+		public static ServiceFabricHostBuilder<StatelessService> AddKestrelListener<TStartup>(
+			this ServiceFabricHostBuilder<StatelessService> builder,
 			string name,
 			ServiceFabricIntegrationOptions options,
 			Action<IWebHostBuilder>? builderExtension = null)
 				where TStartup : class =>
-					builder.AddKestrelListener<TStartup, StatelessServiceContext>(name, options, builderExtension);
+					builder.AddKestrelListener<TStartup,StatelessService,StatelessServiceContext>(
+						name,
+						options,
+						service => service.Context,
+						builderExtension);
 
 		/// <summary>
 		/// Add Kestrel service listener to stateful service
 		/// </summary>
 		/// <typeparam name="TStartup">The type containing the startup methods for the web listener</typeparam>
-		public static ServiceFabricHostBuilder<StatefulServiceContext> AddKestrelListener<TStartup>(
-			this ServiceFabricHostBuilder<StatefulServiceContext> builder,
+		public static ServiceFabricHostBuilder<StatefulService> AddKestrelListener<TStartup>(
+			this ServiceFabricHostBuilder<StatefulService> builder,
 			string name,
 			ServiceFabricIntegrationOptions options,
 			Action<IWebHostBuilder>? builderExtension = null)
 				where TStartup : class =>
-					builder.AddKestrelListener<TStartup, StatefulServiceContext>(name, options, builderExtension);
+					builder.AddKestrelListener<TStartup,StatefulService,StatefulServiceContext>(
+						name,
+						options,
+						service => service.Context,
+						builderExtension);
 
-		private static ServiceFabricHostBuilder<TContext> AddKestrelListener<TStartup, TContext>(
-			this ServiceFabricHostBuilder<TContext> builder,
+		private static ServiceFabricHostBuilder<TService> AddKestrelListener<TStartup,TService,TContext>(
+			this ServiceFabricHostBuilder<TService> builder,
 			string name,
 			ServiceFabricIntegrationOptions options,
+			Func<TService,TContext> getContext,
 			Action<IWebHostBuilder>? builderExtension = null)
 				where TStartup : class
 				where TContext : ServiceContext =>
-					builder.AddServiceListener(new KestrelListenerBuilder<TStartup, TContext>(
+					builder.AddServiceListener(new KestrelListenerBuilder<TStartup,TService,TContext>(
 						name,
 						options,
-						builder => BuilderExtension<TContext>(builder, builderExtension)));
+						getContext,
+						builder => builder.BuilderExtension<TContext>(builderExtension)));
 
-		private static void BuilderExtension<TContext>(IWebHostBuilder builder, Action<IWebHostBuilder>? builderExtension)
+		private static void BuilderExtension<TContext>(this IWebHostBuilder builder, Action<IWebHostBuilder>? builderExtension)
 			where TContext : ServiceContext
 		{
 			builderExtension?.Invoke(builder);
--- a/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 			ServiceFabricIntegrationOptions options,
 			Action<IWebHostBuilder>? builderExtension = null)
 				where TStartup : class =>
-					builder.AddKestrelListener<TStartup,StatelessService,StatelessServiceContext>(
+					builder.AddKestrelListener<TStartup, StatelessService, StatelessServiceContext>(
 						name,
 						options,
 						service => service.Context,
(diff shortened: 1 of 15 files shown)


src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -25,57 +26,53 @@ public static class HostBuilderExtensions
 		/// <summary>
 		/// Add actions that will be executed inside service fabric service RunAsync method
 		/// </summary>
-		public static ServiceFabricHostBuilder<TContext> AddServiceAction<TContext>(
-			this ServiceFabricHostBuilder<TContext> builder,
-			Func<TContext, CancellationToken, Task> action)
-				where TContext : ServiceContext =>
-					builder.AddServiceAction(new ServiceAction<TContext>(action));
+		public static ServiceFabricHostBuilder<TService> AddServiceAction<TService>(
+			this ServiceFabricHostBuilder<TService> builder,
```
muiriswoulfe: >builder [](start = 43, length = 7)

I'd argue that if you're going to add a `<summary>`, you should also add `<returns>`, `<param>` and `<typeparam>` comments although I realise you feel differently about this. If you're making this change, it applies throughout

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -25,57 +26,53 @@ public static class HostBuilderExtensions
...
 		/// <summary>
 		/// Configures host to run service fabric stateless service with initializded Omex dependencies
 		/// </summary>
 		public static IHost BuildStatelessService(
 			this IHostBuilder builder,
 			string serviceName,
-			Action<ServiceFabricHostBuilder<StatelessServiceContext>> builderAction) =>
-				builder.BuildServiceFabricService<OmexStatelessServiceRunner, StatelessServiceContext>(serviceName, builderAction);
+			Action<ServiceFabricHostBuilder<StatelessService>> builderAction) =>
+				builder.BuildServiceFabricService<OmexStatelessServiceRunner,StatelessService,StatelessServiceContext>(serviceName, builderAction);
```
muiriswoulfe: >StatelessService [](start = 65, length = 16)

Here and on the equivalent line below, leave a space after `,` as per out coding conventions to enhance readability

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -91,16 +88,16 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 			return collection.AddOmexServices();
 		}
 
-		private static IHost BuildServiceFabricService<TRunner,TContext>(
+		private static IHost BuildServiceFabricService<TRunner,TService,TContext>(
```
muiriswoulfe: >, [](start = 56, length = 1)

As above, leave a space after `,`

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -25,57 +26,53 @@ public static class HostBuilderExtensions
 		/// <summary>
 		/// Add actions that will be executed inside service fabric service RunAsync method
 		/// </summary>
-		public static ServiceFabricHostBuilder<TContext> AddServiceAction<TContext>(
-			this ServiceFabricHostBuilder<TContext> builder,
-			Func<TContext, CancellationToken, Task> action)
-				where TContext : ServiceContext =>
```
muiriswoulfe: >where TContext : ServiceContext [](start = 4, length = 31)

You previously had a `where` clause in the code to narrow where the methods could be used. Is there any scope for doing something similar here again?

src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
```diff
@@ -17,39 +18,49 @@ public static class HostBuilderExtensions
...
-		public static ServiceFabricHostBuilder<StatelessServiceContext> AddKestrelListener<TStartup>(
-			this ServiceFabricHostBuilder<StatelessServiceContext> builder,
+		public static ServiceFabricHostBuilder<StatelessService> AddKestrelListener<TStartup>(
+			this ServiceFabricHostBuilder<StatelessService> builder,
 			string name,
 			ServiceFabricIntegrationOptions options,
 			Action<IWebHostBuilder>? builderExtension = null)
 				where TStartup : class =>
-					builder.AddKestrelListener<TStartup, StatelessServiceContext>(name, options, builderExtension);
+					builder.AddKestrelListener<TStartup,StatelessService,StatelessServiceContext>(
```
muiriswoulfe: >, [](start = 40, length = 1)

Add spaces after `,`

muiriswoulfe: Applies throughout

---
In reply to: [395664015](https://github.com/microsoft/Omex/pull/172#discussion_r395664015) [](ancestors = 395664015)

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
muiriswoulfe: Even narrowing to `class` could be beneficial, although further narrowing of course would be even better

---
In reply to: [395663670](https://github.com/microsoft/Omex/pull/172#discussion_r395663670) [](ancestors = 395663670)

src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
```diff
@@ -14,24 +14,30 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
...
 
 		/// <inheritdoc />
-		public ICommunicationListener Build(TServiceContext context) =>
-			new KestrelCommunicationListener(context, Name, (url, listener) => BuildWebHost(context, url, listener));
+		public ICommunicationListener Build(TService service)
+		{
+			TServiceContext context = m_getContext(service);
+			return new KestrelCommunicationListener(context, Name, (url, listener) => BuildWebHost(context, url, listener));
+		}
+
```
muiriswoulfe: Extra blank line with trailing whitespace

tests/Extensions/Hosting.Services.UnitTests/HostBuilderExtensionsTests.cs
```diff
@@ -68,30 +67,30 @@ public void AddServiceActionUsingObject_RegisterInstance()
...
 
 			HostBuilder hostBuilder = new HostBuilder();
-			new ServiceFabricHostBuilder<ServiceContext>(hostBuilder).AddServiceAction(action);
+			new ServiceFabricHostBuilder<object>(hostBuilder).AddServiceAction(action);
 
-			IServiceAction<ServiceContext> resolvedAction = hostBuilder.Build().Services.GetService<IServiceAction<ServiceContext>>();
+			IServiceAction<object> resolvedAction = hostBuilder.Build().Services.GetService<IServiceAction<object>>();
 
-			await resolvedAction.RunAsync(MockStatelessServiceContextFactory.Default, CancellationToken.None);
+			await resolvedAction.RunAsync(new object(), CancellationToken.None);
```
muiriswoulfe: >RunAsync [](start = 24, length = 8)

Consider adding `.ConfigureAwait(false)`. Applies thrroughout

tests/Extensions/Hosting.Services.Web.UnitTests/KestrelListenerBuilderTests.cs
```diff
@@ -1,35 +1,37 @@
...
 		[TestMethod]
-		public void StatelessService_TypesRegistered() => CheckTypeRegistration(MockStatelessServiceContextFactory.Default);
+		public void StatelessService_TypesRegistered() => CheckTypeRegistration((StatelessService)new MockStatelessService(), service => service.Context);
 
 		[TestMethod]
-		public void StatefulService_TypesRegistered() => CheckTypeRegistration(MockStatefulServiceContextFactory.Default);
+		public void StatefulService_TypesRegistered() => CheckTypeRegistration((StatefulService)new MockStatefulService(), service => service.Context);
 
-		private void CheckTypeRegistration<TContext>(TContext context)
+		private void CheckTypeRegistration<TService, TContext>(TService service, Func<TService,TContext> getContex)
```
muiriswoulfe: >getContex [](start = 99, length = 9)

Typo: `getContext`

muiriswoulfe:
Great to see this change being made, and great to see the variables being changed to reflect the new naming. Too often we leave the variable names unchanged, leading to confusion. Thanks for doing a thorough job with this change.

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: I've been trying to find something, but StatefulService and StatelessService don't have an interface or base class in common. There is an option of creating one but that will require to expose more classes and will complicate method declaration, because it will be `TService : IServiceFabricService<TContext>` so we would have to have one more generic parameter everywhere.
An argument in favor of leaving it like that is that this library controls the only possible values for this generics (it always used for sealed class with internal constructor), but it could be easily changed by somebody in the future.

muiriswoulfe: I agree with you that creating yet more classes for this purpose doesn't seem like an ideal solution. It's a pity there's nothing we can use, but it should be fine as is. Perhaps if you add the `<typeparam>` comment, you can indicate there the only supported values. That way you'll make it clear to callers that you won't support them if they pass in entirely wrong values. I realise it's internal for now, but as you said this could be easily changed, particularly if someone were to decide to say fork the repo.

AndreyTretyak: You are right, it makes sense here. I'll add them here. 
Probably I was afraid that if I'll add them in one case I would have to add it everywhere :)

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: I'll add type parameters to explain limitations

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: Added typecomments

src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
AndreyTretyak: Added, thank you

src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
AndreyTretyak: Removed, thanks

tests/Extensions/Hosting.Services.UnitTests/HostBuilderExtensionsTests.cs
AndreyTretyak: Added it, thank you

tests/Extensions/Hosting.Services.Web.UnitTests/KestrelListenerBuilderTests.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -25,57 +26,57 @@ public static class HostBuilderExtensions
...
-					builder.AddServiceAction(new ServiceAction<TContext>(action));
+		/// <typeparam name="TService">Type of service fabric service, currently should be only StatelessService or StatefulService</typeparam>
```
muiriswoulfe: >StatelessService [](start = 90, length = 16)

`<see cref="StatelessService"/>` will add hyperlinks in the documentation

AndreyTretyak: Updated, thanks

PR #173, opened 2020-03-23
Add Omex middlewares to Hosted.Services.Web

AndreyTretyak:
- `ActivityEnrichmentMiddleware` to add additional properties to `Activity`
- `ResponseHeadersMiddleware` add headers to responces
- `ObsoleteCorrelationHeadersMiddleware` accepts old correlations
- `UseOmexExceptionHandler` adds exceptioin handlers

### DIFF
--- a/src/Extensions/Abstractions/Activities/ActivityExtensions.cs
+++ b/src/Extensions/Abstractions/Activities/ActivityExtensions.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// <summary>
 		/// Set result
 		/// </summary>
-		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		/// <remarks>This property won't be transferred to child activity or via web requests</remarks>
 		public static Activity SetResult(this Activity activity, TimedScopeResult result) =>
 			activity.AddTag(ActivityTagKeys.Result, ActivityResultStrings.ResultToString(result));
 
--- a/src/Extensions/Abstractions/Activities/ActivityExtensions.cs
+++ b/src/Extensions/Abstractions/Activities/ActivityExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Globalization;
+using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
 
 namespace Microsoft.Omex.Extensions.Abstractions.Activities
 {
@@ -40,6 +41,27 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// <remarks>This property would be transfered to child activity and via web requests</remarks>
 		public static Activity MarkAsTransaction(this Activity activity) =>
 			activity.AddBaggage(TransactionMarkerKey, TransactionMarkerValue);
+
+		/// <summary>
+		/// Set result
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		public static Activity SetResult(this Activity activity, TimedScopeResult result) =>
+			activity.AddTag(ActivityTagKeys.Result, ActivityResultStrings.ResultToString(result));
+
+		/// <summary>
+		/// Set sub type
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		public static Activity SetSubType(this Activity activity, string subType) =>
+			activity.AddTag(ActivityTagKeys.SubType, subType);
+
+		/// <summary>
+		/// Set metadata
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		public static Activity SetMetadata(this Activity activity, string metadata) =>
+			activity.AddTag(ActivityTagKeys.Metadata, metadata);
 
 		/// <summary>
 		/// Get correlation guid that is used by old Omex services
--- a/src/Extensions/Abstractions/Activities/TimedScope.cs
+++ b/src/Extensions/Abstractions/Activities/TimedScope.cs
@@ -37,6 +37,11 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// </summary>
 		public void Stop() => s_listener.StopActivity(Activity, this);
 
+		/// <remarks>
+		/// Activity property should not be public,
+		/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
+		/// After move to net core 5, <see cref="TimedScope"/> should be replaced with <see cref="Activity"/> completly
+		/// </remarks>
 		internal Activity Activity { get; }
 
 		void IDisposable.Dispose() => Stop();
--- a/src/Extensions/Abstractions/Activities/TimedScope.cs
+++ b/src/Extensions/Abstractions/Activities/TimedScope.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 
 		/// <remarks>
 		/// Activity property should not be public,
-		/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
+		/// since it may lead to calls to Activity.Stop that won't be properly logged until net core 5
 		/// After move to net core 5, <see cref="TimedScope"/> should be replaced with <see cref="Activity"/> completly
 		/// </remarks>
 		internal Activity Activity { get; }
--- a/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
+++ b/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
@@ -1,13 +1,15 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
-
 namespace Microsoft.Omex.Extensions.Abstractions.Activities
 {
 	/// <summary>
 	/// Extensions to TimedScopes class to set additional information
 	/// </summary>
+	/// <remarks>
+	/// This extensions class propagest calls of some activity extensions methond, to avoid making Activity property public,
+	/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
+	/// </remarks>
 	public static class TimedScopesExtensions
 	{
 		/// <summary>
@@ -16,7 +18,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
 		public static TimedScope SetResult(this TimedScope timedScope, TimedScopeResult result)
 		{
-			timedScope.Activity.AddTag(ActivityTagKeys.Result, ActivityResultStrings.ResultToString(result));
+			timedScope.Activity.SetResult(result);
 			return timedScope;
 		}
 
@@ -26,7 +28,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
 		public static TimedScope SetSubType(this TimedScope timedScope, string subType)
 		{
-			timedScope.Activity.AddTag(ActivityTagKeys.SubType, subType);
+			timedScope.Activity.SetSubType(subType);
 			return timedScope;
 		}
 
@@ -36,7 +38,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
 		public static TimedScope SetMetadata(this TimedScope timedScope, string metadata)
 		{
-			timedScope.Activity.AddTag(ActivityTagKeys.Metadata, metadata);
+			timedScope.Activity.SetMetadata(metadata);
 			return timedScope;
 		}
 	}
--- a/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
+++ b/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 	/// Extensions to TimedScopes class to set additional information
 	/// </summary>
 	/// <remarks>
-	/// This extensions class propagest calls of some activity extensions methond, to avoid making Activity property public,
+	/// This extensions class propagates calls of some activity extensions methods, to avoid making Activity property public,
 	/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
 	/// </remarks>
 	public static class TimedScopesExtensions
--- a/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
+++ b/src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 	/// </summary>
 	/// <remarks>
 	/// This extensions class propagates calls of some activity extensions methods, to avoid making Activity property public,
-	/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
+	/// since it may lead to calls to Activity.Stop that won't be properly logged until net core 5
 	/// </remarks>
 	public static class TimedScopesExtensions
 	{
--- a/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
@@ -23,14 +23,14 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 		/// Adds the default exception handling logic that will display a developer exception page in develop and a short message during deployment
 		/// </summary>
 		/// <param name="builder">Application builder</param>
-		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if the service needs to accept a request with old style correlation</param>
-		public static IApplicationBuilder UseOmexMiddlewares(this IApplicationBuilder builder, bool acceptObsoleteCorrelationHeader)
+		/// <param name="enableCorrelationHeaderBackwardCompatibility">Set it to true if the service needs to accept a request with old style correlation</param>
+		public static IApplicationBuilder UseOmexMiddlewares(this IApplicationBuilder builder, bool enableCorrelationHeaderBackwardCompatibility)
 		{
 			builder
 				.UseMiddleware<ActivityEnrichmentMiddleware>()
 				.UseMiddleware<ResponseHeadersMiddleware>();
 
-			if (acceptObsoleteCorrelationHeader)
+			if (enableCorrelationHeaderBackwardCompatibility)
 			{
 #pragma warning disable CS0618 // We need to add this middleware to accept old correlation header
 				builder.UseMiddleware<ObsoleteCorrelationHeadersMiddleware>();
--- a/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net;
+using System.Text.Json;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Diagnostics;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web
+{
+	/// <summary>
+	/// Extension to add Omex dependencies to ApplicationBuilder
+	/// </summary>
+	public static class ApplicationBuilderExtensions
+	{
+		/// <summary>
+		/// Adds the default exception heandling logic that will display developer expeption page in develop and short message during deployment
+		/// </summary>
+		/// <param name="builder">Application builder</param>
+		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if service need to accept request with old style correlation</param>
+		public static IApplicationBuilder UseOmexMiddlewares(this IApplicationBuilder builder, bool acceptObsoleteCorrelationHeader)
+		{
+			builder
+				.UseMiddleware<ActivityEnrichmentMiddleware>()
+				.UseMiddleware<ResponseHeadersMiddleware>();
+
+			if (acceptObsoleteCorrelationHeader)
+			{
+#pragma warning disable CS0618 // We need to add this middleware to accept old correlation header
+				builder.UseMiddleware<ObsoleteCorrelationHeadersMiddleware>();
+#pragma warning restore CS0618
+			}
+
+			return builder;
+		}
+
+		/// <summary>
+		/// Adds the default exception heandling logic that will display developer expeption page in develop and short message during deployment
+		/// </summary>
+		public static IApplicationBuilder UseOmexExceptionHandler(this IApplicationBuilder builder, IHostEnvironment environment)
+		{
+			if (environment.IsDevelopment())
+			{
+				builder.UseDeveloperExceptionPage();
+			}
+			else
+			{
+				builder.UseExceptionHandler(appError =>
+				{
+					appError.Run(context =>
+					{
+						context.Response.StatusCode = (int)HttpStatusCode.InternalServerError;
+						context.Response.ContentType = "application/json";
+
+						IExceptionHandlerFeature contextFeature = context.Features.Get<IExceptionHandlerFeature>();
+
+						Dictionary<string, string> errorDict = new Dictionary<string, string>
+						{
+							{ "Message", "Internal Server Error" },
+							{ "ErrorMessage", contextFeature?.Error?.Message ?? string.Empty },
+							{ "RequestId", Activity.Current?.Id ?? context.TraceIdentifier },
+							{ "Suggestion", "For local debugging, enable the set ASPNETCORE_ENVIRONMENT environment variable to 'Development' and restart the app." }
+						};
+
+						return context.Response.WriteAsync(JsonSerializer.Serialize(errorDict));
+					});
+				});
+			}
+
+			return builder;
+		}
+	}
+}
--- a/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 		/// Adds the default exception handling logic that will display a developer exception page in develop and a short message during deployment
 		/// </summary>
 		/// <param name="builder">Application builder</param>
-		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if service need to accept request with old style correlation</param>
+		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if the service needs to accept a request with old style correlation</param>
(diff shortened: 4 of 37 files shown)


src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -41,6 +42,27 @@ public static class ActivityExtensions
 		public static Activity MarkAsTransaction(this Activity activity) =>
 			activity.AddBaggage(TransactionMarkerKey, TransactionMarkerValue);
 
+		/// <summary>
+		/// Set result
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
+		public static Activity SetResult(this Activity activity, TimedScopeResult result) =>
```
dakanev-msft: >activity [](start = 49, length = 8)

What happens when activity is null?

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+	/// Extension to add Omex dependencies to ApplicationBuilder
+	/// </summary>
+	public static class ApplicationBuilderExtensions
+	{
+		/// <summary>
+		/// Adds the default exception heandling logic that will display developer expeption page in develop and short message during deployment
+		/// </summary>
+		/// <param name="builder">Application builder</param>
+		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if service need to accept request with old style correlation</param>
+		public static IApplicationBuilder UseOmexMiddlewares(this IApplicationBuilder builder, bool acceptObsoleteCorrelationHeader)
```
dakanev-msft: >acceptObsoleteCorrelationHeader [](start = 94, length = 31)

I am thinking, would the name enableBackwardCompatibility better in this context?

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+						context.Response.ContentType = "application/json";
+
+						IExceptionHandlerFeature contextFeature = context.Features.Get<IExceptionHandlerFeature>();
+
+						Dictionary<string, string> errorDict = new Dictionary<string, string>
+						{
+							{ "Message", "Internal Server Error" },
+							{ "ErrorMessage", contextFeature?.Error?.Message ?? string.Empty },
+							{ "RequestId", Activity.Current?.Id ?? context.TraceIdentifier },
+							{ "Suggestion", "For local debugging, enable the set ASPNETCORE_ENVIRONMENT environment variable to 'Development' and restart the app." }
```
dakanev-msft: >enable the set [](start = 45, length = 14)

Is this a typo, should it be

For local debugging, set the ...  ?

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Enrich request activity with Result, SubType and Methadata
```
dakanev-msft: >Methadata [](start = 54, length = 9)

nit: typo

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+			}
+
+			int statusCode = context.Response.StatusCode;
+
+			activity.SetResult(GetResult(statusCode))
+				.SetSubType($"{context.Request.Scheme}:{context.Connection.LocalPort}")
+				.SetMetadata(statusCode.ToString());
+		}
+
+		internal ActivityEnrichmentMiddleware(RequestDelegate next) => m_next = next;
```
dakanev-msft: >ActivityEnrichmentMiddleware [](start = 11, length = 28)

From readability perspective, to keep it aligned with the rest of the code, would it be better to have the constructor as first?

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,132 @@
...
+		/// Invoke middleware
+		/// </summary>
+		public Task InvokeAsync(HttpContext context)
+		{
+			ExtractCorrelationFromRequest(context.Request);
+			context.Response.OnStarting(SetCorrelationHeadersToResponce, context.Response);
+			return Task.CompletedTask;
+		}
+
+		private Task SetCorrelationHeadersToResponce(object state)
```
dakanev-msft: >Task [](start = 10, length = 4)

Do we have to return Task here?

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,132 @@
...
+
+			uint? transactionId = ParseUint(ExtractParameter(dataSources, s_transactionsNames));
+			if (transactionId.HasValue)
+			{
+				activity.SetObsoleteTransactionId(transactionId.Value);
+			}
+			else
+			{
+				return;
+			}
```
dakanev-msft: no need for return

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,132 @@
...
+				? value.FirstOrDefault(s => !string.IsNullOrWhiteSpace(s))
+				: null;
+
+		/// <summary>
+		/// Checks if the context is for a request that contains identifiers indicating that the request originated from an Office client or not
+		/// </summary>
+		private static bool IsClientRequest(HttpRequest request) =>
+			request.Headers.ContainsKey(OfficeClientVersionHeader)
+				? true
+				: request.Query.Count < 1 // Don't check empty params
```
dakanev-msft: >request.Query.Count < 1  [](start = 6, length = 24)

Would it be more readable to say

request.Query.Count == 0?

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,132 @@
...
+			"MS-CorrelationId",			// Correlation Id header used by other Microsoft services
+			"corr",						// Correlation query parameter
+			"HTTP_X_CORRELATIONID" };
+
+		private static readonly string[] s_officeClientQueryParameters = {
+			"client",					// Identifying the client application and platform
+			"av",						// Identifies the client application, platform and partial version
+			"app" };					// Identifies the client application
+
+		private const string CorrelationHeader = "X-CorrelationId";
```
dakanev-msft: nit: empty line between the members

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: Activity not nullable here, but if you pass null it will fail with `NullReferenceException`

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
AndreyTretyak: I agree that it sounds much better, but BackwardCompatibility seems a bit too generic to me because there might be many compatibility scenarios.

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
AndreyTretyak: Yes, it's much better this way, thank you

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
AndreyTretyak: Sure I think it would be better like this also, I was worried that less accessible member will be upper then more accessible, but I'm fine with changing it.

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Unfortunately yes, it's a structure of a delegate required by `OnStarting` method

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Thank you, it was for another code that was deleted :)

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Sure can do it like this also, I think I was worried about the case if Count will return negative value :)

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: I don't think it's regulated for private members, but you are right it's better like this

AndreyTretyak: Fixed, thanks

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
dakanev-msft: Should we add validation?

---
In reply to: [397166844](https://github.com/microsoft/Omex/pull/173#discussion_r397166844) [](ancestors = 397166844)

src/Extensions/Abstractions/Activities/TimedScope.cs
```diff
@@ -37,6 +37,11 @@ public TimedScope Start()
 		/// </summary>
 		public void Stop() => s_listener.StopActivity(Activity, this);
 
+		/// <remarks>
+		/// Activity property should not be public,
+		/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
```
vlad-ion:
```suggestion
		/// since it may lead to calls to Activity.Stop that won't be properly logged until net core 5
```

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -41,6 +42,27 @@ public static class ActivityExtensions
 		public static Activity MarkAsTransaction(this Activity activity) =>
 			activity.AddBaggage(TransactionMarkerKey, TransactionMarkerValue);
 
+		/// <summary>
+		/// Set result
+		/// </summary>
+		/// <remarks>This property won't be transfered to child activity or via web requests</remarks>
```
vlad-ion:
```suggestion
		/// <remarks>This property won't be transferred to child activity or via web requests</remarks>
```

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
```diff
@@ -1,13 +1,15 @@
...
 
-using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
-
 namespace Microsoft.Omex.Extensions.Abstractions.Activities
 {
 	/// <summary>
 	/// Extensions to TimedScopes class to set additional information
 	/// </summary>
+	/// <remarks>
+	/// This extensions class propagest calls of some activity extensions methond, to avoid making Activity property public,
```
vlad-ion:
```suggestion
	/// This extensions class propagates calls of some activity extensions methods, to avoid making Activity property public,
```

src/Extensions/Abstractions/Activities/TimedScopesExtensions.cs
```diff
@@ -1,13 +1,15 @@
...
+	/// since it may lead to calls to Activity.Stop that won't be properly logged untill net core 5
```
vlad-ion:
```suggestion
	/// since it may lead to calls to Activity.Stop that won't be properly logged until net core 5
```

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web
+{
+	/// <summary>
+	/// Extension to add Omex dependencies to ApplicationBuilder
+	/// </summary>
+	public static class ApplicationBuilderExtensions
+	{
+		/// <summary>
+		/// Adds the default exception heandling logic that will display developer expeption page in develop and short message during deployment
```
vlad-ion:
```suggestion
		/// Adds the default exception handling logic that will display developer exception page in develop and a short message during deployment
```

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+		/// </summary>
+		/// <param name="builder">Application builder</param>
+		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if service need to accept request with old style correlation</param>
```
vlad-ion:
```suggestion
		/// <param name="acceptObsoleteCorrelationHeader">Set it to true if the service needs to accept request with old style correlation</param>
```

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
vlad-ion: I think it's fine either way, but an more specific alternative for Daniel's suggestion could be: enableCorrelationHeaderCompatibility or enableCorrelationHeaderBackwardCompatibility.

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+#pragma warning disable CS0618 // We need to add this middleware to accept old correlation header
+				builder.UseMiddleware<ObsoleteCorrelationHeadersMiddleware>();
+#pragma warning restore CS0618
+			}
+
+			return builder;
+		}
+
+		/// <summary>
+		/// Adds the default exception heandling logic that will display developer expeption page in develop and short message during deployment
```
vlad-ion:
```suggestion
		/// Adds the default exception handling logic that will display a developer exception page in develop and a short message during deployment
```

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+				.SetSubType($"{context.Request.Scheme}:{context.Connection.LocalPort}")
+				.SetMetadata(statusCode.ToString());
+		}
+
+		private TimedScopeResult GetResult(int? statusCode) =>
+			statusCode >= 100 && statusCode < 400
+				? TimedScopeResult.Success
+				: statusCode >= 400 && statusCode < 500
+					? TimedScopeResult.ExpectedError
+					: TimedScopeResult.SystemError;
```
vlad-ion: I know this should never happen, but in case the statusCode is null or <100 (so invalid) is it ok to set SystemError as the result? Just asking.

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,129 @@
...
+	[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
+	internal class ObsoleteCorrelationHeadersMiddleware
+	{
+		/// <summary>
+		/// Invoke middleware
+		/// </summary>
+		public Task InvokeAsync(HttpContext context)
+		{
+			ExtractCorrelationFromRequest(context.Request);
+			context.Response.OnStarting(SetCorrelationHeadersToResponce, context.Response);
```
vlad-ion:
```suggestion
			context.Response.OnStarting(SetCorrelationHeadersToResponse, context.Response);
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,129 @@
...
+			return Task.CompletedTask;
+		}
+
+		private Task SetCorrelationHeadersToResponce(object state)
```
vlad-ion:
```suggestion
		private Task SetCorrelationHeadersToResponse(object state)
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,129 @@
...
+			return null;
+		}
+
+		private static string? ExtractParameter(IQueryCollection dataSources, string name) =>
+			dataSources.TryGetValue(name, out StringValues value) && value.Count > 0
+				? value.FirstOrDefault(s => !string.IsNullOrWhiteSpace(s))
+				: null;
+
+		/// <summary>
+		/// Checks if the context is for a request that contains identifiers indicating that the request originated from an Office client or not
```
vlad-ion:
```suggestion
		/// Checks if the context is for a request that contains identifiers indicating that the request originated from an Office client
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,129 @@
...
+			"X-TransactionId" };
+
+		private static readonly string[] s_correlationIdNames = {
+			CorrelationHeader,
+			"MS-CorrelationId",			// Correlation Id header used by other Microsoft services
+			"corr",						// Correlation query parameter
+			"HTTP_X_CORRELATIONID" };
+
+		private static readonly string[] s_officeClientQueryParameters = {
+			"client",					// Identifying the client application and platform
```
vlad-ion:
```suggestion
			"client",					// Identifies the client application and platform
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -0,0 +1,129 @@
...
+			"av",						// Identifies the client application, platform and partial version
+			"app" };					// Identifies the client application
+
+		private const string CorrelationHeader = "X-CorrelationId";
+
+		private const string OfficeClientVersionHeader = "X-Office-Version";
```
vlad-ion: Where is this used?

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT license.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Omex.Extensions.Logging;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Adds Omex headers to responces, like MachineId and BuildVersion
```
vlad-ion:
```suggestion
	/// Adds Omex headers to responses, like MachineId and BuildVersion
```

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+	public class ResponseHeadersMiddleware
+	{
+		internal ResponseHeadersMiddleware(IExecutionContext context) => m_context = context;
+
+		/// <summary>
+		/// Invoke middleware
+		/// </summary>
+		public Task InvokeAsync(HttpContext context)
+		{
+			context.Response.OnStarting(SetResponceHeaders, context.Response);
```
vlad-ion:
```suggestion
			context.Response.OnStarting(SetResponseHeaders, context.Response);
```

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+			return Task.CompletedTask;
+		}
+
+		private Task SetResponceHeaders(object state)
```
vlad-ion:
```suggestion
		private Task SetResponseHeaders(object state)
```

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -0,0 +1,36 @@
...
+		{
+			HttpResponse response = (HttpResponse)state;
+			response.Headers.Add("X-Machine", m_context.MachineId);
+			response.Headers.Add("X-BuildVersion", m_context.BuildVersion); //Renamed from X-OfficeVersion
```
vlad-ion: This seems strangely related to my other question above, but the naming in the comment has one less "-" than there

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
+using Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares;
+using Microsoft.Omex.Extensions.TimedScopes.UnitTests;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Hosting.Services.Web.UnitTests
+{
+	[TestClass]
```
vlad-ion: Suggesting [TestCategory()]

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
```diff
@@ -0,0 +1,67 @@
...
+	public class ActivityEnrichmentMiddlewareTests
+	{
+		[DataTestMethod]
+		[DataRow(99, TimedScopeResult.SystemError)]
```
vlad-ion: Test wit null value also, as the handling code had int?

tests/Extensions/Hosting.Services.Web.UnitTests/ResponseHeadersMiddlewareTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares;
+using Microsoft.Omex.Extensions.Logging;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Hosting.Services.Web.UnitTests
+{
+	[TestClass]
```
vlad-ion: TestCategory

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: I don't think it's needed here. 
My current view on null validation is that consumers should use non-nullable reference types or be responsible for not passing null. 
This should reduce the amount of code and simplify validation after moving to non-nullable reference types.

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
AndreyTretyak: I like `enableCorrelationHeaderBackwardCompatibility`, thank you

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
AndreyTretyak: Oh thanks, the type should not be nullable.
About status code less than 100 method should return `SystemError`
As far as I understand (I stole this method):
```
Min Int - 99 |100 - 399 | 400      -    499 | 500 - Max Int
SystemError | Success  | ExpectedError | SystemError 
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Renamed, thanks

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Renamed. thank you

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: It's used in `IsClientRequest` method, logic is that if we have this header then its a client request and we should drop correlation.

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
AndreyTretyak: Renamed, thank you

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
AndreyTretyak: Renamed, thank you

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
AndreyTretyak: That's one of the reasons for renaming, client header is `X-Office-Version` and our current version header `X-OfficeVersion` we use them for different purposes and that really confusing.

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
AndreyTretyak: It was a type mistake, type should be `int` not `int?`, thank you for finding it

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
AndreyTretyak: Not sure that it's needed, but added category, thank you

tests/Extensions/Hosting.Services.Web.UnitTests/ResponseHeadersMiddlewareTests.cs
AndreyTretyak: Not sure that it's needed, but added category, thank you

PR #177, opened 2020-03-25
Add stub initialize for comparability classes to use in unit tests

AndreyTretyak:
Add stub initialize for comparability classes to use in unit tests

### DIFF
--- a/src/Extensions/Compatibility/OmexCompatibilityIntializerService.cs
+++ b/src/Extensions/Compatibility/OmexCompatibilityIntializerService.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


src/Extensions/Compatibility/OmexCompatibilityIntializerService.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+
+namespace Microsoft.Omex.Extensions.Compatibility
+{
+	internal sealed class OmexCompatibilityIntializerService : IHostedService
```
dakanev-msft: Please add comments for the class

AndreyTretyak: It's an internal class it won't generate any documentation

PR #181, opened 2020-03-28
Add extension methods to register middlewares

AndreyTretyak:
Add extension methods to register middlewares and enable their resolution from DI


src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// This middleware should be removed after all services moved to use <see cref="Activity"/>
 	/// </remarks>
 	[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
-	internal class ObsoleteCorrelationHeadersMiddleware
+	public class ObsoleteCorrelationHeadersMiddleware
```
K-Cully: Doesn't InvokeAsync now need null guards added, since it will now be consumed externally?

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -12,7 +12,10 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// </summary>
 	public class ResponseHeadersMiddleware
```
K-Cully: Same question, since this is public and can be consumed by projects which don't use the new flag, why do the methods not have null guards?

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
K-Cully: Since this is public and can be consumed by projects which don't use the new flag, why do the methods not have null guards?

AndreyTretyak: You are right :( Only acceptable usage for these classes if for application builder:
`builder.UseMiddleware<ObsoleteCorrelationHeadersMiddleware>()`, but by making them public there is no way of enforcing it, I'll create extension methods for adding middlewares and will make all classes internal.

AndreyTretyak: Done, now all of them are internal and there are extension methods to register them

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
AndreyTretyak: Solved here https://github.com/microsoft/Omex/pull/181#discussion_r400235016

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
K-Cully: @AndreyTretyak I'm not seeing any changes to the access modifiers.

AndreyTretyak: @K-Cully sorry it looks like I have not pushed changes, should be there now

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -23,23 +24,47 @@ public static class ApplicationBuilderExtensions
...
+		/// Add middleware that adds Omex headers to responses, like MachineId and BuildVersion
+		/// </summary>
+		public static IApplicationBuilder UseResponseHeadersMiddleware(this IApplicationBuilder builder) =>
+			builder.UseMiddleware<ResponseHeadersMiddleware>();
+
+		/// <summary>
+		/// Add middleware that adds Omex headers to responses, like MachineId and BuildVersion
+		/// </summary>
+		[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
+		public static IApplicationBuilder UseObsoleteCorrelationHeadersMiddleware(this IApplicationBuilder builder) =>
```
K-Cully: Do these need to be public?

AndreyTretyak: Unfortunately yes, it's for the case when you need want to add only this middleware

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -23,23 +24,47 @@ public static class ApplicationBuilderExtensions
...
+		/// <summary>
+		/// Add middleware that adds Omex headers to responses, like MachineId and BuildVersion
+		/// </summary>
+		public static IApplicationBuilder UseResponseHeadersMiddleware(this IApplicationBuilder builder) =>
+			builder.UseMiddleware<ResponseHeadersMiddleware>();
+
+		/// <summary>
+		/// Add middleware that adds Omex headers to responses, like MachineId and BuildVersion
+		/// </summary>
+		[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
```
gijskoning:
```suggestion
		[Obsolete("Use it only if you need to communicate with services that use old correlation", false)]
```

src/Extensions/Hosting.Services.Web/ApplicationBuilderExtensions.cs
```diff
@@ -23,23 +24,47 @@ public static class ApplicationBuilderExtensions
 		/// Adds the default exception handling logic that will display a developer exception page in develop and a short message during deployment
 		/// </summary>
 		/// <param name="builder">Application builder</param>
+		/// <param name="environment">Enviroment information</param>
 		/// <param name="enableCorrelationHeaderBackwardCompatibility">Set it to true if the service needs to accept a request with old style correlation</param>
-		public static IApplicationBuilder UseOmexMiddlewares(this IApplicationBuilder builder, bool enableCorrelationHeaderBackwardCompatibility)
+		public static IApplicationBuilder UseOmexMiddlewares(
```
artempushkin: Middlewares strikes me as odd, Middleware would sound better without losing the meaning.

AndreyTretyak: I've wanted to emphasize there are many of them, not sure how grammatically correct it is.

PR #184, opened 2020-03-31
Fix middleware registration

AndreyTretyak:
Fix middleware registration

### DIFF
--- a/src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
+++ b/src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 				.AddSingleton<ActivityEnrichmentMiddleware>()
 				.AddSingleton<ResponseHeadersMiddleware>();
 
-#pragma warning disable CS0618 // We need to register all middlewares even obsolete
+#pragma warning disable CS0618 // We need to register all middlewares, even if obsolete
 			services.AddSingleton<ObsoleteCorrelationHeadersMiddleware>();
 #pragma warning restore CS0618
 		}
--- a/src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
+++ b/src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
@@ -6,6 +6,7 @@ using System.Fabric;
 using System.IO;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares;
 using Microsoft.ServiceFabric.Services.Communication.AspNetCore;
 using Microsoft.ServiceFabric.Services.Communication.Runtime;
 
@@ -64,9 +65,22 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 		{
 			services.AddSingleton<ServiceContext>(context);
 			services.AddSingleton(context);
+			AddMiddlewares(services);
 		}
 
 		private readonly ServiceFabricIntegrationOptions m_options;
+
 		private readonly Action<IWebHostBuilder> m_builderExtension;
+
+		private static void AddMiddlewares(IServiceCollection services)
+		{
+			services
+				.AddSingleton<ActivityEnrichmentMiddleware>()
+				.AddSingleton<ResponseHeadersMiddleware>();
+
+#pragma warning disable CS0618 // We need to register all middlewares even obsolete
+			services.AddSingleton<ObsoleteCorrelationHeadersMiddleware>();
+#pragma warning restore CS0618
+		}
 	}
 }
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
@@ -11,16 +11,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// <summary>
 	/// Enrich request activity with Result, SubType and Metadata
 	/// </summary>
-	internal class ActivityEnrichmentMiddleware
+	internal class ActivityEnrichmentMiddleware : IMiddleware
 	{
-		public ActivityEnrichmentMiddleware(RequestDelegate next) => m_next = next;
-
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public async Task InvokeAsync(HttpContext context)
+		async Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
 		{
-			await m_next(context).ConfigureAwait(false);
+			await next(context).ConfigureAwait(false);
 
 			Activity? activity = Activity.Current;
 
@@ -42,7 +37,5 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 				: statusCode >= 400 && statusCode < 500
 					? TimedScopeResult.ExpectedError
 					: TimedScopeResult.SystemError;
-
-		private readonly RequestDelegate m_next;
 	}
 }
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
@@ -20,12 +20,9 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// This middleware should be removed after all services moved to use <see cref="Activity"/>
 	/// </remarks>
 	[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
-	internal class ObsoleteCorrelationHeadersMiddleware
+	internal class ObsoleteCorrelationHeadersMiddleware : IMiddleware
 	{
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public Task InvokeAsync(HttpContext context)
+		Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
 		{
 			ExtractCorrelationFromRequest(context.Request);
 			context.Response.OnStarting(SetCorrelationHeadersToResponse, context.Response);
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
@@ -28,8 +28,6 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 			return Task.CompletedTask;
 		}
 
-
-
 		private readonly IExecutionContext m_context;
 	}
 }
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
@@ -10,14 +10,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// <summary>
 	/// Adds Omex headers to responses, like MachineId and BuildVersion
 	/// </summary>
-	internal class ResponseHeadersMiddleware
+	internal class ResponseHeadersMiddleware : IMiddleware
 	{
 		public ResponseHeadersMiddleware(IExecutionContext context) => m_context = context;
 
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public Task InvokeAsync(HttpContext context)
+		Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
 		{
 			context.Response.OnStarting(SetResponseHeaders, context.Response);
 			return Task.CompletedTask;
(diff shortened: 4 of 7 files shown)


src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -20,12 +20,9 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
...
 	/// </remarks>
 	[Obsolete("Use it only if you need to comunicate with services that use old correlation", false)]
-	internal class ObsoleteCorrelationHeadersMiddleware
+	internal class ObsoleteCorrelationHeadersMiddleware : IMiddleware
 	{
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public Task InvokeAsync(HttpContext context)
+		Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```
aszczepanski, line 25: Can't it be simplified to `InvokeAsync`?

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
```diff
@@ -11,16 +11,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// <summary>
 	/// Enrich request activity with Result, SubType and Metadata
 	/// </summary>
-	internal class ActivityEnrichmentMiddleware
+	internal class ActivityEnrichmentMiddleware : IMiddleware
```
aszczepanski, line 14: Just for my own understanding, could you please explain why this was causing issues before?

AndreyTretyak: I'm not sure that it's the best solution but from my experiments and debugging with middleware resolution it looks like without interface middleware constructor able to resolve only limited amount of objects (like HttpContext and the one passed to Use method), but if your class implements an interface then it's properly resolved from DI. I'm not sure why it's done like this and it's possible that I'm misunderstanding something. 
Here the link to resolution if interface implemented https://github.com/dotnet/aspnetcore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http.Abstractions/src/Extensions/UseMiddlewareExtensions.cs#L46
And here the link to resolution in other cases (only parameters used for resolution) https://github.com/dotnet/aspnetcore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http.Abstractions/src/Extensions/UseMiddlewareExtensions.cs#L92

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: It can, I've decided to use explicit interface implementation to emphasize the fact that the method should not be called from the instance of a class.

src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
```diff
@@ -64,9 +65,22 @@ private void ConfigureServices(TContext context, IServiceCollection services)
...
 			services.AddSingleton<ServiceContext>(context);
 			services.AddSingleton(context);
+			AddMiddlewares(services);
 		}
 
 		private readonly ServiceFabricIntegrationOptions m_options;
+
 		private readonly Action<IWebHostBuilder> m_builderExtension;
+
+		private static void AddMiddlewares(IServiceCollection services)
```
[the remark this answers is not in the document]
AndreyTretyak: I've tried that initially, but could not be an extension method because it's private and not in the static class, so the call would be a bit awkward ex: `AddMiddlewares(collection).Add...` and I've decided to just live it as a simple call since it's an only place where it's needed.

src/Extensions/Hosting.Services.Web/KestrelListenerBuilder.cs
```diff
@@ -64,9 +65,22 @@ private void ConfigureServices(TContext context, IServiceCollection services)
...
+		{
+			services
+				.AddSingleton<ActivityEnrichmentMiddleware>()
+				.AddSingleton<ResponseHeadersMiddleware>();
+
+#pragma warning disable CS0618 // We need to register all middlewares even obsolete
```
vlad-ion, line 81:
```suggestion
#pragma warning disable CS0618 // We need to register all middlewares, even if obsolete
```

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
```diff
@@ -11,16 +11,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
...
 	{
-		public ActivityEnrichmentMiddleware(RequestDelegate next) => m_next = next;
-
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public async Task InvokeAsync(HttpContext context)
+		async Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```
vlad-ion, line 16: I think it's more readable to leave the access mofidier in:
```suggestion
		private async Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
vlad-ion, line 25: Same as before:
```suggestion
		private Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -10,14 +10,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
...
-	internal class ResponseHeadersMiddleware
+	internal class ResponseHeadersMiddleware : IMiddleware
 	{
 		public ResponseHeadersMiddleware(IExecutionContext context) => m_context = context;
 
-		/// <summary>
-		/// Invoke middleware
-		/// </summary>
-		public Task InvokeAsync(HttpContext context)
+		Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```
vlad-ion, line 17:
```suggestion
		private Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
```

aszczepanski:
It seems that it's not following the recommended way of passing middleware dependencies, as the dependencies should be passed in constructors: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/write?view=aspnetcore-3.1#middleware-dependencies

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
```diff
@@ -31,6 +28,8 @@ private Task SetResponseHeaders(object state)
 			return Task.CompletedTask;
 		}
 
+
+
```
vlad-ion, line 32:
```suggestion
```

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
```diff
@@ -39,11 +39,11 @@ Task RequestDelegate(HttpContext context)
 			}
 
 			HttpContext context = new DefaultHttpContext();
-			ActivityEnrichmentMiddleware middleware = new ActivityEnrichmentMiddleware(RequestDelegate);
+			IMiddleware middleware = new ActivityEnrichmentMiddleware();
 
 			Activity activity = new Activity("TestName").Start();
 
-			await middleware.InvokeAsync(context);
```
vlad-ion, line 46: Does it make sense to have ConfigureAwait(false) here?

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
```diff
@@ -61,8 +61,8 @@ static Task RequestDelegate(HttpContext context)
 			}
 
 			HttpContext context = new DefaultHttpContext();
-			ActivityEnrichmentMiddleware middleware = new ActivityEnrichmentMiddleware(RequestDelegate);
-			await middleware.InvokeAsync(context);
+			IMiddleware middleware = new ActivityEnrichmentMiddleware();
+			await middleware.InvokeAsync(context, RequestDelegate);
```
vlad-ion, line 65: Same, configureawait?

tests/Extensions/Hosting.Services.Web.UnitTests/ListenerValidator.cs
```diff
@@ -70,6 +71,13 @@ public void ValidateOmexTypesRegistered(IWebHost host)
 			m_logsEventSourceMock.Invocations.Clear();
 			logger.LogError("TestMessage");
 			Assert.AreNotEqual(0, m_logsEventSourceMock.Invocations.Count, "Omex logger should be registred in WebHost");
+
+			ResolveType<ActivityEnrichmentMiddleware>(host);
+			ResolveType<ResponseHeadersMiddleware>(host);
+
+#pragma warning disable CS0618 // Obsolete middlewares also should be resolvable
```
vlad-ion, line 78:
```suggestion
#pragma warning disable CS0618 // Obsolete middlewares should also be resolvable
```

tests/Extensions/Hosting.Services.Web.UnitTests/ResponseHeadersMiddlewareTests.cs
```diff
@@ -29,9 +29,9 @@ public async Task InvokeAsync_WithoutActivity_NotFailing()
 			ResponseFeatureForCallbacks feature = new ResponseFeatureForCallbacks();
 			HttpContext context = new DefaultHttpContext();
 			context.Features.Set<IHttpResponseFeature>(feature);
-			ResponseHeadersMiddleware middleware = new ResponseHeadersMiddleware(mock.Object);
+			IMiddleware middleware = new ResponseHeadersMiddleware(mock.Object);
 
-			await middleware.InvokeAsync(context);
+			await middleware.InvokeAsync(context, s => Task.CompletedTask);
```
vlad-ion, line 34: Same, configureawait? Although here you give it a completed task here so probably doesn't make any difference.

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
AndreyTretyak: I don't think it will build like that, explicit interface implementation doesn't allow access modifier (as in interface)

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
AndreyTretyak: Answer here https://github.com/microsoft/Omex/pull/184#discussion_r402215265

src/Extensions/Hosting.Services.Web/Middlewares/ResponseHeadersMiddleware.cs
AndreyTretyak: Answered here https://github.com/microsoft/Omex/pull/184#discussion_r402215265

tests/Extensions/Hosting.Services.Web.UnitTests/ResponseHeadersMiddlewareTests.cs
AndreyTretyak: Added, thanks

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
AndreyTretyak: Added, thank you

tests/Extensions/Hosting.Services.Web.UnitTests/ActivityEnrichmentMiddlewareTests.cs
AndreyTretyak: I don't think it will change anything but I'll added it everywhere, thank you

AndreyTretyak:
> It seems that it's not following the recommended way of passing middleware dependencies, as the dependencies should be passed in constructors: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/write?view=aspnetcore-3.1#middleware-dependencies

We are getting all of the dependencies from the constructor or `InvokeAsync` method so I think it's still following the recommended way of getting dependencies. I've found an approach with `IMiddleware` described here https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility?view=aspnetcore-3.1 and I'll try to do more investigation to understand how to make DI work without interface implementation but I don't see any way to do it from source code for now.

src/Extensions/Hosting.Services.Web/Middlewares/ActivityEnrichmentMiddleware.cs
vlad-ion: Interesting, ok then.

PR #185, opened 2020-03-31
ServiceFabric service strop should stop generic host lifetime

AndreyTretyak:
ServiceFabric service strop should stop generic host lifetime


AndreyTretyak:
> move around service instances without shutting down the hosting process. In general I think it should be up to SF to start and kill the processes.

The purpose of this change is to stop generic host when SF completes its task returned from ServiceRunner (the one that usual SF pattern awaits in `Main` method), so it to achieve the same behavior.
So this change giving more control to SF, I'm not sure that would be the problem here

artempushkin:
It seems like you'd be stopping the generic host right after the service registration, before it even had a chance to actually run.

AndreyTretyak:
> It seems like you'd be stopping the generic host right after the service registration, before it even had a chance to actually run.

Looks like I've misunderstood lifetime of `RegisterServiceAsync` task lifetime and it's controlling only service registration, not service runtime, so I'll close PR

PR #190, opened 2020-04-01
[FHL] Create aspect for timedscope wrapping

AndreyTretyak:
Attribute for wrapping method into  timedscopes


src/Extensions/Aspects/Microsoft.Omex.Extensions.Aspects.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>$(TargetFrameworksValue)</TargetFrameworks>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="MethodBoundaryAspect.Fody" Version="2.0.118" />
```
gikoning, line 6: Should this be in dependencies.props?

src/Extensions/Aspects/ActivityAttribute.cs
```diff
@@ -0,0 +1,131 @@
...
+		/// </summary>
+		public override void OnException(MethodExecutionArgs args)
+		{
+			TimedScope scope = (TimedScope)args.MethodExecutionTag;
+			HandleResult(scope, args.Exception);
+		}
+
+		private bool IsAwaitable(object obj, out Task? result)
+		{
+			result = null;
```
gikoning, line 80: Could move this to the parameter?

src/Extensions/Aspects/ActivityAttribute.cs
```diff
@@ -0,0 +1,131 @@
...
+			if (obj == null)
+			{
+				return false;
+			}
+
+			if (obj is Task task)
+			{
+				result = task;
+			}
+			else if (obj is ValueTask valueTask)
```
gikoning, line 90: I guess ValueTask is a subtask of Task? First check if obj is ValueTask before checking if it is a Task. Since both will return true

gikoning: Nevermind it isn't a subclass of Task

src/Extensions/Aspects/Microsoft.Omex.Extensions.Aspects.csproj
AndreyTretyak: Yes, it should, it's still a draft I'll be changing it couple times

vskh:
Suggesting to add some test with usecase demo.

src/Extensions/Aspects/Microsoft.Omex.Extensions.Aspects.csproj
AndreyTretyak: Replaced with version variables

AndreyTretyak:
> Suggesting to add some test with usecase demo.

Added some unit tests, still hope to remove the need in the xml in the root

dakanev-msft:
> using System;

License header missing

---
Refers to: tests/Extensions/Aspects.UnitTests/ActivityAttributeTests.cs:1 in af1d8c5. [](commit_id = af1d8c5ffee38b0a2d94ae4791cecd9fb44120ef, deletion_comment = False)

PR #194, opened 2020-04-03
Add simplified syntax of remoting listener creation

AndreyTretyak:
This will allow remoting listeners registration like this:
```csharp
builder.AddRemotingListener<MyService>("ServiceName", RemotingSettings.WithEndpoint("ResName"));
```


src/Extensions/Hosting.Services.Remoting/Remoting.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+	/// <summary>
+	/// Static methods to simplify creation of <see cref="FabricTransportRemotingListenerSettings" />
+	/// </summary>
+	public static class Remoting
+	{
+		/// <summary>
+		/// Creates <see cref="FabricTransportRemotingListenerSettings" /> with provided EndpointResourceName
+		/// </summary>
+		public static FabricTransportRemotingListenerSettings EndpointResourceName(string value)
```
muiriswoulfe, line 16: >EndpointResourceName [](start = 56, length = 20)

I'd suggest calling this `ToEndpointResourceName` for alignment with similar methods

muiriswoulfe, line 16: >value [](start = 84, length = 5)

You could make this an extension method. Similar designs have been used for similar code in the past, although the use of the `string` type here means this feels a little less beneficial that for more esoteric types

AndreyTretyak: I've changed name to 'RemotingSettings.WithEndpoint()'

AndreyTretyak: Yes I feel like `string` is too generic and with new name, it should be better as regular static method

PR #197, opened 2020-04-08
Add wrappers for remoting to attach activity

AndreyTretyak:
Add wrappers for remoting to attach activity.
This PR will add `OmexServiceProxyFactory.Instance` for remoting client.
And will integrate `OmexServiceRemotingDispatcher` into remoting listeners.
Besides that, it will add a library for adding and reading omex headers

PR #200, opened 2020-04-10
Remove handling of unrecoverable exceptions

AndreyTretyak:
Remove handling of unrecoverable exceptions
This is a follow up on @K-Cully comments to this PR https://github.com/microsoft/Omex/pull/197

### DIFF
--- a/src/Extensions/Abstractions/ExceptionHelpers.cs
+++ b/src/Extensions/Abstractions/ExceptionHelpers.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+
+namespace Microsoft.Omex.Extensions.Abstractions
+{
+	/// <summary>
+	/// Helper methods for <see cref="Exception" />
+	/// </summary>
+	public static class ExceptionHelpers
+	{
+		/// <summary>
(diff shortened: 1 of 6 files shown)


src/Extensions/Abstractions/ExceptionHelpers.cs
```diff
@@ -0,0 +1,21 @@
...
+	/// </summary>
+	public static class ExceptionHelpers
+	{
+		/// <summary>
+		/// Returns true if exception could be handled or logged and false for unrecoverable exeptions
+		/// like <see cref="OutOfMemoryException" /> and <see cref="StackOverflowException" />
+		/// </summary>
+		public static bool IsRecoverable(this Exception exception) =>
+			!(exception is OutOfMemoryException)
+			|| !(exception is StackOverflowException);
```
artempushkin, line 19: Starting with the .NET Framework 2.0, you can't catch a StackOverflowException object with a try/catch block, and the corresponding process is terminated by default (from https://docs.microsoft.com/en-us/dotnet/api/system.stackoverflowexception?redirectedfrom=MSDN&view=netframework-4.8)

artempushkin: Also see my comment in another PR about OutOfMemory.

AndreyTretyak:
According to references provided by @artempushkin I'll close this PR, feel free to reopen it if you have suggestions about how it should be handled differently

PR #207, opened 2020-04-17
Aggregation of small fixes from old PR

AndreyTretyak:
Aggregation of small fixes from old PR

### DIFF
--- a/src/Extensions/Abstractions/Accessors/Accessor.cs
+++ b/src/Extensions/Abstractions/Accessors/Accessor.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.Omex.Extensions.Hosting.Services
+namespace Microsoft.Omex.Extensions.Abstractions.Accessors
 {
 	/// <summary>
 	/// Type for accessing an object that is not available during dependency injection container creation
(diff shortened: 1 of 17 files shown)


src/Extensions/Abstractions/Accessors/Accessor.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 
-namespace Microsoft.Omex.Extensions.Hosting.Services
+namespace Microsoft.Omex.Extensions.Abstractions.Accessors
```
artempushkin, line 7: Maybe Extensions.DependencyInjection.Accessors? How many accessors do we have? Would simply Extensions.DependencyInjection work?

AndreyTretyak: I've put only classes required for Accessor initialization (ex. `Accessor` or `IAccessorSetter`) into this namespace to separate it from `IAccessor` class that needed to consume Accessor.

AndreyTretyak: We could put it into `Microsoft.Omex.Extensions.Abstractions.DependencyInjection ` although I'm not sure what else we would put there

AndreyTretyak: I think it's better to keep it separate

PR #212, opened 2020-05-18
Updating packages

aszczepanski:


Dependencies.props
```diff
@@ -29,25 +29,23 @@
...
-    <Microsoft_Extensions_Hosting_Version>3.1.3</Microsoft_Extensions_Hosting_Version>
-    <Microsoft_Extensions_Logging_Abstractions_Version>3.1.3</Microsoft_Extensions_Logging_Abstractions_Version>
-    <Microsoft_Extensions_DependencyInjection_Abstractions_Version>3.1.3</Microsoft_Extensions_DependencyInjection_Abstractions_Version>
-    <Microsoft_Extensions_Logging_Version>3.1.3</Microsoft_Extensions_Logging_Version>
-    <Microsoft_Extensions_DependencyInjection_Version>3.1.3</Microsoft_Extensions_DependencyInjection_Version>
-    <Microsoft_Extensions_Options_Version>3.1.3</Microsoft_Extensions_Options_Version>
-    <Microsoft_Extensions_Diagnostics_HealthChecks_Version>3.1.3</Microsoft_Extensions_Diagnostics_HealthChecks_Version>
-    <Microsoft_Extensions_Http_Version>3.1.3</Microsoft_Extensions_Http_Version>
-    <System_Threading_Tasks_Extensions_Version>4.5.3</System_Threading_Tasks_Extensions_Version>
+    <Microsoft_Extensions_Hosting_Abstractions_Version>3.1.4</Microsoft_Extensions_Hosting_Abstractions_Version>
```
muiriswoulfe, line 32: For packages like these and some of the ones above, I believe there would be merit in using a single variable to control the version of all of them, as they should all be updated in lockstep.

I think that's for another change though, so happy to sign off now.

PR #213, opened 2020-06-11
Add validation helpers

AndreyTretyak:
Add helpers methods for argument validation

### DIFF
--- a/src/Extensions/Abstractions/Validation.cs
+++ b/src/Extensions/Abstractions/Validation.cs
@@ -12,28 +12,6 @@ namespace Microsoft.Omex.Extensions.Abstractions
 	public static class Validation
 	{
 		/// <summary>
-		/// Throw exception if value is null, otherwise returns not-nullable value
-		/// </summary>
-		/// <typeparam name="T">type of the value</typeparam>
-		/// <param name="value">value to check</param>
-		/// <param name="name">name of the value for exception message</param>
-		/// <exception cref="ArgumentNullException">if value is null</exception>
-		[MethodImpl(MethodImplOptions.AggressiveInlining)]
-		public static T ThrowIfNull<T>(T? value, string? name = null) where T : class
-			=> value ?? throw new ArgumentNullException(name);
-
-		/// <summary>
-		/// Throw exception if value is null, otherwise returns not-nullable value
-		/// </summary>
-		/// <typeparam name="T">type of the value</typeparam>
-		/// <param name="value">value to check</param>
-		/// <param name="name">name of the value for exception message</param>
-		/// <exception cref="ArgumentNullException">if value is null</exception>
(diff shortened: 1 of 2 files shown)


src/Extensions/Abstractions/Validation.cs
```diff
@@ -0,0 +1,59 @@
...
+		/// <summary>
+		/// Throw exception if value is null, otherwise returns not-nullable value
+		/// </summary>
+		/// <typeparam name="T">type of the value</typeparam>
+		/// <param name="value">value to check</param>
+		/// <param name="name">name of the value for exception message</param>
+		/// <exception cref="ArgumentNullException">if value is null</exception>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static T ThrowIfNull<T>(T? value, string? name = null) where T : class
+			=> value ?? throw new ArgumentNullException(name);
```
K-Cully, line 23: what happens here if `name` is null`?

tests/Extensions/Abstractions.UnitTests/ValidationTests.cs
```diff
@@ -0,0 +1,54 @@
...
+		[TestMethod]
+		public void ThrowIfNullOrWhiteSpace_WhenValueValidRefType_ReturnsIt()
+		{
+			object expected = new object();
+			object actual = Validation.ThrowIfNull(expected);
+			Assert.AreEqual(expected, actual);
+		}
+
+		[TestMethod]
+		public void ThrowIfNullOrWhiteSpace_WhenValueValidStruct_ReturnsIt()
```
K-Cully, line 21: These two tests should be ThrowIfNull_...

src/Extensions/Abstractions/Validation.cs
AndreyTretyak: If the name is `null` message would be "Value cannot be null". https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQiAM7EGYAEA3uvqfogkSWTcABZgD2A7vlAKasCCYA5hAFt2UYADkIAGwkBRAB4BjdgAdgAS0ZQAFFEkSAlAG5qpAL7oTQA=
If the name is "test" message would be "Value cannot be null. (Parameter 'test')" https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQiAM7EGYAEA3uvqfogkSWTcABZgD2A7vlAKasCCYA5hAFt2UYADkIAGwkBRAB4BjdgAdgAS0ZQAFACJg7AM7BtASgDc1UgF90loA===

tests/Extensions/Abstractions.UnitTests/ValidationTests.cs
AndreyTretyak: Thank you, too much copy/paste

PR #219, opened 2020-06-24
Fix null tag name in ETW event

AndreyTretyak:
Fix `null` tag name in ETW event since null argument fails event

### DIFF
--- a/src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
+++ b/src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Omex.Extensions.Logging
 			string buildVersion = m_executionContext.BuildVersion;
 			string machineId = m_executionContext.MachineId;
 
-			string tagName = eventId.Name;
+			string tagName = eventId.Name ?? string.Empty;
 			// In case if tag created using Tag.Create (line number and file in description) it's better to display decimal number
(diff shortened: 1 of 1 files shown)


src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
```diff
@@ -41,7 +41,7 @@ public void LogMessage(Activity? activity, string category, LogLevel level, Even
 			string buildVersion = m_executionContext.BuildVersion;
 			string machineId = m_executionContext.MachineId;
 
-			string tagName = eventId.Name;
+			string tagName = eventId.Name ?? string.Empty;
```
K-Cully, line 44: Can eventId be null here too?

AndreyTretyak: It's a struct, so we safe there

K-Cully: as in should this be `eventId?.Name` and does the below code need to be updated to handle null eventId values also?

AndreyTretyak: `eventId` is of type [EventId](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.eventid?view=dotnet-plat-ext-3.1) that is readonly struct, so it could have `null` value, `default` value for it will be `EventId { Id: 0, Name: null }`.

PR #220, opened 2020-06-25
Improvements in logging classes for better adaptability

sebbe33:
In order to support more error details in `ILogEventSender` implementations, a change has been made to pass along the exception, if one was logged.

We've made `TagsExtensions` public, as it needs to be consumed from outside this project. The utils method in there was also converted into an extension method for `EventId` for a more object-oriented approach to consuming the functionality

### DIFF
--- a/src/Extensions/Logging/ILogEventSender.cs
+++ b/src/Extensions/Logging/ILogEventSender.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+using System;
 using System.Diagnostics;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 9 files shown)


src/Extensions/Logging/ILogEventSender.cs
```diff
@@ -30,6 +31,7 @@ public interface ILogEventSender
 		/// <param name="eventId">event Id</param>
 		/// <param name="threadId">Id of the thread</param>
 		/// <param name="message">Log message</param>
-		void LogMessage(Activity? activity, string category, LogLevel level, EventId eventId, int threadId, string message);
+		/// <param name="exception">The exception, if any, associated with the log.</param>
+		void LogMessage(Activity? activity, string category, LogLevel level, EventId eventId, int threadId, string message, Exception exception);
```
AndreyTretyak, line 35: I think `exception` should be marked as nullable her and in implementation, like `Exception? exception`, I'll also fix method in OmexLogger during formatter fix

sebbe33: Ah, right, forgot about that you turned off nullable reference types

PR #221, opened 2020-06-25
Fix logging of Exceptions

AndreyTretyak:
Fix logging of Exception since it's ignored by default formatter

### DIFF
--- a/src/Extensions/Logging/Internal/OmexLogger.cs
+++ b/src/Extensions/Logging/Internal/OmexLogger.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Omex.Extensions.Logging
 			string message = formatter(state, exception);
 			if (exception != null)
 			{
-				message = string.Concat(message, exception); // we need to concatenate with exception since default formatter ignore it https://github.com/aspnet/Logging/issues/442
+				message = string.Concat(message, Environment.NewLine, exception); // we need to concatenate with exception since default formatter ignore it https://github.com/aspnet/Logging/issues/442
 			}
 			
 			int threadId = Thread.CurrentThread.ManagedThreadId;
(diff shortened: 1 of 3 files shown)


src/Extensions/Logging/Internal/OmexLogger.cs
```diff
@@ -22,14 +22,19 @@ public OmexLogger(ILogEventSender logsEventSource, IExternalScopeProvider extern
...
 		{
 			if (!IsEnabled(logLevel))
 			{
 				return;
 			}
 
 			string message = formatter(state, exception);
+			if (exception != null)
+			{
+				message = string.Concat(message, exception); // we need to concatenate with exception since default formatter ignore it https://github.com/aspnet/Logging/issues/442
```
aszczepanski, line 35: Do we want to add a new line between `message` and `exception`?

AndreyTretyak: Yes, I think it makes sense, added, thanks

src/Extensions/Logging/Internal/OmexLogger.cs
```diff
@@ -22,14 +22,19 @@ public OmexLogger(ILogEventSender logsEventSource, IExternalScopeProvider extern
...
 		{
 			if (!IsEnabled(logLevel))
 			{
 				return;
 			}
 
 			string message = formatter(state, exception);
+			if (exception != null)
+			{
+				message = string.Concat(message, Environment.NewLine, exception); // we need to concatenate with exception since default formatter ignore it https://github.com/aspnet/Logging/issues/442
```
gijskoning, line 35:
```suggestion
				message = string.Concat(message, Environment.NewLine, exception); // We need to concatenate with exception since the default formatter ignores it https://github.com/aspnet/Logging/issues/442
```

PR #223, opened 2020-07-02
Add Certificate Reader

AndreyTretyak:
Add Certificate Reader


src/Extensions/Hosting/Certificates/CertificateInformation.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.Omex.Extensions.Hosting.Certificates
+{
+	internal readonly struct CertificateInformation
+	{
+		public readonly X509Certificate2 Certificate { get; }
```
dakanev-msft, line 10: >public [](start = 2, length = 6)

nit: spacing

src/Extensions/Hosting/Certificates/CertificateReader.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.Linq;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Extensions.Logging;
+using Microsoft.Omex.Extensions.Abstractions;
+
+namespace Microsoft.Omex.Extensions.Hosting.Certificates
+{
+	internal class CertificateReader : ICertificateReader
+	{
+		private readonly ICertificateStore m_certificateStore;
```
dakanev-msft, line 16: >private [](start = 2, length = 7)

Do we keep spacing between private member variables?

src/Extensions/Hosting/Certificates/CertificateReader.cs
```diff
@@ -0,0 +1,117 @@
...
+		{
+			if (refreshCache || !m_certificatesCache.TryGetValue(storeName, out CertificateInformation[] certificates))
+			{
+				m_logger.LogInformation(Tag.Create(), "Updating certificates cache for store '{0}'.", storeName);
+
+				certificates = m_certificateStore.GetAllCertificates(storeName, StoreLocation.LocalMachine).ToArray();
+
+				m_logger.LogInformation(Tag.Create(), "Inserting '{0}' certificates to certificates cache for store '{1}'.", certificates.Length, storeName);
+
+				m_certificatesCache[storeName] = certificates;
```
dakanev-msft, line 90: >m_certificatesCache [](start = 4, length = 19)

Probably not a practical concern, but the cache never expires old certs

src/Extensions/Hosting/Certificates/CertificateReader.cs
```diff
@@ -0,0 +1,117 @@
...
+			}
+
+			return certificates ?? Array.Empty<CertificateInformation>();
+		}
+
+		private static X509Certificate2? PickBestCertificate(IEnumerable<X509Certificate2> certificates)
+		{
+			DateTime now = DateTime.Now;
```
dakanev-msft, line 98: > DateTime.Now; [](start = 17, length = 14)

For testability, it is better to inject the time

src/Extensions/Hosting/Certificates/ICertificateReader.cs
```diff
@@ -0,0 +1,47 @@
...
+	/// </summary>
+	public interface ICertificateReader
+	{
+		/// <summary>
+		/// Returns all certificates which match commonName parameter
+		/// </summary>
+		/// <param name="commonName">Common name of certificates</param>
+		/// <param name="refreshCache">Refresh cache before looking for certificate</param>
+		/// <param name="storeName">Certificate store</param>
+		IEnumerable<X509Certificate2> GetCertificatesByCommonName(string commonName, bool refreshCache = false, StoreName storeName = StoreName.My);
```
dakanev-msft, line 21: >StoreName storeName = StoreName.My) [](start = 106, length = 35)

Would it be better to search in all stores if store is not specified? It will have a higher chance of finding the cert

src/Extensions/Hosting/Certificates/CertificateReader.cs
AndreyTretyak: I've considered it, but in this case, does not any useful functionality since you can build unit tests around DataTime.Now, but creating certificates with `NotBefore = DateTime.Now.AddDays(-1)` to get required state

src/Extensions/Hosting/Certificates/CertificateReader.cs
AndreyTretyak: I don't think we enforce empty lines between private members especially since they don't have comments, to me it makes sense to combine them without spaces, what do you think?

src/Extensions/Hosting/Certificates/CertificateInformation.cs
AndreyTretyak: Discussion moved here https://github.com/microsoft/Omex/pull/223#discussion_r448915415

src/Extensions/Hosting/Certificates/ICertificateReader.cs
AndreyTretyak: I don't think there is an option to specify something like `All` and just going thought all of them would probably be a bit overhead since we usually have them in a single store, but if user of a class needs it they could implement loop through all of it by themself

src/Extensions/Hosting/Certificates/CertificateReader.cs
AndreyTretyak: That's a valid concern, but considering the usual certificate expiry time, I think we could delay with adding this part and use parameter to refresh cache for cases that need it.

PR #226, opened 2020-07-10
Add both correlationId and activityId into timedscope event

AndreyTretyak:
Add activityId as an additional field to TimedScope events along size with old correlation. Also, fix logCategory passing for TimedScopes.

### DIFF
--- a/src/Extensions/TimedScopes/Internal/TimedScopeEventSource.cs
+++ b/src/Extensions/TimedScopes/Internal/TimedScopeEventSource.cs
@@ -17,11 +17,12 @@ namespace Microsoft.Omex.Extensions.TimedScopes
 			string metadata,
 			string userHash,
 			string serviceName,
+			string logCategory,
 			string result,
 			string correlationId,
 			string activityId,
 			long durationMs) =>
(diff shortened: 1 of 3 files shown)


src/Extensions/TimedScopes/Internal/TimedScopeEventSource.cs
```diff
@@ -17,27 +17,26 @@ internal sealed class TimedScopeEventSource : EventSource
...
 			string serviceName,
+			string logCategory,
 			string result,
 			string correlationId,
+			string activityId,
 			long durationMs) =>
-			WriteEvent((int)EventSourcesEventIds.LogTimedScope, name, subType, metadata, userHash, serviceName, m_logCategory, result, correlationId, durationMs);
+			WriteEvent((int)EventSourcesEventIds.LogTimedScope, name, subType, metadata, userHash, serviceName, logCategory, result, correlationId, activityId, durationMs);
 
 		[Event((int)EventSourcesEventIds.LogTimedScopeTestContext, Level = EventLevel.Informational, Version = 3)]
```
aszczepanski, line 27: You should bump version

AndreyTretyak: Thank you, good point, updated

PR #227, opened 2020-07-16
Add support for KestrelServerOptions

lexesjan:
Also added unit tests with the addition of KestrelServerOptions.

### DIFF
--- a/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
+++ b/src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web
 			string name,
 			ServiceFabricIntegrationOptions options,
 			Action<IWebHostBuilder>? builderExtension = null,
-			Action<KestrelServerOptions>? kestrelOptions = null)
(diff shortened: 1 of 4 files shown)


src/Extensions/Hosting.Services.Web/HostBuilderExtensions.cs
```diff
@@ -22,12 +23,14 @@ public static class HostBuilderExtensions
 			this ServiceFabricHostBuilder<OmexStatelessService, StatelessServiceContext> builder,
 			string name,
 			ServiceFabricIntegrationOptions options,
-			Action<IWebHostBuilder>? builderExtension = null)
+			Action<IWebHostBuilder>? builderExtension = null,
+			Action<KestrelServerOptions>? kestrelOptions = null)
```
AndreyTretyak, line 27: Probably, it's better to use the second overload with both parameters `Action<WebHostBuilderContext, KestrelServerOptions>` since it would provide more configurability and user won't need to specify the type in Activity so it won't be a lot of overhead.

lexesjan: Sounds good, changed Action<WebHostBuilderContext> to Action<WebHostBuilderContext, KestrelServerOptions>

PR #228, opened 2020-07-17
Fix Semmle warnings in unit tests

AndreyTretyak:
Fix Semmle warnings in unit tests

### DIFF
--- a/src/System/Extensions/XmlExtensions.cs
+++ b/src/System/Extensions/XmlExtensions.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Omex.System.Extensions
 				}
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
+			XmlReaderSettings settings = new XmlReaderSettings { XmlResolver = null };
 			settings.ConformanceLevel = ConformanceLevel.Document;
 			settings.IgnoreWhitespace = true;
 			settings.IgnoreComments = true;
--- a/src/System/Extensions/XmlExtensions.cs
+++ b/src/System/Extensions/XmlExtensions.cs
@@ -226,7 +226,7 @@ namespace Microsoft.Omex.System.Extensions
 				using (XmlReader schemaReader = XmlReader.Create(schemaStream))
 				{
 					schemas.Add(schemaNamespace, schemaReader);
-					XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
+					XmlReaderSettings settings = new XmlReaderSettings { XmlResolver = null };
 					settings.ValidationType = ValidationType.Schema;
 					settings.Schemas = schemas;
 
@@ -299,7 +299,7 @@ namespace Microsoft.Omex.System.Extensions
 				schemas.Add(schemaUri, reader);
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
+			XmlReaderSettings settings = new XmlReaderSettings { XmlResolver = null };
 			settings.ValidationType = ValidationType.Schema;
 			settings.Schemas = schemas;
 			return settings;
--- a/src/System/Extensions/XmlExtensions.cs
+++ b/src/System/Extensions/XmlExtensions.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Omex.System.Extensions
 				}
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings();
+			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
 			settings.ConformanceLevel = ConformanceLevel.Document;
 			settings.IgnoreWhitespace = true;
 			settings.IgnoreComments = true;
@@ -226,7 +226,7 @@ namespace Microsoft.Omex.System.Extensions
 				using (XmlReader schemaReader = XmlReader.Create(schemaStream))
 				{
 					schemas.Add(schemaNamespace, schemaReader);
-					XmlReaderSettings settings = new XmlReaderSettings();
+					XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
 					settings.ValidationType = ValidationType.Schema;
 					settings.Schemas = schemas;
 
@@ -299,7 +299,7 @@ namespace Microsoft.Omex.System.Extensions
 				schemas.Add(schemaUri, reader);
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings();
+			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
 			settings.ValidationType = ValidationType.Schema;
 			settings.Schemas = schemas;
 			return settings;
--- a/tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
+++ b/tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
@@ -127,7 +127,7 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks.UnitTests
 
 			Mock<ICodePackageActivationContext> mock = new Mock<ICodePackageActivationContext>();
 			mock.Setup(c => c.GetEndpoint(parameters.EndpointName)).Returns(new System.Fabric.Description.EndpointResourceDescription());
-			ServiceContext context = MockStatelessServiceContextFactory.Create(mock.Object, "", new Uri("http://localhost"), Guid.Empty, 0);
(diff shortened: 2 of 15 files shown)


src/System/Extensions/XmlExtensions.cs
```diff
@@ -117,7 +117,7 @@ public static bool IsWellFormedXml(this Stream stream, CompressionType compressi
 				}
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings();
+			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
```
muiriswoulfe, line 120: FYI, in these constructor calls (throughout the file), you can exclude the `()`

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
```diff
@@ -127,7 +127,7 @@ private async Task<(MockClient, HealthCheckResult)> RunHealthCheckAsync(HttpHeal
 
 			Mock<ICodePackageActivationContext> mock = new Mock<ICodePackageActivationContext>();
 			mock.Setup(c => c.GetEndpoint(parameters.EndpointName)).Returns(new System.Fabric.Description.EndpointResourceDescription());
-			ServiceContext context = MockStatelessServiceContextFactory.Create(mock.Object, "", new Uri("http://localhost"), Guid.Empty, 0);
+			ServiceContext context = MockStatelessServiceContextFactory.Create(mock.Object, "", new Uri("https://localhost"), Guid.Empty, 0);
```
muiriswoulfe, line 130: I realise you didn't add this, but it would be good to replace `""` with `string.Empty` as you're making this change given that the latter is more efficient

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpHealthCheckParametersTests.cs
```diff
@@ -76,7 +76,7 @@ public void Constructor_InvalidSheme_ThrowException()
 		[TestMethod]
 		public void Constructor_AbsoluteUri_ThrowException()
 		{
-			Assert.ThrowsException<ArgumentException>(() => Create(relatedUri: new Uri("http://localhost")));
+			Assert.ThrowsException<ArgumentException>(() => Create(relatedUri: new Uri("https://localhost"), scheme: Uri.UriSchemeHttps));
```
muiriswoulfe, line 79: Do you need to add the scheme given that the URL itself specifies `https`?

tests/Gating.UnitTests/GateContextUnitTests.cs
```diff
@@ -21,6 +21,11 @@ namespace Microsoft.Omex.Gating.UnitTests
 	/// </summary>
 	public sealed class GateContextUnitTests : UnitTestBase
 	{
+#pragma warning disable SM02239 // hard coded e-mail for unit tests
+		private static readonly string s_userIdentifierInGroup = "test1@microsoft.com";
```
muiriswoulfe, line 25: These can be made `const`, which is preferable for `string`s as they're more efficient than `static readonly`

tests/Gating.UnitTests/GateContextUnitTests.cs
```diff
@@ -21,6 +21,11 @@ namespace Microsoft.Omex.Gating.UnitTests
 	/// </summary>
 	public sealed class GateContextUnitTests : UnitTestBase
 	{
+#pragma warning disable SM02239 // hard coded e-mail for unit tests
```
muiriswoulfe, line 24: FYI, when I spoke to the Semmle team before, their advice was to create another build configuration that only builds product code (and not test code) and to only run this through Semmle in order to avoid false positives within unit tests.

The disadvantage of course is that as new projects are added, people can forget to add them to the new configuration and we lose Semmle coverage.

I'm not sure it's really worth doing here as you don't have many suppressions. It was different for another project we had where we had a lot of suppressions akin to this and having to always suppress them was annoying.

tests/Gating.UnitTests/GateContextUnitTests.cs
```diff
@@ -21,6 +21,11 @@ namespace Microsoft.Omex.Gating.UnitTests
...
+		private static readonly string s_userIdentifierInGroup = "test1@microsoft.com";
+		private static readonly string s_userIdentifierNotInGroup = "test1@apple.com";
```
muiriswoulfe, line 26: I'd suggest using @contoso.com, given that this is a Microsoft address.

AndreyTretyak: Updated, thank you

src/System/Extensions/XmlExtensions.cs
AndreyTretyak: Fixed, thanks

tests/Gating.UnitTests/GateContextUnitTests.cs
AndreyTretyak: Good point, thanks, replaced

src/System/Extensions/XmlExtensions.cs
```diff
@@ -226,7 +226,7 @@ public static bool ConformsToXmlSchema(this Stream xmlStream, Stream schemaStrea
 				using (XmlReader schemaReader = XmlReader.Create(schemaStream))
 				{
 					schemas.Add(schemaNamespace, schemaReader);
-					XmlReaderSettings settings = new XmlReaderSettings();
+					XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
```
muiriswoulfe, line 229: Can also remove the `()` here

src/System/Extensions/XmlExtensions.cs
```diff
@@ -299,7 +299,7 @@ public static XmlReaderSettings GetXmlReaderSettings(this Type type, string sche
 				schemas.Add(schemaUri, reader);
 			}
 
-			XmlReaderSettings settings = new XmlReaderSettings();
+			XmlReaderSettings settings = new XmlReaderSettings() { XmlResolver = null };
```
muiriswoulfe, line 302: Can also remove the `()` here

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
AndreyTretyak: Good point, replaced. 
Just in case if you are interested, I remember hearing something that JIT in netcore,  replaces calls to `""` with `String.Empty` , I was not able to find any exact mentions of it, but resulted JIT Asm looks quite simular in both cases https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMBmAAhzIGEyBvIsxi8zKMgWQAoBKOhpgZlwBOTgGU4AJwCWAOwDmAOgCiqAA5wAntwDc/AYyGiARMd36yAXyKWgA==

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpHealthCheckParametersTests.cs
AndreyTretyak: The method contract here is that it should receive relative url, so it won't have a scheme. This unit test checks that absolute url is an invalid parameter, but for some reason, I've wanted to keep other inputs totally valid, although it won't have any effect here.

tests/Gating.UnitTests/GateContextUnitTests.cs
AndreyTretyak: Thank you for explaining, I was wondering what would be Semmle team suggestions since they wanted to avoid suppressions. As you point out it this case we don't have a lot to suppress so I'll probably go with it to avoid supporting another build configuration

src/System/Extensions/XmlExtensions.cs
AndreyTretyak: Sorry, I've forgotten about other cases

src/System/Extensions/XmlExtensions.cs
AndreyTretyak: Removed, thanks

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
muiriswoulfe: Thanks for the info. In older versions of .NET, there was definitely an issue as many called this out. I know there was a concerned effort in .NET Core to improve perf, so I guess this change was made as part of that. I still think making the change makes sense as my understanding is that we want this project to work in both .NET Framework and .NET Core

PR #231, opened 2020-07-25
Remove TimedScopeGen since we defining them manually

AndreyTretyak:
We've switched to creating TimedScopeDefinitions manually so TimedScopeGen could be deprecated now


gijskoning:
Will this break any project that uses the timedscopeGen?

AndreyTretyak:
> Will this break any project that uses the timedscopeGen?

No projects using this one, current project using TimedScopeGen from Omex repo

An alternative to deleting this one would be switching all projects to this one and delete the one in Omex. @vlad-ion how complicated do you think it will be?

vlad-ion:
If we are moving away from using a generator anyway then removing is fine. Updating the projects in the Omex repo is not worth it. If no projects that we port out of that repo will use a generator then we don't need this one and when everything is out of the Omex repo then we don't need that one anymore either.

vlad-ion (approved):

gikoning (approved):

aszczepanski (approved):

PR #233, opened 2020-08-11
Improve exception reporting

AndreyTretyak:
Log all exception accusing in the app after host start and move exception handling logic during the build


src/Extensions/Hosting/ExceptionLogging/ExceptionLoggingHostedService.cs
```diff
@@ -0,0 +1,38 @@
...
+
+		public Task StopAsync(CancellationToken cancellationToken)
+		{
+			AppDomain.CurrentDomain.FirstChanceException -= FirstChanceException;
+			return Task.CompletedTask;
+		}
+
+		private void FirstChanceException(object sender, FirstChanceExceptionEventArgs e)
+		{
+			m_logger.LogError(Tag.Create(), e.Exception, "Exception during application execution");
```
artempushkin, line 35: Have you actually tested it? I have a feeling it's not going to end well. You'll be logging EVERY exception, including the expected and handled ones within .NET.

AndreyTretyak: .net code if quite good in not throwing and catching exceptions instead of logic, but after testing it looks like SF still produces a lot of COM exceptions, so you are probably right. Although I still think it would be useful to know about those, for performance reasons, we could probably remove this for now.
I've changed the approach to use host wrapper, what do you think about this option?

artempushkin (approved):

PR #234, opened 2020-08-15
Bug fix, pass report data if healthcheck failed with Exception

AndreyTretyak:
Fix for a bug found by @artempushkin, currently, report data ignored in case of exception.

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	/// <summary>
+	/// Additional parameters to configure health check
+	/// </summary>
+	public class HealthCheckParameters
+	{
+		/// <summary>
+		/// Additional data to add to health check result
+		/// </summary>
+		public IReadOnlyDictionary<string, object> ReportData { get; }
+
+		/// <summary>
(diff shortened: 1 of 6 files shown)


vlad-ion (approved):

artempushkin (approved):

src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
```diff
@@ -0,0 +1,44 @@
...
+			{
+				ReportData = s_emptyDictionary;
+			}
+			else
+			{
+				// should be replaced by passing enumerable to constructor after we drop full framework support
+				Dictionary<string, object> dictionary = new Dictionary<string, object>(reportData.Length);
+				foreach (KeyValuePair<string, object> pair in reportData)
+				{
+					dictionary.Add(pair.Key, pair.Value);
```
K-Cully, line 34: This assumes every Key in the array of KeyValuePair objects is unique and could throw.

K-Cully (changes requested):

src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
K-Cully: When fixing this, please also add UT coverage for this case.

AndreyTretyak: What would be the proper behavior for you here?
Do you think we should override values with the same key?

AndreyTretyak: Discussed in a call we'll add a more descriptive error message here.

AndreyTretyak: The discussion moved here: https://github.com/microsoft/Omex/pull/237

PR #237, opened 2020-08-19
Wrap dictionary exceptions for HealthCheckParameters

AndreyTretyak:
Wrap dictionary exceptions for HealthCheckParameters

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/HealthCheckParameters.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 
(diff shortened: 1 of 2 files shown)


gikoning (approved):

K-Cully (approved):

AndreyTretyak:
I'm still having doubts about adding this code since without it error will look like this
```
System.ArgumentException: An item with the same key has already been added. Key: EscalationData
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException[T](T key)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at Microsoft.Omex.Extensions.Diagnostics.HealthChecks.HttpHealthCheckParameters..ctor(String endpointName, Uri relatedUri, HttpMethod method, String scheme, Nullable`1 expectedStatus, Func`3 additionalCheck, KeyValuePair`2[] reportData)
```
And it's still relatively easy to understand the reason of the error.
So taking in the account that it's a really rare use case I'm not sure that we should handle it specially and setting a bad precedent of overcomplicating code.

Does anybody have strong opinions about it?

gikoning:
I think it is a nice addition and you implemented it already. Thus I don't see why not to add it

PR #240, opened 2020-08-20
Remove netstandard2.1 and net472

AndreyTretyak:
Try to reduce targets
netstandard2.1 -> netcoreapp3.1
net472 -> netstandard2.0


src/Gating.Example/Microsoft.Omex.Gating.Example.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(NetCoreVersions)</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
```
AndreyTretyak, line 4: Write target explicitly to simplify the example

PR #243, opened 2020-08-26
Add middleware for user identity tracking

AndreyTretyak:


src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Level of complience for user indetity logging
+	/// </summary>
+	public enum UserIdentiyComlienceLevel
```
aszczepanski, line 9: typo - compliance

aszczepanski: in a few files

src/Extensions/Hosting.Services.Web/Middlewares/SaltProviders/RotatingSaltProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+		private readonly RandomNumberGenerator m_random;
+		private readonly IMemoryOwner<byte> m_currentSaltMemory;
+		private readonly ISystemClock m_systemClock;
+		private DateTimeOffset m_saltGenerationTime;
+
+		public RotatingSaltProvider(ISystemClock systemClock)
+		{
+			m_random = new RNGCryptoServiceProvider();
+			m_currentSaltMemory = MemoryPool<byte>.Shared.Rent(SaltLength);
+			m_saltGenerationTime = DateTime.MinValue;
```
aszczepanski, line 24: The actual salt provider should be synchronized across all nodes/regions, so that user hashes are the same for the same user in a specific time frame when they hit various VMs. Do I understand correctly that this will only work on a single VM?

aszczepanski (changes requested):

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Level of complience for user indetity logging
```
gikoning, line 7:
```suggestion
	/// Level of compliance for user identity logging
```

gikoning: Identity type in multiple places

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
```diff
@@ -0,0 +1,21 @@
...
+	/// </summary>
+	public enum UserIdentiyComlienceLevel
+	{
+		/// <summary>
+		/// User indentity would be changed after some period (usualy 40 hours not more then 48 hours) to avoid tracking of user information
+		/// </summary>
+		OrgId,
+
+		/// <summary>
+		/// Constant user indertity in logs that could be correlated until logs are stored
```
gikoning, line 17:
```suggestion
		/// Constant user identity in logs that could be correlated until logs are stored
```

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentiyMiddleware.cs
```diff
@@ -0,0 +1,90 @@
...
+			saltSpan.CopyTo(uidSpan.Slice(ipBytesWritten));
+
+			using IMemoryOwner<byte> hashMemoryOwner = MemoryPool<byte>.Shared.Rent(HashSize);
+			Span<byte> hashSpan = hashMemoryOwner.Memory.Span;
+			if (!m_hashAlgorithm.TryComputeHash(uidSpan, hashSpan, out int hashBytesWritten))
+			{
+				return string.Empty;
+			}
+
+			// TODO: Should be replaced by Convert.ToHexStringUpper after Net 5 release, https://github.com/dotnet/runtime/issues/17837#issuecomment-639018813
```
gikoning, line 80: I am not sure if you should replace it. At least only after you know every usage of this method is by Net 5 and I have the feeling that can take some time.
Why not use and if statement and implement it now? Such that both version stay supported.
``` 
#if NET5_0: 
#endif
```

AndreyTretyak: We are not building for net 5.0 now so it might be confusing, but we could add it when we will have it as a target

src/Extensions/Hosting.Services.Web/Middlewares/SaltProviders/RotatingSaltProvider.cs
AndreyTretyak: Yes, with current implementation created salt would be different on each machine (but then passed as part of the request in the case if service would need to make nested calls) since it's the easiest way of ensuring that information is not traceable back to the user.
In the future, we might add another implementation of salt provider that would call some special 
service or storage for proper salt value the same for all instances, but it's much more dangerous.

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
AndreyTretyak: Fixed, thank you

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComlienceLevel.cs
AndreyTretyak: Updated, thnaks

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentiyMiddleware.cs
AndreyTretyak: Added now

tests/Extensions/Hosting.Services.Web.UnitTests/TempUtLocation.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.Extensions.Internal;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+using Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.UnitTests
+{
+	[TestClass]
+	public class UserIdentiyMiddlewareTests
```
AndreyTretyak, line 21: Tests would be moved into appropriate files after merging of PR about web listener change (https://github.com/microsoft/Omex/pull/278), to avoid merge conflicts.

AndreyTretyak: Updated

src/Extensions/Hosting.Services.Web/Middlewares/SaltProviders/RotatingSaltProvider.cs
K-Cully: I think this approach is fine for getting this integrated, then update to a salt service later, if necessary.

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentiyMiddleware.cs
```diff
@@ -0,0 +1,93 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Adds user information if needed
+	/// </summary>
+	internal class UserIdentiyMiddleware : IMiddleware, IDisposable
```
K-Cully, line 19: Is the intention that we extend this internally to use identity information and that the public version only uses IPs?

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyMiddlewareOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+	/// <summary>
+	/// Options for UserIdentiyMiddleware
+	/// </summary>
+	public class UserIdentiyMiddlewareOptions
+	{
+		/// <summary>
+		/// Set required compliance level for logging user identity
+		/// </summary>
+		public UserIdentiyComplianceLevel LoggingComlience { get; set; } = UserIdentiyComplianceLevel.OrgId;
```
K-Cully, line 14: typo `LoggingCompliance`

tests/Extensions/Hosting.Services.Web.UnitTests/Middlewares/RotatingSaltProviderTests.cs
```diff
@@ -0,0 +1,40 @@
...
+			CollectionAssert.AreEqual(intialSalt, provider.GetSalt().ToArray());
+
+			systemClock
+				.SetupGet(c => c.UtcNow)
+				.Returns(DateTimeOffset.Now.AddHours(47));
+
+			CollectionAssert.AreNotEqual(intialSalt, provider.GetSalt().ToArray());
+		}
+	}
+}
```
K-Cully, line 40: Do you think it is worth creating a test that executes several provider.GetSalt() calls in parallel, during a salt rotation, to verify that it is thread-safe (i.e. there won't be different salts generated by each task)?

K-Cully (changes requested):

src/Extensions/Hosting.Services.Web/Middlewares/Options/UserIdentiyComplianceLevel.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+	/// <summary>
+	/// Level of compliance for user identity logging
+	/// </summary>
+	public enum UserIdentiyComplianceLevel
+	{
+		/// <summary>
+		/// User identity would be changed after some period (usually 40 hours not more then 48 hours) to avoid tracking of user information
+		/// </summary>
+		OrgId,
```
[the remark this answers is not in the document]
AndreyTretyak: Added clarification, thanks

src/Extensions/Hosting.Services.Web/Middlewares/SaltProviders/RotatingSaltProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+			m_systemClock = systemClock;
+		}
+
+		public Span<byte> GetSalt()
```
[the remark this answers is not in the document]
AndreyTretyak: Added logging and comments

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentiyMiddleware.cs
AndreyTretyak: My initial intension was that everybody would be using just IP, but I've abstracted it under `IUserIdentityProvider` so implementations could be replaced with some other identity if needed

tests/Extensions/Hosting.Services.Web.UnitTests/Middlewares/RotatingSaltProviderTests.cs
AndreyTretyak: Good point, I've added a test feel free to suggest improvements

K-Cully (approved):

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentiyMiddleware.cs
K-Cully: Good, an IP only identifies a device, not a user.
If a user is authenticated their user ID is a better identifier, since it follows them across devices.

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/Options/UserIdentiyComplianceLevel.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Level of compliance for user identity logging
+	/// </summary>
+	public enum UserIdentiyComplianceLevel
```
aszczepanski, line 9: I'm wondering if it's something that should be stored in our public GitHub repo or maybe in our GitHub Enterprise repo. What do we do in other similar cases when the implementation is specific to MS?

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/SaltProviders/RotatingSaltProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Buffers;
+using System.Security.Cryptography;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Logging;
+using Microsoft.Omex.Extensions.Abstractions;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	internal class RotatingSaltProvider : ISaltProvider
```
aszczepanski, line 13: You could add a comment to make it clear that this implementation will generate different salts (and hashes) on different VMs

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentiyMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
+{
+	/// <summary>
+	/// Adds user information if needed
+	/// </summary>
+	internal class UserIdentiyMiddleware : IMiddleware, IDisposable
```
aszczepanski, line 17: Consider renaming to `UserHashIdentityMiddleware` or something similar

src/Extensions/Hosting.Services.Web/ServiceCollectionExtensions.cs
```diff
@@ -15,7 +19,22 @@ public static class ServiceCollectionExtensions
 		/// </summary>
 		public static IServiceCollection AddOmexMiddleware(this IServiceCollection services)
 		{
+			services.TryAddSingleton<ISystemClock, SystemClock>();
+			services.TryAddSingleton<IUserIdentityProvider, IpBasedUserIdentityProvider>();
```
aszczepanski, line 23: I suppose we'll need a chain of identity providers here, or an identity provider with some fallback logic. It might be worth adding more specific interfaces like `IRotatingSaltProvider` or another way to use multiple providers from the beginning.

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/Options/UserIdentiyComplianceLevel.cs
AndreyTretyak: As far as I understand currently only two cases exist:
1. UserHash should not be traceable to the end-user after 48 hours
2. No required limitation on user hash traceability
And it's more legal than code requirements so I don't see a lot of other options appearing here. Essentially it's a switch to disable anonymizing of hashes if needed. 
And if we would get some new options I don't see problems adding them here.

src/Extensions/Hosting.Services.Web/ServiceCollectionExtensions.cs
AndreyTretyak: I've added the ability of chainging `IUserIdentityProvider`'s but I don't think it would be useful for salt providers

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentiyMiddleware.cs
AndreyTretyak: Renamed

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/SaltProviders/RotatingSaltProvider.cs
AndreyTretyak: Added comment, thanks

src/Extensions/Hosting.Services.Web/ServiceCollectionExtensions.cs
aszczepanski: Oh right, I meant `IIpBasedUserIdentityProvider`. But I suppose we can tackle it later when it's needed

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/Options/UserIdentiyComplianceLevel.cs
aszczepanski: Ok, makes sense. I was just wondering if LiveId/OrgId is something that should be included here, maybe nomenclature like DataController/DataProcessor should be used instead

aszczepanski (approved):

K-Cully (approved):

PR #244, opened 2020-08-28
Detect correlation Id in request header

ling1726:
Allow the obsolete correlation middleware to also detect correlation Ids set in the request header

### DIFF
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 				return;
 			}
 
-			Guid? oldCorrelation = ExtractCorrelationIdFromQuery(request) ?? ExtractCorrelationIdFromHeader(request);
+			Guid? oldCorrelation = ExtractCorrelationIdFromHeader(request) ?? ExtractCorrelationIdFromQuery(request);
 			if (oldCorrelation.HasValue)
 			{
 				activity.SetObsoleteCorrelationId(oldCorrelation.Value);
--- a/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 				return;
 			}
 
-			Guid? oldCorrelation = ExtractCorrelationId(request);
+			Guid? oldCorrelation = ExtractCorrelationIdFromQuery(request) ?? ExtractCorrelationIdFromHeader(request);
 			if (oldCorrelation.HasValue)
 			{
 				activity.SetObsoleteCorrelationId(oldCorrelation.Value);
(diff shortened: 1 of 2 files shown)


AndreyTretyak (dismissed):

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -89,6 +95,25 @@ private void ExtractCorrelationFromRequest(HttpRequest request)
 				? value.FirstOrDefault(s => !string.IsNullOrWhiteSpace(s))
 				: null;
 
+		private static string? ExtractHeader(IHeaderDictionary headers, IEnumerable<string> names)
+		{
+			foreach (string name in names)
+			{
+				string? value = ExtractHeader(headers, name);
```
csundlof, line 102: You can replace the ExtractHeader function with
`headers[name]`

When accessing a key that is not present in an IHeaderDictionary, StringValues.Empty is returned.

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -89,6 +95,25 @@ private void ExtractCorrelationFromRequest(HttpRequest request)
...
+				if (!string.IsNullOrWhiteSpace(value))
```
gijskoning, line 103: You are checking twice for nullorwhitespace. You could add an out parameter to the function and return true when it is not null?

gijskoning, line 103: You are checking twice for nullorwhitespace. You could add an out parameter to the function and return true when it is not null?
```suggestion
				bool notNull = ExtractHeader(headers, name, out string? value);
				if (notNull)
				{
```

tests/Extensions/Hosting.Services.Web.UnitTests/ObsoleteCorrelationHeadersMiddlewareTests.cs
```diff
@@ -0,0 +1,91 @@
...
+		[Obsolete]
+		[DataTestMethod]
+		[DataRow("MS-CorrelationId")]
+		[DataRow("ms-correlationid")]
+		[DataRow("corr")]
+		[DataRow("HTTP_X_CORRELATIONID")]
+		public async Task SetsObsoleteCorrelationIdFromHeaderToActivity(string correlationName)
+		{
+			// Arrange
+			Guid obsoleteCorrelation = new Guid();
```
csundlof, line 42: Using new Guid() will create a default (all 0) GUID. In order to make ensure that you don't always return the default GUID it would make sense to test using a real value here. Could be hardcoded, or as below
```suggestion
			Guid obsoleteCorrelation = Guid.NewGuid();
```

tests/Extensions/Hosting.Services.Web.UnitTests/ObsoleteCorrelationHeadersMiddlewareTests.cs
```diff
@@ -0,0 +1,91 @@
...
+		[Obsolete]
+		[DataTestMethod]
+		[DataRow("MS-CorrelationId")]
+		[DataRow("ms-correlationid")]
+		[DataRow("corr")]
+		[DataRow("HTTP_X_CORRELATIONID")]
+		public async Task SetsObsoleteCorrelationIdFromHeaderToHeader(string correlationName)
+		{
+			// Arrange
+			Guid obsoleteCorrelation = new Guid();
```
csundlof, line 61: Same point regarding GUID usage here.

csundlof (changes requested):

src/Extensions/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
ling1726: Resolving in favour of another comment, the `ExtractHeader` function was never needed since IHeaderDictionary will return String.Empty if the key is not present.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.iheaderdictionary.item?view=aspnetcore-3.1#Microsoft_AspNetCore_Http_IHeaderDictionary_Item_System_String_

csundlof (approved):

AndreyTretyak (approved):

gikoning (approved):

PR #246, opened 2020-09-17
Attempt remove net472 as a targets for packages

AndreyTretyak:
Theoretically, we don't need net472 as a target anymore since it could use netstandard2.0 instead, we'll see how it will work :)

### DIFF
--- a/Dependencies.props
+++ b/Dependencies.props
@@ -5,25 +5,24 @@
     <Microsoft_AspNetCore_Http_Abstractions_Version>2.2.0</Microsoft_AspNetCore_Http_Abstractions_Version>
     <Microsoft_AspNetCore_Http_Version>2.2.2</Microsoft_AspNetCore_Http_Version>
     <Microsoft_Azure_DocumentDB_Version>2.11.0</Microsoft_Azure_DocumentDB_Version>
-    <Microsoft_Net_Test_Sdk_Version>16.6.1</Microsoft_Net_Test_Sdk_Version>
+    <Microsoft_Net_Test_Sdk_Version>16.7.1</Microsoft_Net_Test_Sdk_Version>
     <Microsoft_SourceLink_GitHub_Version>1.0.0</Microsoft_SourceLink_GitHub_Version>
     <Microsoft_SourceLink_AzureRepos_Git_Version>1.0.0</Microsoft_SourceLink_AzureRepos_Git_Version>
-    <MSTest_TestAdapter_Version>2.1.1</MSTest_TestAdapter_Version>
-    <MSTest_TestFramework_Version>2.1.1</MSTest_TestFramework_Version>
+    <MSTest_TestAdapter_Version>2.1.2</MSTest_TestAdapter_Version>
+    <MSTest_TestFramework_Version>2.1.2</MSTest_TestFramework_Version>
     <System_Configuration_ConfigurationManager_Version>4.7.0</System_Configuration_ConfigurationManager_Version>
     <System_ServiceModel_Primitives_Version>4.7.0</System_ServiceModel_Primitives_Version>
     <System_Threading_Tasks_Version>4.3.0</System_Threading_Tasks_Version>
-    <Xunit_Runner_VisualStudio_Version>2.4.2</Xunit_Runner_VisualStudio_Version>
+    <Xunit_Runner_VisualStudio_Version>2.4.3</Xunit_Runner_VisualStudio_Version>
+    <Moq_Version>4.14.5</Moq_Version>
   </PropertyGroup>
   <PropertyGroup Label="Package Versions Core. AutoUpdate">
     <Bond_Core_CSharp_Version_Core>9.0.0</Bond_Core_CSharp_Version_Core>
-    <Moq_Version_Core>4.14.3</Moq_Version_Core>
     <System_Collections_Immutable_Version_Core>1.7.1</System_Collections_Immutable_Version_Core>
   </PropertyGroup>
   <PropertyGroup Label="Package Versions. Pinned">
     <Bond_Core_CSharp_Version>6.0.0</Bond_Core_CSharp_Version>
     <Microsoft_Azure_DocumentDb_Version>2.5.1</Microsoft_Azure_DocumentDb_Version>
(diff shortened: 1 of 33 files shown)


muiriswoulfe (approved):

muiriswoulfe:
If you search the repo for "net472", you can see some other items related to this framework -- references in .props files and a very small code workaround to facilitate use of the framework. Both of these should be possible to remove now with this change.

AndreyTretyak:
> If you search the repo for "net472", you can see some other items related to this framework -- references in .props files and a very small code workaround to facilitate use of the framework. Both of these should be possible to remove now with this change.

Thank you for noticing that. Unfortunately, some of them are needed but was able to unify some of our targets

Dependencies.props
```diff
@@ -5,25 +5,24 @@
...
-    <MSTest_TestAdapter_Version>2.1.1</MSTest_TestAdapter_Version>
-    <MSTest_TestFramework_Version>2.1.1</MSTest_TestFramework_Version>
+    <MSTest_TestAdapter_Version>2.1.2</MSTest_TestAdapter_Version>
+    <MSTest_TestFramework_Version>2.1.2</MSTest_TestFramework_Version>
     <System_Configuration_ConfigurationManager_Version>4.7.0</System_Configuration_ConfigurationManager_Version>
     <System_ServiceModel_Primitives_Version>4.7.0</System_ServiceModel_Primitives_Version>
     <System_Threading_Tasks_Version>4.3.0</System_Threading_Tasks_Version>
-    <Xunit_Runner_VisualStudio_Version>2.4.2</Xunit_Runner_VisualStudio_Version>
+    <Xunit_Runner_VisualStudio_Version>2.4.3</Xunit_Runner_VisualStudio_Version>
+    <Moq_Version>4.14.5</Moq_Version>
```
muiriswoulfe, line 17: Consider sorting. I believe the list is otherwise sorted.

AndreyTretyak: Oh, missed that, thank you, sorted now

src/Extensions/Hosting.Services.Web/Microsoft.Omex.Extensions.Hosting.Services.Web.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(NetCoreVersion)</TargetFrameworks>
+    <TargetFrameworks>$(NetCoreVersions)</TargetFrameworks>
```
muiriswoulfe, line 3: Just to confirm -- is this not available for .NET Standard by design?

Directory.Build.props
```diff
@@ -3,11 +3,14 @@
...
     <NetStandardVersions>netstandard2.0</NetStandardVersions>
-    <NetFrameworkVersions>net472</NetFrameworkVersions> <!-- Framework version that should be targeted alongside with .net standard -->
-    <LibraryTargetFrameworks>$(NetCoreVersions);$(NetStandardVersions);$(NetFrameworkVersions)</LibraryTargetFrameworks>
-    <LibraryTargetFrameworks Condition="'$(OS)' != 'Windows_NT'">$(NetCoreVersions);$(NetStandardVersions)</LibraryTargetFrameworks>
+    <LibraryTargetFrameworks>$(NetCoreVersions);$(NetStandardVersions)</LibraryTargetFrameworks>
     <ExecutableTargetFrameworks>$(NetCoreVersions)</ExecutableTargetFrameworks>
-</PropertyGroup>
+  </PropertyGroup>
+  <PropertyGroup Label="UnitTest Targets">
+    <NetFameworkVersion>net472</NetFameworkVersion>
```
muiriswoulfe, line 10: Do we wish to retain .NET 4.7.2 here?

src/Extensions/Hosting.Services.Web/Microsoft.Omex.Extensions.Hosting.Services.Web.csproj
AndreyTretyak: Yes, Asp .Net Core only supports net core starting from version 3

Directory.Build.props
```diff
@@ -3,11 +3,14 @@
...
+    <UnitTestTargetValues>$(NetCoreVersions);$(NetFameworkVersion)</UnitTestTargetValues>
```
muiriswoulfe, line 11: Given that the non-unit test version of this is `LibraryTargetFrameworks`, I think `UnitTestTargetFrameworks` might be a better name here

tests/Extensions/Directory.Build.props
```diff
@@ -1,19 +1,14 @@
 <Project>
   <Import Project="$([MSBuild]::GetPathOfFileAbove('$(_DirectoryBuildPropsFile)', '$(MSBuildThisFileDirectory)../'))" />
   <PropertyGroup>
-    <NetFameworkVersion>net472</NetFameworkVersion>
-    <NetCoreVersion>netcoreapp3.1</NetCoreVersion>
-    <TargetFrameworksValue>$(NetFameworkVersion);$(NetCoreVersion)</TargetFrameworksValue>
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
-    <IsPackable>false</IsPackable>
-    <IsTestProject>true</IsTestProject>
```
muiriswoulfe, line 10: Do we definitely want to remove `IsTestProject`? It seems like something that would be beneficial, but I can't find much useful information about it on the web

muiriswoulfe (approved):
A few small comments

tests/Extensions/Directory.Build.props
AndreyTretyak: It's defined in the `tests/Directory.Build.props` that is also applied here

Directory.Build.props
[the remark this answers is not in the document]
AndreyTretyak: Thank you missed that one

Directory.Build.props
AndreyTretyak: Makes sense, thank you

muiriswoulfe (approved):

gikoning (approved):

PR #248, opened 2020-09-21
Delete projects that are out of the solution

AndreyTretyak:
Some project excluded from the solution and build, so we could delete them since all the required logic moved to extensions now


dyllew3 (approved):

artempushkin (approved):
Kill it with fire.

PR #249, opened 2020-09-28
Update System.Diagnostics to v5 to fix property setters 

AndreyTretyak:


dyllew3 (dismissed):

dyllew3 (approved):

src/Extensions/Abstractions/Activities/ActivityHacks.cs
```diff
@@ -0,0 +1,94 @@
...
+
+			s_valueField.SetValue(node, new KeyValuePair<string, string?>(key, value));
+			return true;
+		}
+
+		private static Type? s_listNodeType;
+		private static FieldInfo? s_baggageField;
+		private static FieldInfo? s_headField;
+		private static FieldInfo? s_valueField;
+		private static FieldInfo? s_nextField;
```
paolsson-msft, line 92: Maybe it's more readable to set these inline? 

```
private static FieldInfo? s_baggageField = activity.GetType().GetField("_baggage", BindingFlags.NonPublic | BindingFlags.Instance);

File: src/Extensions/Abstractions/Activities/ActivityHacks.cs [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+		{
+			s_baggageField ??= activity.GetType().GetField("_baggage", BindingFlags.NonPublic | BindingFlags.Instance);
+			if (s_baggageField == null)
+			{
+				return false;
+			}
+
+			object? baggage = s_baggageField.GetValue(activity);
+
+			s_headField ??= s_baggageField.FieldType.GetField("_first", BindingFlags.NonPublic | BindingFlags.Instance);
```
paolsson-msft, line 48: `s_headField` field name vs `"_first"` field name

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -45,12 +71,15 @@ public void SetObsoleteCorrelationId_SetsValue()
 		{
 			Activity activity1 = new Activity("CorrelationTest1");
 			Activity activity2 = new Activity("CorrelationTest2");
-			Guid correlation = Guid.NewGuid();
+			Guid intiallyCorrelation = Guid.NewGuid();
+			Guid udatedCorrelation = Guid.NewGuid();
```
paolsson-msft, line 75:
```suggestion
			Guid updatedCorrelation = Guid.NewGuid();
```

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -45,12 +71,15 @@ public void SetObsoleteCorrelationId_SetsValue()
 		{
 			Activity activity1 = new Activity("CorrelationTest1");
 			Activity activity2 = new Activity("CorrelationTest2");
-			Guid correlation = Guid.NewGuid();
+			Guid intiallyCorrelation = Guid.NewGuid();
```
paolsson-msft, line 74:
```suggestion
			Guid initialCorrelation = Guid.NewGuid();
```

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -59,12 +88,15 @@ public void SetObsoleteTransactionId_SetsValue()
...
 
-			activity1.SetObsoleteTransactionId(id);
+			activity1.SetObsoleteTransactionId(initialId);
+			activity1.SetObsoleteTransactionId(updatedId);
+			CheckThatKeyNotDuplicated(activity1.Baggage);
 
-			Assert.AreEqual(id, activity1.GetObsolteteTransactionId());
-			Assert.AreNotEqual(id, activity2.GetObsolteteTransactionId());
+			Assert.AreEqual(updatedId, activity1.GetObsolteteTransactionId());
+			Assert.AreNotEqual(updatedId, activity2.GetObsolteteTransactionId());
```
paolsson-msft, line 99: That `GetObsolteteTransactionId` has a little typo. Should be `GetObsoleteTransactionId`

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
```diff
@@ -13,18 +15,40 @@ namespace Microsoft.Omex.Extensions.TimedScopes.UnitTests
...
+
+			activty.Start();
+
+			Guid? guidAfterStart = activty.GetRootIdAsGuid();
+
+			Assert.IsTrue(guidAfterStart.HasValue);
+			Assert.AreEqual(activty.RootId, guidAfterStart.GetValueOrDefault().ToString().Replace("-", ""));
+		}
+
 		[TestMethod]
```
paolsson-msft, line 37: I'd suggest you keep some of the previous tests as-is, so that you test your behaviour in isolation.

Test one: You can set the has give no previous has.
Test two: You can replace previous hash given new hash.

paolsson-msft: Holds throughout the PR

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -95,14 +104,14 @@ public static class ActivityExtensions
...
 
 		private const string UserHashKey = "UserHash";
 		private const string HealthCheckMarkerKey = "HealthCheckMarker";
 		private const string HealthCheckMarkerValue = "true";
 		private const string ObsoleteCorrelationId = "ObsoleteCorrelationId";
 		private const string ObsoleteTransactionId = "ObsoleteTransactionId";
-		private const string CorrelationIdObsoleteMessage = "Please use Activity.Id for new services instead";
-		private const string TransactionIdObsoleteMessage = "Please use Activity.TraceId for new services instead";
+		private const string CorrelationIdObsoleteMessage = "Please use Activity.Id or Activity.GetRootIdAsGuid() for new services instead";
+		private const string TransactionIdObsoleteMessage = "Please use Activity.IsHealthCheck() or Activity.TraceId for new services instead";
```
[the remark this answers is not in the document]
AndreyTretyak: You are right, thanks

src/Extensions/Abstractions/Activities/ActivityHacks.cs
AndreyTretyak: The reason behind lazy initialization here is that this hacky options would be called only when Set called twice for the same key, we are woking on minimizing those cases, so ideally it would be never executed and fileds won't be initialized at all. Beside that only part of those fields could be initialized like that since some are of them are receved based on existing values.

src/Extensions/Abstractions/Activities/ActivityHacks.cs
AndreyTretyak: To me, `head` makes more sense since it's the begging of the LinkedList, I don't sure that renaming would add more clarity (ex. `s_firstField` looks confusing)

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Oh, thank you for noticing

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: I considered that initially but it feels like a lot of repeated tests

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Fixed

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Updated, thanks

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: Updated, thanks

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
paolsson-msft: Well, there's a difference in repeating yourself when writing tests (i.e. the test code is repeated); that can be refactored!

If you keep the initial tests as-is you are exercising a different pattern (can I set the property?) vs (can I overwrite a previous set property?)

If one of them breaks (overwriting a previous property) you have narrowed it down quickly to what behaviour is broken. This can definitely be helpful in this scenario when you are building on top of the internals of the System.Diagnostics.

Obviously not a deal-breaker, but something I see as good-practice.

paolsson-msft (approved):

src/Extensions/Abstractions/Activities/ActivityHacks.cs
paolsson-msft: Mainly thinking since it's 100% coupled to the internals.

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
AndreyTretyak: I'm sorry for poor explanation, I understand your point here. What I've ment there is that is that whne update behaivour breaks you would get a message like "Expected: 'updated value', but actual: `initial value`", in case if setter will breacks it would be: `Expected: 'updated value', but actual: 'null'`, so it's quite easy to understand what gone wrong and test catch both errors. What I've ment about duplication is that we would have those two tests:
```CSharp
		[TestMethod]
		public void SetMetadata_SetsValue()
		{
			Activity activity1 = new Activity("SetMetadataTest1");
			Activity activity2 = new Activity("SetMetadataTest2");
			string value = "some metadata";

			activity1.SetMetadata(value);
			CheckThatKeyNotDuplicated(activity1.TagObjects);

			Assert.AreEqual(value, activity1.GetTag(ActivityTagKeys.Metadata));
			Assert.IsNull(activity2.GetTag(ActivityTagKeys.Metadata));
		}

		[TestMethod]
		public void SetMetadata_UpdatesValue()
		{
			Activity activity1 = new Activity("UpdateMetadataTest1");
			Activity activity2 = new Activity("UpdateMetadataTest2");
			string initialValue = "initial metadata";
			string updatedValue = "updated metadata";

			activity1.SetMetadata(initialValue);
			activity1.SetMetadata(updatedValue);
			CheckThatKeyNotDuplicated(activity1.TagObjects);

			Assert.AreEqual(updatedValue, activity1.GetTag(ActivityTagKeys.Metadata));
			Assert.IsNull(activity2.GetTag(ActivityTagKeys.Metadata));
		}
```
So only clearity that we are adding is diffirent test name, and I'm not sure worth of a burden supporting both sets of test in the future (currently 7 pairs).

I've been looking into refactoring test to reduce duplication but so far it looks overcomplicated

AndreyTretyak: About generalizing code, I'm managed to make something that works, but what do you think about it readability? It would be definitely easier to add new setters and check patters to it
```CSharp
[DataTestMethod]
[DynamicData(nameof(GetSettersToCheck), DynamicDataSourceType.Method)]
public void SetMethod_UpdatesValue(
	bool shouldTransfer,
	Func<Activity, object> getter,
	Action<Activity, object> setter,
	Func<int, object> valueGenerator)
{
	object value = valueGenerator(0);

	Activity activity1 = new Activity("TopActivity").Start();

	setter(activity1, value);

	Activity activity2 = new Activity("NestedActivity").Start();

	AssertActivitiesPropertiesAndStop(activity1, activity2, value, shouldTransfer, getter);
}

[DataTestMethod]
[DynamicData(nameof(GetSettersToCheck), DynamicDataSourceType.Method)]
public void SetMethod_SetsValue(
	bool shouldTransfer,
	Func<Activity, object> getter,
	Action<Activity, object> setter,
	Func<int, object> valueGenerator)
{
	object initialValue = valueGenerator(0);
	object updatedValue = valueGenerator(1);

	Activity activity1 = new Activity("TopActivity").Start();

	setter(activity1, initialValue);
	setter(activity1, updatedValue);

	Activity activity2 = new Activity("NestedActivity").Start();

	AssertActivitiesPropertiesAndStop(activity1, activity2, updatedValue, shouldTransfer, getter);
}

private static void AssertActivitiesPropertiesAndStop(
	Activity activity1,
	Activity activity2,
	object value,
	bool shouldTransfer,
	Func<Activity, object> getter)
{
	AssertActivityProperty(activity1, value, getter);
	AssertActivityProperty(activity2, shouldTransfer ? value : null, getter);

	activity2.Stop();
	activity2.Stop();
}

private static void AssertActivityProperty(Activity activity, object? value, Func<Activity, object> getter)
{
	CheckThatKeyNotDuplicated(activity.Baggage);
	CheckThatKeyNotDuplicated(activity.TagObjects);
	Assert.AreEqual(value, getter(activity));
}

private static void CheckThatKeyNotDuplicated<T>(IEnumerable<KeyValuePair<string, T?>> collection)
	where T : class
{
	string[] keys = collection.GroupBy(c => c.Key).Where(g => g.Count() > 1).Select(g => g.Key).ToArray();
	Assert.IsFalse(keys.Any(), "Duplicated keys found: " + string.Join(",", keys));
}

public static IEnumerable<object[]> GetSettersToCheck()
{
	static object[] CreateTestData<T>(
		bool shouldTransfer,
		Func<Activity, T> getter,
		Action<Activity, T> setter,
		Func<int, T> valueGenerator)
			=> ConvertToNonGeneric(
				shouldTransfer,
				a => getter(a)!,
				(a, v) => setter(a, (T)v),
				i => valueGenerator(i)!);

	static object[] ConvertToNonGeneric(
		bool shouldTransfer,
		Func<Activity, object> getter,
		Action<Activity, object> setter,
		Func<int, object> valueGenerator)
			=> new object[] { shouldTransfer, getter, setter, valueGenerator };

	yield return CreateTestData(false, a => a.GetTag(ActivityTagKeys.SubType), (a, v) => a.SetSubType(v!), s => s + " subtype");
	yield return CreateTestData(false, a => a.GetTag(ActivityTagKeys.Metadata), (a,v) => a.SetMetadata(v!), s => s + " metadata");

	string[] timedScopeResults = new[]
	{
		TimedScopeResult.SystemError.ToString(),
		TimedScopeResult.ExpectedError.ToString(),
		TimedScopeResult.Success.ToString()
	};
	yield return CreateTestData(false, a => a.GetTag(ActivityTagKeys.Result), (a, v) => a.SetResult((TimedScopeResult)Enum.Parse(typeof(TimedScopeResult), v!)), s => timedScopeResults[s % timedScopeResults.Length]);

	yield return CreateTestData(true, a => a.GetUserHash(), (a, v) => a.SetUserHash(v), s => s + " userHash");
	yield return CreateTestData(true, a => a.IsHealthCheck(), (a, v) => a.MarkAsHealthCheck(), s => true);

#pragma warning disable CS0618
	yield return CreateTestData(true, a => a.GetObsoleteCorrelationId(), (a, v) => a.SetObsoleteCorrelationId(v.GetValueOrDefault()), s => Guid.NewGuid());
	yield return CreateTestData(true, a => a.GetObsoleteteTransactionId(), (a, v) => a.SetObsoleteTransactionId(v.GetValueOrDefault()), s => (uint)s);
#pragma warning restore CS0618
}
```

paolsson-msft: It's obviously nothing super important. I think I would go with your initial approach over this, to have the verbosity instead of less code :)

AndreyTretyak: Okay, thanks, I'll merge it like it is and try to improve it in the future

dyllew3 (approved):

tests/Extensions/Abstractions.UnitTests/Activities/ActivityExtensionsTests.cs
paolsson-msft: Hopefully it will go away in the future :P

PR #252, opened 2020-09-30
Fix error when reflection used for old version of Activity

AndreyTretyak:
Remove error on failed attempt to update baggage


dyllew3 (approved):

paolsson-msft (approved):

paolsson-msft:
Can we UT this?

AndreyTretyak:
At least for now, I'm not sure how to reliably unit test it, I think it requires couple services interacting to reproduce. In general, the whole reflection part could be done testable with some mocks, but it would require more work and I hope we would just remove it in the future.

PR #254, opened 2020-10-05
Dyllew initial logging

dyllew3:
Adding static logger to omex extensions


src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Omex.Extensions.Logging
+{
+	/// <summary>
+	/// InitStaticLogger is the logger to be used before the uls logger is set
```
AndreyTretyak, line 11: nit:We don't use ULS term in opensource since it could be confusing, maybe here we could replace it with something like "before proper ILogger from DI"

src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+	/// not to be used as main logger
+	/// </summary>
+	public sealed class InitStaticLogger
```
AndreyTretyak, line 14: Why is not class static? What do you think about the name `InitializationLogger`, and then a field with logger could be called `Instance`

src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+			ServiceInitializationEventSource eventSource = ServiceInitializationEventSource.Instance;
+			eventSource.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging, message, args);
+			s_logger.LogInformation(eventId, message, args);
+		}
+
+		/// <summary>
+		/// Set the logger category
+		/// </summary>
+		/// <param name="category">Category</param>
+		public static void SetLoggerCategory(string category)
```
AndreyTretyak, line 39: I don't think we need this functionality, It's fine to have a single category if people need to have categories they would have to use proper logger

src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+		}).CreateLogger("Initial-Logging");
+
+		/// <summary>
+		/// Log on success
+		/// </summary>
+		/// <param name="serviceNameForLogging">Service name for logging</param>
+		/// <param name="message">Message to log</param>
+		/// <param name="eventId">Event id</param>
+		/// <param name="args">Extra arguments to be logged</param>
+		public static void Succeed(string serviceNameForLogging, string message, EventId eventId, params object[] args)
```
AndreyTretyak, line 28: What do you think about having a standard message and only pass service name as a single parameter. Same goes for failure

src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+		{
+			ServiceInitializationEventSource eventSource = ServiceInitializationEventSource.Instance;
+			eventSource.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging, message, args);
```
AndreyTretyak, line 31: nit: I would say since it's a single usage you could use it without saving to a local variable, but it's up to you

src/Extensions/Logging/ServiceCollectionExtensions.cs
```diff
@@ -34,14 +34,38 @@ public static ILoggingBuilder AddOmexLogging(this ILoggingBuilder builder)
 			return builder;
 		}
 
+		/// <summary>
+		/// Load the initial logger
+		/// </summary>
+		/// <param name="builder">The extension method argument</param>
+		public static ILoggingBuilder LoadInitialLogger(this ILoggingBuilder builder)
```
AndreyTretyak, line 41: nit: I would prefer `LoadInitializationLogger`, but you are better selecting appropriate naming

src/Extensions/Logging/ServiceCollectionExtensions.cs
```diff
@@ -34,14 +34,38 @@ public static ILoggingBuilder AddOmexLogging(this ILoggingBuilder builder)
...
+		{
+			builder.AddConsole();
+			builder.Services.TryAddTransient<IServiceContext, EmptyServiceContext>();
+			builder.Services.TryAddTransient<IExecutionContext, BasicMachineInformation>();
+			builder.Services.TryAddTransient<IExternalScopeProvider, LoggerExternalScopeProvider>();
+			builder.Services.TryAddTransient<IActivityProvider, ReplayableActivityProvider>();
```
AndreyTretyak, line 47: I don't think we need `IActivityProvider` here, if we do then we should use `SimpleActivityProvider` instead

AndreyTretyak: Or alternatively, you could use `AddOmexLogger` instead to make sure that the same logic used in both cases

src/Extensions/Logging/ServiceCollectionExtensions.cs
```diff
@@ -34,14 +34,38 @@ public static ILoggingBuilder AddOmexLogging(this ILoggingBuilder builder)
...
+
+			builder.Services.TryAddSingleton(p => OmexLogEventSource.Instance);
+			builder.Services.TryAddTransient<ILogEventReplayer, OmexLogEventSender>();
+			builder.Services.TryAddTransient<ILogEventSender, OmexLogEventSender>();
+
+			builder.Services.TryAddEnumerable(ServiceDescriptor.Transient<IActivityStopObserver, ReplayableActivityStopObserver>());
```
AndreyTretyak, line 53: This type also not needed as far as I know

src/Extensions/Logging/ServiceInitializationEventSource.cs
```diff
@@ -58,12 +58,26 @@ public void LogHostFailed(string exception, string serviceType)
...
+		/// Log host build succeeded
+		/// </summary>
+		/// <param name="hostProcessId">Host process id</param>
+		/// <param name="serviceType">The service type</param>
+		/// <param name="message">The message to be logged</param>
+		/// <param name="args">Extra arguments</param>
 		[Event((int)EventSourcesEventIds.GenericHostBuildSucceeded, Level = EventLevel.Informational, Message = "{2}", Version = 1)]
-		private void LogHostBuildSucceeded(int hostProcessId, string serviceType, string message) =>
-			WriteEvent((int)EventSourcesEventIds.GenericHostBuildSucceeded, hostProcessId, serviceType, message);
+		public void LogHostBuildSucceeded(int hostProcessId, string serviceType, string message, params object[] args) =>
```
AndreyTretyak, line 69: I don't think EventSource will work with `params` and `object` type and also I think it's not needed here

src/Extensions/Logging/ServiceInitializationEventSource.cs
```diff
@@ -5,13 +5,13 @@
...
 
-namespace Microsoft.Omex.Extensions.Hosting.Services
+namespace Microsoft.Omex.Extensions.Logging
 {
 	/// <summary>
 	/// Service Fabric event source
 	/// </summary>
 	[EventSource(Name = "Microsoft-OMEX-HostLogs")] //TODO: new event source should be registred GitHub Issue #187
-	internal sealed class ServiceInitializationEventSource : EventSource
+	public sealed class ServiceInitializationEventSource : EventSource
```
AndreyTretyak, line 14: I don't think it needs to be public

src/Extensions/Logging/ServiceInitializationEventSource.cs
AndreyTretyak: The same goes for second event

tests/Extensions/Logging.UnitTests/ServiceCollectionTests.cs
```diff
@@ -48,6 +49,17 @@ public void AddOmexLoggerOnLogBuilder_RegistersLogger()
...
+			ILoggerFactory factory = LoggerFactory.Create(builder => {
+				builder.LoadInitialLogger();
+			});
+
+			ILogger logger = factory.CreateLogger<OmexLogger>();
+			Assert.IsNotNull(logger);
+		}
+
 		private T ValidateTypeRegistration<T>(IServiceCollection collection)
 			where T : class
```
AndreyTretyak, line 64: Here a sample of checking that events reported to event source https://github.com/microsoft/Omex/blob/master/tests/Extensions/Logging.UnitTests/OmexLogEventSenderTests.cs
But we would need to check that running in pararel with other test won't interfere (since source is static) so maybe you could just extend existing test to check that initial logger also createas events

src/Extensions/Logging/InitStaticLogger.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.Omex.Extensions.Logging
+{
+	/// <summary>
+	/// InitStaticLogger is the logger to be used before the uls logger is set
+	/// not to be used as main logger
+	/// </summary>
+	public sealed class InitStaticLogger
+	{
+		private static ILogger s_logger = LoggerFactory.Create(builder =>
```
AndreyTretyak, line 16: We would probably need to make logger field publicly (like `Instance` property) so people could use it before generic host created, for example in configuration provider. Alternatively, we could have another method to report those events

AndreyTretyak: Actually, what do you think about this API
```CSharp
public static InitializationLogger
{
     static void InitializationSucceded(string serviceName)
     static void InitializationEvent(string message);
     static void InitializationFailure(Exception? exception = null, string message = "");
}
```

src/Extensions/Logging/ServiceInitializationEventSource.cs
dyllew3: This needs to be public so that this package and hosting can use it

src/Extensions/Logging/InitStaticLogger.cs
dyllew3: Thinks its good the only change I would make though is add optional message param for succeed

src/Extensions/Logging/ServiceInitializationEventSource.cs
AndreyTretyak: Could we use InitializationLogger in Hosting instead? It looks like it's proper usage

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+		 {
+			 builder.LoadInitialisationLogger();
+		 }).CreateLogger("Initial-Logging");
+
+		/// <summary>
+		/// Log on success
+		/// </summary>
+		/// <param name="serviceNameForLogging">Service name for logging</param>
+		/// <param name="message">Message to log</param>
+		public static void InitilisationSucceed(string serviceNameForLogging, string message = "")
```
AndreyTretyak, line 29: Sorry if I'm wrong here I'm not great with English, but should not it be `Initialization`?

dyllew3: Oh yeah went with english spelling, thats the american can switch it to american version

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+
+		/// <summary>
+		/// Log on failure
+		/// </summary>
+		/// <param name="serviceNameForLogging">Service name for logging</param>
+		/// <param name="ex">Exception to log</param>
+		/// <param name="message">Message to log</param>
+		public static void InitilisationFail(string serviceNameForLogging,  Exception? ex = null, string message = "")
+		{
+			ServiceInitializationEventSource.Instance.LogHostFailed(ex?.Message ?? string.Empty, serviceNameForLogging, message);
```
AndreyTretyak, line 45: it should be `ex?.ToSting()` instead, since `ex?.Message` won't include the stack trace

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+			Instance.LogInformation(ex, message);
```
AndreyTretyak, line 46: I think for the failure we should log it as error

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+
+		/// <summary>
+		/// Log on success
+		/// </summary>
+		/// <param name="serviceNameForLogging">Service name for logging</param>
+		/// <param name="message">Message to log</param>
+		public static void InitilisationSucceed(string serviceNameForLogging, string message = "")
+		{
+			string logMessage = string.IsNullOrWhiteSpace(message) ?
+				$"Initilisation logger logging success to {serviceNameForLogging}" : message ;
```
AndreyTretyak, line 32: In my mind success what about successful initialization of the service(or host), not a logger, maybe we could adjust the message into something like "Initialization successful for {serviceNameForLogging}. {message}"

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging, logMessage);
+			Instance.LogInformation(logMessage);
```
AndreyTretyak, line 34: Consider using `Instance.LogInformation(Tag.Create(), logMessage);` so we would have tags, here and in failure method

src/Extensions/Logging/ServiceCollectionExtensions.cs
```diff
@@ -34,6 +34,17 @@ public static ILoggingBuilder AddOmexLogging(this ILoggingBuilder builder)
 			return builder;
 		}
 
+		/// <summary>
+		/// Load the initial logger
+		/// </summary>
+		/// <param name="builder">The extension method argument</param>
+		public static ILoggingBuilder LoadInitialisationLogger(this ILoggingBuilder builder)
```
AndreyTretyak, line 41: I think this one could be made `internal` instead of `public`, or even moved into `private` method in the Initialization logger like `ILogger CreateInitialisationLogger()`

src/Extensions/Logging/InitialisationLogger.cs
```diff
@@ -0,0 +1,49 @@
...
+		/// <summary>
+		/// Instance of logger
+		/// </summary>
+		public static ILogger Instance { get; private set; } = LoggerFactory.Create(builder =>
+		 {
+			 builder.LoadInitialisationLogger();
+		 }).CreateLogger("Initial-Logging");
+
+		/// <summary>
+		/// Log on success
```
AndreyTretyak, line 25: I think we would need to clarify usage a bit, my idea was that this method would be used to indicate that service was successfully started (although I'm not sure its usefulness since generic host also does that, but that's something we used to do).

src/Extensions/Logging/Internal/ServiceInitializationEventSource.cs
```diff
@@ -23,47 +23,67 @@ internal sealed class ServiceInitializationEventSource : EventSource
...
 
+		/// <summary>
+		/// Log host build succeeded
+		/// </summary>
+		/// <param name="hostProcessId">Host process id</param>
+		/// <param name="serviceType">The service type</param>
+		/// <param name="message">The message to be logged</param>
 		[Event((int)EventSourcesEventIds.GenericHostBuildSucceeded, Level = EventLevel.Informational, Message = "{2}", Version = 1)]
-		private void LogHostBuildSucceeded(int hostProcessId, string serviceType, string message) =>
+		private void LogHostBuildSucceededInternal(int hostProcessId, string serviceType, string message) =>
```
AndreyTretyak, line 76: I'm a bit worried about the need of remaining private methods, but as far as I know, it's not the part of the contract so it should be fine

AndreyTretyak (dismissed):
Look good, thank you

src/Extensions/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -107,13 +107,13 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 				// get proper application name from host
 				serviceName = host.Services.GetService<IHostEnvironment>().ApplicationName;
 
-				ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging);
+				InitializationLogger.InitilizationSucceed(serviceNameForLogging);
```
vlad-ion, line 110: Typo. I would even change the name of the function completely as it sounds weird the way it is. It sounds more like a bool value than an action.
```suggestion
				InitializationLogger.InitializationSucceed(serviceNameForLogging);
```

src/Extensions/Logging/InitializationLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+using System;
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+using Microsoft.Omex.Extensions.Abstractions;
+using Microsoft.Omex.Extensions.Logging;
+
+namespace Microsoft.Omex.Extensions.Logging
+{
+	/// <summary>
+	/// InitStaticLogger is the logger to be used before the proper ILogger from DI is set
```
vlad-ion, line 13:
```suggestion
	/// InitStaticLogger is the logger to be used before the proper ILogger from DI is set.
```

src/Extensions/Logging/InitializationLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+	/// not to be used as main logger
```
vlad-ion, line 14:
```suggestion
	/// Not to be used as main logger.
```

src/Extensions/Logging/InitializationLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+	/// </summary>
+	public static class InitializationLogger
+	{
+		/// <summary>
+		/// Instance of logger
+		/// </summary>
+		public static ILogger Instance { get; private set; } = LoggerFactory.Create(builder =>
+		 {
+			 builder.LoadInitializationLogger();
```
vlad-ion, line 23:
```suggestion
			builder.LoadInitializationLogger();
```

tests/Extensions/Logging.UnitTests/InitializationLoggerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+			listener.EnableEvents(ServiceInitializationEventSource.Instance, eventLevel);
+
+			string message = "Test message";
+			string category = "Test category";
+			Activity activity = new Activity("Test activity");
+			activity.Start().Stop(); // start and stop activity to get correlation id
+
+			InitializationLogger.InitilizationSucceed(category, message);
+
+
```
vlad-ion, line 31:
```suggestion
```

tests/Extensions/Logging.UnitTests/InitializationLoggerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+			EventWrittenEventArgs eventInfo = listener.EventsInformation.Single(e => e.EventId == (int)EventSourcesEventIds.GenericHostBuildSucceeded);
+
+			AssertPayload(eventInfo, "message", "Initilization successful for Test category, Test message");
+
+			string newMessage = "New message";
+			InitializationLogger.InitilizationFail(category, new Exception("Not expected to be part of the event"), newMessage);
+
+			eventInfo = listener.EventsInformation.Single(e => e.EventId == (int)EventSourcesEventIds.GenericHostFailed);
+			AssertPayload(eventInfo, "message", newMessage);
+
```
vlad-ion, line 41:
```suggestion
```

src/Extensions/Logging/InitializationLogger.cs
AndreyTretyak: Do we require a dot at the end of the comment?

vlad-ion (approved):

AndreyTretyak (approved):

AndreyTretyak (approved):

vlad-ion (approved):

PR #257, opened 2020-10-08
Fixed a bug in HttpClient handler configuration

artempushkin:
ConfigurePrimaryHttpMessageHandler must return a new instance of HttpClientHandler every time, otherwise you hit an ObjectDisposed exception. See https://github.com/dotnet/runtime/issues/27610 for an example of the issue.


AndreyTretyak (approved):
Thank you for the fix!

dyllew3 (approved):

PR #262, opened 2020-10-13
Preserve baggage order during remoting

AndreyTretyak:
Mirroring baggage transfer fix from Kestrel into our remoting  wrapper


src/Extensions/Services.Remoting/OmexRemotingHeadersExtensions.cs
```diff
@@ -77,8 +77,13 @@ public static void ExtractActivityFromIncomingRequest(this IServiceRemotingReque
 
 				if (headers.TryGetHeaderValue(TraceStateHeaderName, out byte[] baggageBytes))
 				{
-					foreach (KeyValuePair<string, string> pair in DeserializeBaggage(baggageBytes))
+					KeyValuePair<string, string>[] baggage = DeserializeBaggage(baggageBytes);
+
+					// AddBaggage adds items at the beginning of the list, so we need to add them in reverse to keep the same order as the client
+					// An order could be important if baggage has two items with the same key (that is allowed by the contract)
+					for (int i = baggage.Length; i >= 0; i--)
```
paolsson-msft, line 84: `baggage.Length - 1`?

paolsson-msft (dismissed):

src/Extensions/Services.Remoting/OmexRemotingHeadersExtensions.cs
AndreyTretyak: Thank you, I wonder why UT not failing I would need to check them again

AndreyTretyak: Thank you for your comment it helped to uncover a combination of two issues and as a result, this code was never executed.
Fixed now, but we would have to wait until the release of  rtm3 of the package since the order is still wrong

dakanev-msft (approved):
:shipit:

dakanev-msft (approved):

artempushkin (approved):

Issue #260, opened 2020-10-09
Add Unit Tests to check that when service resolution impossible it will throw a proper exception


AndreyTretyak:
Add unit tests that will check the creation of Remote listener builders from src/Extensions/Hosting.Services.Remoting/HostBuilderExtensions.cs  should throw a proper exception when it's impossible to resolve Service type (not NullReference but about the failure to resolve type)

MatusTomko:
I want to do this issue.

MatusTomko:
thank you

AndreyTretyak:
@MatusTomko we got a pull request for this issue #263, how is your progress on it? Maybe you would like to work on another issue? (I'll try to add some more with `good first issue` tag)

MatusTomko:
@AndreyTretyak it will be good if you have another good first isssue for me. I think that issue #265 is good. Can you assigned to me?

AndreyTretyak:
> @AndreyTretyak Andrey Tretyak FTE it will be good if you have another good first isssue for me. I think that issue #265 is good. Can you assigned to me?

Sure, but you need to comment on that issue so I could assign it on you

PR #263, opened 2020-10-13
Add unit tests that document proper exception is thrown when service is not found #260

EpicGuy4000:


tests/Extensions/Hosting.Services.Remoting.UnitTests/HostBuilderExtensionsTests.cs
```diff
@@ -77,6 +79,44 @@ public void AddRemotingListener_ToStatelessServiceUsingIServiceType_RegisterType
...
+		[TestMethod]
+		public void AddRemotingListener_ToStatelessServiceUsingIServiceTypeThatIsNotRegistered_BuildThrowsInvalidOperationException()
+		{
+			string expectedName = nameof(AddRemotingListener_ToStatelessServiceUsingIServiceType_RegisterType);
+			FabricTransportRemotingListenerSettings settings = new FabricTransportRemotingListenerSettings();
+			HostBuilder builder = new HostBuilder();
+			ServiceFabricHostBuilder<OmexStatelessService, StatelessServiceContext> sfBuilder =
+				MockServiceFabricHostBuilder.CreateMockBuilder<OmexStatelessService, StatelessServiceContext>(builder);
+
+			sfBuilder.AddRemotingListener<MockService>(expectedName, settings);
```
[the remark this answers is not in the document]
EpicGuy4000: Thanks for the feedback Andrey.
I have updated the extension method to register the type, and have reflected the behavior in unit tests.

src/Extensions/Hosting/Certificates/CertificateReader.cs
```diff
@@ -79,7 +79,8 @@ public IEnumerable<X509Certificate2> GetCertificatesByCommonName(string commonNa
 
 		private CertificateInformation[] LoadCertificates(StoreName storeName, bool refreshCache)
 		{
-			if (refreshCache || !m_certificatesCache.TryGetValue(storeName, out CertificateInformation[]? certificates))
+			CertificateInformation[]? certificates;
```
AndreyTretyak, line 82: What was the reason behind this change?

EpicGuy4000: I was getting error "Converting null literal or possible null value to non-nullable type" in IDE and made the auto-fix from the list of actions. On second inspection (now) I believe the correct fix would be to switch back to in line out variable declaration but remove the '?' from the certificates type, as it does not match the signature of TryGetValue method.
Would you like me to apply it? Or rollback the fix altogether?

tests/Extensions/Hosting.Services.Remoting.UnitTests/MockServiceWithDependencies.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.Omex.Extensions.Hosting.Services.Remoting.UnitTests
+{
+	internal class MockServiceWithDependencies : IMockService
+	{
+		private readonly IMockServiceDependency m_dependency;
+
+		public MockServiceWithDependencies(IMockServiceDependency dependency) =>
+			m_dependency = dependency;
+	}
+}
```
AndreyTretyak, line 10: Add an empty line at the end of the file
```suggestion
}

```

src/Extensions/Hosting/Certificates/CertificateReader.cs
AndreyTretyak: Strange it does not show any warning to me, what IDE are you using? It's probably something to do with type marking as nullable, logically if the item is absent then it would return null. 
I would suggest removing this change since it's unrelated to the issue

AndreyTretyak (approved):
Thank you for your contribution. It looks good. Please fix minor comments and we could merge it.

src/Extensions/Hosting/Certificates/CertificateReader.cs
EpicGuy4000: I am using JetBrains Rider.
I will remove the change

AndreyTretyak: Okay, that explains the difference. 
I've checked the option of marking the type as non-nullable and as expected it causes a warning in VS.

So it's probably better to apply your change since that would work in both VS and Rider without warnings to avoid future problems for contributors using any of those.

AndreyTretyak (dismissed):

vlad-ion:
Should the new files have the copyright header?

AndreyTretyak:
> Should the new files have the copyright header? @AndreyTretyak

Yes, totally missed that, @EpicGuy4000  Could you please add them into new files that you've created or put classes in a single file alternatively (sorry, we have not merged .editorconfig change to add it automatically)

EpicGuy4000:
> > Should the new files have the copyright header? @AndreyTretyak
> 
> Yes, totally missed that, @EpicGuy4000 Could you please add them into new files that you've created or put classes in a single file alternatively (sorry, we have not merged .editorconfig change to add it automatically)

Added the copyright header to all new files.
Sorry for not adding them in the first place.
I look forward to seeing this repository evolve further 😄

AndreyTretyak (approved):

artempushkin (approved):
Seems good.

PR #267, opened 2020-10-19
Fix logic of adding remoting headers

AndreyTretyak:
Due to inversed condition, headers were not added during remoting transfer for a week now.
UT were ignoring this failure because they have not had `await` for a `Task`, also fixed in this PR


gikoning (approved):

AndreyTretyak:
Would be fixed, here https://github.com/microsoft/Omex/pull/262

PR #270, opened 2020-10-29
Switched SF health check publisher to StringBuilder pool.

artempushkin:
Each health check was creating a new StringBuilder on each publish. I siwtched to pooled StringBuilders to decrease allocations, that should give us some perf boost.

### DIFF
--- a/Dependencies.props
+++ b/Dependencies.props
@@ -22,18 +22,19 @@
     <Xunit_Core_Version>2.4.1</Xunit_Core_Version>
   </PropertyGroup>
   <PropertyGroup Label="Packages for Omex Extensions. AutoUpdate">
-    <Microsoft_Extensions_Hosting_Abstractions_Version>3.1.8</Microsoft_Extensions_Hosting_Abstractions_Version>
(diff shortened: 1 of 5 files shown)


AndreyTretyak (approved):
Nice, thank you!

AndreyTretyak (approved):

vlad-ion (approved):
It would be interesting to have a performance test to compare the gain here.

Also just wondering if there is a different mechanism like a "using" to get the builder from the pool so that you don't forget to return it.

artempushkin:
> It would be interesting to have a performance test to compare the gain here.
> 
> Also just wondering if there is a different mechanism like a "using" to get the builder from the pool so that you don't forget to return it.

I did run a few manual tests, it seems a bit faster, but as I understand the main benefit is fewer allocations, so less work for GC.

There's no "using" pattern for string builders, pool is using weak references internally I guess. The docs say that objects not returned to pool will be GCed, so forgetting to do it will not create a memory leak, but can affect performance somewhat.

PR #271, opened 2020-10-30
Respect failure status passed to health check from registration.

artempushkin:
Health checks must respect context.Registration.FailureStatus argument.

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
@@ -49,6 +49,7 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 			try
 			{
 				HealthCheckResult result = await CheckHealthInternalAsync(context, token).ConfigureAwait(false);
+				result = EnforceFailureStatus(context.Registration.FailureStatus, result);
 
 				scope.SetResult(TimedScopeResult.Success);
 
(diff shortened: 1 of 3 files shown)


AndreyTretyak (approved):

src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
```diff
@@ -49,6 +49,7 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
 			try
 			{
 				HealthCheckResult result = await CheckHealthInternalAsync(context, token).ConfigureAwait(false);
+				result = EnforceFailureStatus(context.Registration.FailureStatus, result);
```
vlad-ion, line 52: Is it worth having a new function for this that is only used here instead of using the else branch in that function directly here? Are there other places where this function will potentially be used?

vlad-ion (approved):

src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
artempushkin: It's a form of self-documentation. If a piece of code deserves a comment, it probably deserves a name, and here I certainly want to explain what's happening.

src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
```diff
@@ -57,13 +58,23 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
 		protected abstract Task<HealthCheckResult> CheckHealthInternalAsync(HealthCheckContext context, CancellationToken token);
+
+		private static HealthCheckResult EnforceFailureStatus(HealthStatus failureStatus, HealthCheckResult result)
+		{
+			if (result.Status == HealthStatus.Healthy)
+			{
+				return result;
+			}
+
+			return new HealthCheckResult(failureStatus, result.Description, result.Exception, result.Data);
```
AndreyTretyak, line 77: I think we should change logic a bit here.
First of all health checks with Error failure status could still report Degraded if they like
Second, we should not create a new struct if the status already correct.

artempushkin: No, I believe the intention behind health checks is to make them return binary yay-nay responses, but user can configure what kind of nay.
Second issue fixed in a follow-up PR.

PR #277, opened 2020-11-05
Update codeql-analysis.yml

AndreyTretyak:
Try fix for CodeQL out of memory mentioned here https://github.com/github/codeql-action/issues/245

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -24,11 +24,17 @@ jobs:
       matrix:
         # Override automatic language detection by changing the below list
         # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
-        language: ['csharp', 'javascript']
+        language: ['csharp']
         # Learn more...
         # https://docs.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#overriding-automatic-language-detection
 
     steps:
+    - name: configure Pagefile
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -24,11 +24,17 @@ jobs:
       matrix:
         # Override automatic language detection by changing the below list
         # Supported options are ['csharp', 'cpp', 'go', 'java', 'javascript', 'python']
-        language: ['csharp', 'javascript']
+        language: ['csharp']
```
muiriswoulfe, line 27: Are we OK not scanning JavaScript? I notice we have 3 JS files in our repo: <https://github.com/microsoft/Omex/search?l=javascript>

muiriswoulfe (approved):
Looks good. Thanks for getting this fixed.

.github/workflows/codeql-analysis.yml
AndreyTretyak: I've seen those but I'm not sure about their function I think it's some kind of samples for DocumentDb scripts and they might not play by the regular js rules. I think we could add js back at any moment, for now, I don't see a lot of usefulness in it.

muiriswoulfe: Makes sense. I think they're also unlikely to change over time.

It would be great though if we could find out what they are and document them so that others could use them. Not something for this PR though.

artempushkin (approved):

PR #283, opened 2020-11-18
Target .NET 5 + UT helpers

AndreyTretyak:
Add net5.0 as an additional target netcoreapp3.1 would be removed soon
Also created an additional project with UT helpers for example for `InNotNull` check with proper nullability attributes

### DIFF
--- a/src/Extensions/Testing.Helpers/NullableAssert.cs
+++ b/src/Extensions/Testing.Helpers/NullableAssert.cs
@@ -1,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Omex.Extensions.Testing.Helpers
+{
+	/// <summary>
+	/// Provider asserts checks with appropriate nullability attributes
+	/// </summary>
+	public static class NullableAssert
+	{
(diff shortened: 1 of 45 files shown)


src/Extensions/Testing.Helpers/NullableAssert.cs
```diff
@@ -0,0 +1,69 @@
...
+	/// <summary>
+	/// Provider asserts checks with appropriate nullability attributes
+	/// </summary>
+	public static class NullableAssert
+	{
+		/// <summary>
+		/// Tests whether the specified object is non-null and throws an exception if it is null
+		/// </summary>
+		public static void IsNotNull([NotNull] object? value, string message = "", params object[] parameters) =>
+#pragma warning disable CS8777 // Assert.IsNotNull would throw if object is null
```
gijskoning, line 18: Could use the ! symbol here?

src/Extensions/Testing.Helpers/NullableAssert.cs
```diff
@@ -0,0 +1,69 @@
...
+			Assert.IsNotNull(value, message, parameters);
```
gikoning, line 19: Can't you use the ! symbol here?

AndreyTretyak: not it's a different warning it's for the whole method due to attribute application

aszczepanski (approved):

dyllew3 (approved):

dimeehan (approved):

PR #285, opened 2020-11-30
Replace TimedScopes with Activity and ActivitySource

AndreyTretyak:
Replace `TimedScopes` with `Activity` and `ActivitySource`

`Microsoft.Omex.Extensions.TimedScopes` - won't be published anymore
`Microsoft.Omex.Extensions.Activities` - would contain ActivityListeners that would be sending metrics to log aggregator

Namespaces also renamed into `Activity` instead of `TimedScope` and for some extension method namespace changed to the namespace of the class that they extending

UTs are commented for now and would need to be updated

**Before:**
```CSharp
// Registration
serviceCollaction.AddTimedScopes(); // already added if you are using any of the Omex helpers to register service
...
// Usage 
// provider - ITimedScopeProvider from DI or SimpleTimedScopeProvider as default implementation
// definition - TimedScopeDefinition that usually defined in a separate static class 
using TimedScope scope = provider.CreateAndStart(definition, TimedScopeResult.SystemError)
     .AddSubType("Some subtype")
     .AddMetdata("Some methadata");
...
Activity.Current?.MarkAsUserError();
```
**After**
```CSharp
// Registration
serviceCollaction.AddOmexActivitySource();
...
// Usage 
// source - ActivitySource from DI or new ActivitySource(name: "Name", version: "1.0.0")
using Activity? scope = source.StartActivity("Scope name")
     ?.MarkAsSystemErrror()
     .AddSubType("Some subtype")
     .AddMetdata("Some methadata");
...
Activity.Current?.MarkAsUserError();
```


gikoning:
Will this pr make it more difficult to switch to the Extensions library?

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -0,0 +1,128 @@
...
+			string metaDataAsString = SanitizeString(metadata, nameof(metadata), name);
+			string userHashAsString = SanitizeString(userHash, nameof(userHash), name);
+			string serviceNameAsString = SanitizeString(serviceName, nameof(serviceName), name);
+			string correlationIdAsString = SanitizeString(correlationId, nameof(correlationId), name);
+			string activityIdAsString = SanitizeString(activityId, nameof(activityId), name);
+			long durationMsAsLong = Convert.ToInt64(durationMs, CultureInfo.InvariantCulture);
+
+			if (isHealthCheck)
+			{
+				m_eventSource.WriteTimedScopeTestEvent(
```
gikoning, line 80: This is still named timedscope?

AndreyTretyak:
> Will this pr make it more difficult to switch to the Extensions library?

Not by much for an external user it's mostly name change, but we would be using mostly native .net functionality, so our code would be much more understandable for others. Also, we should be able to create libraries without a lot of custom dependencies.

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
AndreyTretyak: Thanks, I'll check if it could be renamed with influencing ETW events

AndreyTretyak: Renamed as much as possible

src/Extensions/Activities/IActivityListenerConfigurator.cs
```diff
@@ -0,0 +1,28 @@
...
+	/// </summary>
+	public interface IActivityListenerConfigurator
+	{
+		/// <summary>
+		/// Decides if activity object events that were created using the activity source object should be listened or not
+		/// </summary>
+		bool ShouldListenTo(ActivitySource activity);
+
+		/// <summary>
+		/// Decides if creating System.Diagnostics.Activity objects with a specific data state is allowed
```
dakanev-msft, line 19: >Decides if creating System.Diagnostics.Activity objects with a specific data state is allowed [](start = 6, length = 93)

I would suggest to elaborate more on the difference between the last two methods, they have the same comments.

AndreyTretyak: Copied from official doc :) https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitylistener?view=net-5.0
I have vague suspicions on difference, I'll try to investigate and improve comments.

src/Extensions/Activities/Internal/ActivityListenerInitializerService.cs
```diff
@@ -0,0 +1,77 @@
...
+			IEnumerable<IActivityStartObserver> activityStartObservers,
+			IEnumerable<IActivityStopObserver> activityStopObservers,
+			IActivityListenerConfigurator configurator)
+		{
+			m_activityStartObservers = activityStartObservers.ToArray();
+			m_activityStopObservers = activityStopObservers.ToArray();
+			m_configurator = configurator;
+		}
+
+		public Task StartAsync(CancellationToken cancellationToken)
```
dakanev-msft, line 31: >cancellationToken [](start = 43, length = 17)

Does it make sense to check the cancellation token in the beginning?

src/Extensions/Activities/Internal/ActivityListenerInitializerService.cs
```diff
@@ -0,0 +1,77 @@
...
+			m_listener.Sample += Sample;
+			m_listener.SampleUsingParentId += SampleUsingParentId;
+			m_listener.ShouldListenTo += ShouldListenTo;
+
+			ActivitySource.AddActivityListener(m_listener);
+
+			return Task.CompletedTask;
+		}
+
+		public Task StopAsync(CancellationToken cancellationToken)
```
dakanev-msft, line 45: >public [](start = 2, length = 6)

Do we have comments for public methods in internal classes?

src/Extensions/Activities/Internal/DiagnosticsListener/DiagnosticsObserversInitializer.cs
```diff
@@ -14,13 +14,13 @@
...
 
-namespace Microsoft.Omex.Extensions.TimedScopes
+namespace Microsoft.Omex.Extensions.Activities
 {
 	/// <summary>
 	/// Hosted service for listening to diagnostic events
 	/// </summary>
 	/// <remarks>
-	/// Should be changed after .NET 5 release https://github.com/dotnet/designs/pull/98
+	/// Should be deleted after ASP .NET Core would support new approach
```
dakanev-msft, line 23: > Should be deleted after ASP .NET Core would support new approach [](start = 4, length = 65)

please create an Issue to track this

src/Extensions/Activities/Internal/EventSource/ActivityEventSource.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Diagnostics.Tracing;
+using Microsoft.Omex.Extensions.Abstractions.EventSources;
+
+namespace Microsoft.Omex.Extensions.Activities
+{
+	// Renamed from Microsoft-OMEX-TimedScopes to avoid conflict with sources in other libraries
+	[EventSource(Name = "Microsoft-OMEX-TimedScopes-Ext")]
```
dakanev-msft, line 10: >Microsoft-OMEX-TimedScopes-Ext [](start = 22, length = 30)

shouldn't we change the name not to have TimedScopes? Do we keep it for backward compatibility ?

src/Extensions/Compatibility/TimedScopes/TimedScope.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Diagnostics;
+using Microsoft.Omex.Extensions.Abstractions.Activities;
+
+namespace Microsoft.Omex.Extensions.Compatibility
+{
+	/// <summary>
+	/// Extension for string to start Activity
+	/// </summary>
+	public static class TimedScope
```
dakanev-msft, line 13: >TimedScope [](start = 21, length = 10)

What is the reason for keeping the name?

src/Extensions/Logging/Internal/EventSource/OmexLogEventSource.cs
```diff
@@ -8,7 +8,7 @@
 namespace Microsoft.Omex.Extensions.Logging
 {
 	// Renamed from Microsoft-OMEX-Logs to avoid conflict with sources in other libraries
-	[EventSource(Name = "Microsoft-OMEX-Logs-Ext")] //TODO: new event source should be registered GitHub Issue #187
+	[EventSource(Name = "Microsoft-OMEX-Logs-Ext")]
```
dakanev-msft, line 11: >Microsoft-OMEX-Logs-Ext [](start = 22, length = 23)

is this the correct name?

src/Extensions/Logging/Internal/Replayable/ActivityExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Omex.Extensions.Logging.Replayable;
+
+namespace Microsoft.Omex.Extensions.Logging.Internal.Replayable
+{
+	internal static class ReplayebleActivityExtensions
```
dakanev-msft, line 12: >Replayeble [](start = 23, length = 10)

nit: typo

src/Extensions/Logging/Internal/Replayable/ActivityExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+		public static void AddReplayLog(this Activity activity, LogMessageInformation logEvent, uint maxReplayedEventsPerActivity)
+		{
+			ConcurrentQueue<LogMessageInformation>? queue = activity.GetReplayableLogsInternal();
+			if (queue == null)
+			{
+				queue = new ConcurrentQueue<LogMessageInformation>();
+				activity.SetCustomProperty(ReplayableLogKey, queue);
+			}
+
```
dakanev-msft, line 30: Is this thread safe? Should we not use Interlocked.CompareExchange?

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
```diff
@@ -0,0 +1,47 @@
...
+		{
+			// Replay parent activity
+			if (activity.Parent != null)
+			{
+				ReplayLogs(activity.Parent);
+			}
+
+			foreach (LogMessageInformation log in activity.GetReplayableLogs())
+			{
+				m_sender.LogMessage(activity, log.Category, LogLevel.Information, log.EventId, log.ThreadId, log.Message, log.Exception);
```
dakanev-msft, line 43: >Message [](start = 101, length = 7)

The replayed logs from the legacy system used to be serialized into a JSON with replay tag, it was easy to differentiate them. Do we have something similar here or we just replay the raw logs? Not being able to differentiate them can cause confusion if we hit the max replay limit.

src/Extensions/Activities/Internal/ActivityListenerInitializerService.cs
AndreyTretyak: Internal classes don't contribute to documentations, so it makes sense to add comments only when you need to clarify things for somebody who already see the code. In this case it would be `<inheridoc/>` anyways since the method coming from interface

src/Extensions/Activities/Internal/EventSource/ActivityEventSource.cs
AndreyTretyak: It's used in our log aggregator and changing it would require redeploying configs for everybody (plus maintaining both during update). So I  think we could leave it like this until switching to 1DS when it could be simply deleted.

src/Extensions/Compatibility/TimedScopes/TimedScope.cs
AndreyTretyak: Its compatibility package, which is intended to be used as an intermediate step for switching,
So I thought that keeping the have would help people understand what should be used as timescopes replacement.
I'm also considering just deleting these packages since it's not proving very useful.
I'm open to suggestions about naming

src/Extensions/Logging/Internal/EventSource/OmexLogEventSource.cs
AndreyTretyak: Yes, it's already in our configs, so I've removed TODO

src/Extensions/Compatibility/TimedScopes/TimedScope.cs
dakanev-msft: I am more on the side of not having a compatibility package not to increase the complexity, but it will be interesting to see other opinions as well.

---
In reply to: [561022206](https://github.com/microsoft/Omex/pull/285#discussion_r561022206) [](ancestors = 561022206)

dakanev-msft (changes requested):
:clock1:

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
AndreyTretyak: I think it's useful to keep tags and categories to simplify the search, but it's a good very point that they should be distinguishable so I've added a prefix to the message text.

src/Extensions/Logging/Internal/Replayable/ActivityExtensions.cs
AndreyTretyak: Good point interlocker would be useful here, thank you

src/Extensions/Logging/Internal/Replayable/ActivityExtensions.cs
AndreyTretyak: fixed, thanks

src/Extensions/Activities/Internal/ActivityListenerInitializerService.cs
AndreyTretyak: If I understand correctly in the case of canceled token canceled method won't be called, and since all operations quite fast here I think checking it won't help here it's mostly provided for cases if you are calling some async methods here.

src/Extensions/Activities/Internal/DiagnosticsListener/DiagnosticsObserversInitializer.cs
AndreyTretyak: thanks, added issue

src/Extensions/Logging/Internal/Replayable/ReplayableActivityExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		public static IEnumerable<LogMessageInformation> GetReplayableLogs(this Activity activity) =>
+			activity.GetReplayableLogsInternal() ?? Enumerable.Empty<LogMessageInformation>();
+
+		public static void AddReplayLog(this Activity activity, LogMessageInformation logEvent, uint maxReplayedEventsPerActivity)
+		{
+			ConcurrentQueue<LogMessageInformation>? queue = activity.GetReplayableLogsInternal();
+			if (queue == null)
+			{
+				Interlocked.CompareExchange(ref queue, new ConcurrentQueue<LogMessageInformation>(), null);
```
dakanev-msft, line 28: >queue [](start = 36, length = 5)

Just thinking if we should not read the value of queue again here. If the CAS fails because another thread changed the value, will queue contain the new value or we have to read it again with queue = activity.GetReplayableLogsInternal();?


The pattern is usually to do a while loop until the condition is satisfied, i.e. do an interlocked read, then do a swap and if it fails , do an interlocked read again. The interlocked operations are flushing the memory as well. 

Is that a method called so often? A simple lock might be much easier to implement, but would be a bottleneck if that's called more often.

src/Extensions/Logging/Internal/Replayable/ReplayableActivityExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		private static ConcurrentQueue<LogMessageInformation>? GetReplayableLogsInternal(this Activity activity) =>
+			activity.GetCustomProperty(ReplayableLogKey) as ConcurrentQueue<LogMessageInformation>;
+
+		public static IEnumerable<LogMessageInformation> GetReplayableLogs(this Activity activity) =>
+			activity.GetReplayableLogsInternal() ?? Enumerable.Empty<LogMessageInformation>();
+
+		public static void AddReplayLog(this Activity activity, LogMessageInformation logEvent, uint maxReplayedEventsPerActivity)
+		{
+			ConcurrentQueue<LogMessageInformation>? queue = activity.GetReplayableLogsInternal();
```
dakanev-msft, line 25: >GetReplayableLogsInternal [](start = 60, length = 25)

Will it make sense to create a method in the activity itself - GetOrAddReplaableLogsInternal() which does the swap (or Lazy) internally, to separate the concerns in the object that is responsible for having the information?

AndreyTretyak: Activity class provided with .NET we could not change it

src/Extensions/Logging/Internal/Replayable/ActivityExtensions.cs
AndreyTretyak: After some more thinking I've realized that queue is a local variable so `Interlocker` won't help here.
More discussion of alternatives is here:  https://github.com/microsoft/Omex/pull/285#discussion_r561991954

src/Extensions/Logging/Internal/Replayable/ReplayableActivityExtensions.cs
AndreyTretyak: To be honest I don't see a good solution here.
I've considered something like this:
```
ConcurrentQueue<LogMessageInformation>? queue;
while ((queue = activity.GetReplayableLogsInternal()) == null)
{
	activity.SetCustomProperty(ReplayableLogKey, new ConcurrentQueue<LogMessageInformation>());
}
```
But it would end up with the same result, we could still have one query be replaced by another right after creation.

I also don't see a good way of adding lock here, the only object that could be used as a lock is the `activity` itself but that's dangerous since it's public and could be used in some other lock. 

I also think that it's not the most critical case here since in the worst case we are losing a couple of messages to replay, it's not great but we can allow it considering alternatives.

Besides that, I really hope that I'll eventually manage to convince everybody that log replay is not suitable for microservices and we should use variable report level instead, so we would be able to delete reply code completely.

dakanev-msft (approved):
:shipit:

dakanev-msft (approved):
:shipit:

src/Extensions/Compatibility/TimedScopes/TimedScope.cs
K-Cully: Kind of agree with Daniel here on dropping TimedScopes.
If the shim-layer enabled existing dependent applications running in production to soft-migrate without breaking any metrics, dashboards or alerting I'd think it was a good idea but that is not the case here.

Having a shim-layer that doesn't fix the issues caused by this change may cause developers to assume the change does not have the drastic cutover effects it does.

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
```diff
@@ -0,0 +1,48 @@
...
+			// Replay parent activity
+			if (activity.Parent != null)
+			{
+				ReplayLogs(activity.Parent);
+			}
+
+			foreach (LogMessageInformation log in activity.GetReplayableLogs())
+			{
+				string message = "[Replay] " + log.Message; // add prefix to mark log as replay
+				m_sender.LogMessage(activity, log.Category, LogLevel.Information, log.EventId, log.ThreadId, message, log.Exception);
```
K-Cully, line 44: What would the log output be like here?
Will it do the same as the old replay logic - as in wrap the entire log metadata into the message field as a JSON representation - or is it extracting the metadata and converting all the logs to Info level?

src/Extensions/Logging/Internal/EventSource/OmexLogEventSender.cs
```diff
@@ -117,34 +114,9 @@ level switch
 				_ => m_eventSource.IsEnabled()
 			};
 
-		public bool IsReplayableMessage(LogLevel logLevel) =>
```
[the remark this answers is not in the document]
K-Cully: It was just refactored...

dakanev-msft (approved):
:shipit:

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
AndreyTretyak: It would be different than what we had before. It will repeat the same log (tag, category, etc.) but adds the `[Replay]` prefix to the message to distinguish them.

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
```diff
@@ -0,0 +1,48 @@
...
+
+using System.Diagnostics;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Omex.Extensions.Logging.Internal.Replayable;
+using Microsoft.Omex.Extensions.Logging.Replayable;
+
+namespace Microsoft.Omex.Extensions.Logging
+{
+	internal class OmexLogEventReplayer : ILogEventReplayer
```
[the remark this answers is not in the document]
AndreyTretyak: Added comment, thank you

src/Extensions/Compatibility/TimedScopes/TimedScope.cs
AndreyTretyak: Less code - less problems, deleted Compatability packages completely :)

K-Cully (approved):

src/Extensions/Logging/Internal/Replayable/OmexLogEventReplayer.cs
K-Cully: Oh, then there is a bug here.
If the level was higher than Info, it should get it's original level, rather than being downgraded to Info.

Feel free to fix it as a follow on item though.

AndreyTretyak: Replay stores only log messages lower then informational since others are already reported, so by filtering by activity id you would see all logs associated with it (replayed and regular)

dakanev-msft (approved):

PR #296, opened 2021-01-18
Suppress semmel warning

AndreyTretyak:


tests/Extensions/Hosting.Services.Web.UnitTests/WebEndpointInfoTests.cs
```diff
@@ -22,7 +22,7 @@ public void CreateHttp_SetsProperties()
 			Assert.IsNull(info.SettingForCertificateCommonName);
```
muiriswoulfe, line 22: I'm OK with this change – Semmle is obviously reporting a non-issue here – but in reality would we or library consumers ever use the `CreateHttp()` method? HTTP feels very outdated to me know, so perhaps the real fix here is to remove the offending method rather than suppressing the issue in the test.

muiriswoulfe (approved):
Looks good to me, but I've left a comment to consider. Feel free to go ahead with this change anyway as I have no problem with this suppression 🙂

tests/Extensions/Hosting.Services.Web.UnitTests/WebEndpointInfoTests.cs
AndreyTretyak: It's still needed, but we could make it obsolete

AndreyTretyak: Done here https://github.com/microsoft/Omex/pull/297

artempushkin (approved):

PR #301, opened 2021-01-20
Close host on SF service stop

AndreyTretyak:
Stoping generic host when SF service is closing should speed up service shutdown and make services update more responsive.


dakanev-msft (approved):
:shipit:

aszczepanski (approved):

artempushkin (changes requested):
This seems to go against the design intention of Service Fabric. Even if we don't support that scenario, one executable may generally register several service types and/or run multiple service instances. In that case you would be forcefully killing all of them.

Do you have a use case where this change solves a problem?

AndreyTretyak:
> 

Good point I'll close the PR

PR #305, opened 2021-02-12
Set more properties in default ExectuionContext

AndreyTretyak:
It should provide more logging information for Initialization logger or cases when Logger or ActivitySource used separately

This change goes against the common structure of packages but simplifies usage since most of the customers using Servcie Fabric


NuGet.Config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
AndreyTretyak, line 1: Not sure what's the difference here

dyllew3 (approved):

src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs
```diff
@@ -52,6 +55,22 @@ public BaseExecutionContext(IHostEnvironment? hostEnvironment = null)
...
+			MachineId = FormattableString.Invariant($"{MachineName}_{nodeName}");
+
+			string? nodeIPAddressOrFQDN = GetVariable(NodeIPOrFQDNVariableName);
+
+			if (IPAddress.TryParse(nodeIPAddressOrFQDN, out IPAddress? ipAddress))
+			{
+				ClusterIpAddress = ipAddress;
+			}
+
+			Cluster = GetVariable(ClusterNameVariableName) ?? nodeIPAddressOrFQDN ?? MachineId;
```
artempushkin, line 73: So if cluster name is not specified, it will be different for each replica/instance?

artempushkin (changes requested):

NuGet.Config
artempushkin: Maybe line end

src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs
AndreyTretyak: That's to copy existing behavior, what alternative would you suggest?

src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs
```diff
@@ -52,6 +55,22 @@ public BaseExecutionContext(IHostEnvironment? hostEnvironment = null)
...
+			ApplicationName = GetVariable(ApplicationNameVariableName) ?? DefaultEmptyValue;
+
+			string nodeName = GetVariable(NodeNameVariableName) ?? DefaultEmptyValue;
+			MachineId = FormattableString.Invariant($"{MachineName}_{nodeName}");
+
+			string? nodeIPAddressOrFQDN = GetVariable(NodeIPOrFQDNVariableName);
+
+			if (IPAddress.TryParse(nodeIPAddressOrFQDN, out IPAddress? ipAddress))
+			{
+				ClusterIpAddress = ipAddress;
```
artempushkin, line 70: How is machine IP becoming cluster IP? Cluster can have multiple machines.

artempushkin (changes requested):

src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs
AndreyTretyak: Also a copy of existing behavior in our current infrastructure, I'm not attached to it if you have suggestions about new approach

src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs
artempushkin: Then existing behavior is wrong. Leave it empty if not specified.

artempushkin (approved):
I'm approving on the basis that you're not making the code worse.

AndreyTretyak:
> I'm approving on the basis that you're not making the code worse.

Thanks, we would need to discuss what would be better default values

PR #309, opened 2021-02-22
Extend user identity provider interface

aszczepanski:
Extend user identity provider interface to include user identity type and user id as string

### DIFF
--- a/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IUserIdentityProvider.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IUserIdentityProvider.cs
@@ -12,6 +12,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	public interface IUserIdentityProvider
 	{
 		/// <summary>
+		/// Extracts user identity from the http context
+		/// </summary>
+		UserIdentity GetUserIdentity(HttpContext httpContext);
+
+		/// <summary>
 		/// Max bytes that user identity could take
 		/// </summary>
 		public int MaxBytesInIdentity { get; }
--- a/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IpBasedUserIdentityProvider.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IpBasedUserIdentityProvider.cs
@@ -10,16 +10,32 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 {
 	internal class IpBasedUserIdentityProvider : IUserIdentityProvider
 	{
+		public UserIdentity GetUserIdentity(HttpContext httpContext)
+		{
+			IPAddress? remoteIpAddress = GetIpAddress(httpContext);
+
+			return new UserIdentity
+			{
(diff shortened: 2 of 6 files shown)


src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IUserIdentityProvider.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 	/// </summary>
 	public interface IUserIdentityProvider
 	{
+		/// <summary>
+		/// Extracts user identity from the http context
+		/// </summary>
+		UserIdentity GetUserIdentity(HttpContext httpContext);
```
AndreyTretyak, line 17: Could you please explain a bit why this method needed. 
Initial idea was that `TryWriteBytes` provides the same information as a string in `UserIdentity.User`, just with less allocations, and that middleware should not care where are we getting user identity so I'm not sure why would we use `UserIdentity.UserHashType`?

aszczepanski: It's currently added as a prefix of user hash, e.g. userIdType_hashedUser. If there's an implementation which doesn't need it, it can never call this method.

AndreyTretyak: Also having this enum means that you won't be able to register in DI other implementation of `IUserIdentityProvider` that use different identity type

AndreyTretyak: Do you think having this prefix would be useful in any service?

AndreyTretyak (changes requested):
I think middleware should not know about user identity source, and having enum with its types limits the extensibility of the solution, also I don't this that we should have a method in the interface doing the same as existing (`TryWriteBytes`) but less efficient (allocating `sting` instead of using `Span<byte>`).

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IUserIdentityProvider.cs
K-Cully: I don't believe it is directly used in any services.
As far as I'm aware, it was initially used for testing the user-hash generation as it was an easy mechanism to do so.
It may also be used in some analytics streams - though this could (and should) really be addressed by other means.

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/IdentityProviders/IpBasedUserIdentityProvider.cs
```diff
@@ -10,16 +10,32 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 {
 	internal class IpBasedUserIdentityProvider : IUserIdentityProvider
```
K-Cully, line 11: Is the intent that we provide 2 providers for full or limited IP address (last two places zeroed out) or should this provider have that as configuration?
On a side note, did we ever use the full IP for generating the hash in the old code base or did we always use the IP with the last two positions zeroed?

PR #311, opened 2021-02-24
Log Activity events

AndreyTretyak:
After this change, we would log start and stop of each activity with all of its informmation (like ParentId, Baggage and Tags)


AndreyTretyak:
I've also considered using ThreadLocal<T> instead of `StringBuilderPool`, but it looks like id does not provide a lot of performance win, but makes code more complicated (It also possible that benchmarks do not provide enough multithreading load to demonstrate the difference):

|         Method |        Job | IterationCount | LaunchCount | WarmupCount |     Mean |     Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |----------- |--------------- |------------ |------------ |---------:|----------:|---------:|------:|--------:|-------:|------:|------:|----------:|
| RegularBuilder | DefaultJob |        Default |     Default |     Default | 554.8 ns |  11.06 ns | 11.84 ns |  1.00 |    0.00 | 0.1516 |     - |     - |    1272 B |
|    ThreadLocal | DefaultJob |        Default |     Default |     Default | 357.2 ns |   6.94 ns |  6.49 ns |  0.64 |    0.02 | 0.0534 |     - |     - |     448 B |
|           Pool | DefaultJob |        Default |     Default |     Default | 392.3 ns |   7.81 ns |  7.67 ns |  0.71 |    0.02 | 0.0534 |     - |     - |     448 B |
|                |            |                |             |             |          |           |          |       |         |        |       |       |           |
| RegularBuilder |   ShortRun |              3 |           1 |           3 | 576.5 ns | 219.31 ns | 12.02 ns |  1.00 |    0.00 | 0.1516 |     - |     - |    1272 B |
|    ThreadLocal |   ShortRun |              3 |           1 |           3 | 391.2 ns | 234.18 ns | 12.84 ns |  0.68 |    0.04 | 0.0534 |     - |     - |     448 B |
|           Pool |   ShortRun |              3 |           1 |           3 | 377.2 ns | 170.46 ns |  9.34 ns |  0.65 |    0.00 | 0.0534 |     - |     - |     448 B |


<details>
  <summary>Code of the benchmarks</summary>
  
  ```CSharp
// Copyright (C) Microsoft Corporation. All rights reserved.

using System;
using System.Collections.Generic;
using System.Diagnostics;
using System.Text;
using System.Threading;
using BenchmarkDotNet.Attributes;
using Microsoft.Extensions.ObjectPool;

namespace Infrastructure.Benchmarks
{
	[SimpleJob]
	[MemoryDiagnoser]
	[MarkdownExporterAttribute.GitHub]
	public class ThreadLocalBenchmark
	{
		private Activity? activity;

		private ObjectPool<StringBuilder>? m_stringBuilderPool;

		private ThreadLocal<StringBuilder>? m_threadLocal;

		[GlobalSetup]
		public void GlobalSetup()
		{
			activity = new Activity("A").AddBaggage("B", "V1").AddTag("C", "V2").Start();
			activity.Stop();

			m_stringBuilderPool = new DefaultObjectPoolProvider().CreateStringBuilderPool();

			m_threadLocal = new ThreadLocal<StringBuilder>(() => new StringBuilder());
		}

		[Benchmark(Baseline = true)]
		public string RegularBuilder()
		{
			bool isSuccessful = true;
			StringBuilder builder = new StringBuilder();

			string message = builder
				.Append("Ending Activity ").Append(activity!.OperationName)
				.Append(" with ").Append(isSuccessful ? "success" : "failure")
				.Append(' ')
				.AppendObjStart()
				.AppendParamName("Id").Append(activity.Id).AppendSeparator()
				.AppendParamName("Duration").Append(activity.Duration).AppendSeparator()
				.AppendParamName("Baggage").AppendPairs(activity.Baggage).AppendSeparator()
				.AppendParamName("Tags").AppendPairs(activity.TagObjects).AppendSeparator()
				.AppendObjEnd()
				.ToString();

			builder.Clear();

			return message;
		}

		[Benchmark]
		public string ThreadLocal()
		{
			bool isSuccessful = true;
			StringBuilder builder = m_threadLocal!.Value ?? throw new Exception("AAA");

			string message = builder
				.Append("Ending Activity ").Append(activity!.OperationName)
				.Append(" with ").Append(isSuccessful ? "success" : "failure")
				.Append(' ')
				.AppendObjStart()
				.AppendParamName("Id").Append(activity.Id).AppendSeparator()
				.AppendParamName("Duration").Append(activity.Duration).AppendSeparator()
				.AppendParamName("Baggage").AppendPairs(activity.Baggage).AppendSeparator()
				.AppendParamName("Tags").AppendPairs(activity.TagObjects).AppendSeparator()
				.AppendObjEnd()
				.ToString();

			builder.Clear();

			return message;
		}

		[Benchmark]
		public string Pool()
		{
			bool isSuccessful = true;
			StringBuilder builder = m_stringBuilderPool!.Get();

			string message = builder
				.Append("Ending Activity ").Append(activity!.OperationName)
				.Append(" with ").Append(isSuccessful ? "success" : "failure")
				.Append(' ')
				.AppendObjStart()
				.AppendParamName("Id").Append(activity.Id).AppendSeparator()
				.AppendParamName("Duration").Append(activity.Duration).AppendSeparator()
				.AppendParamName("Baggage").AppendPairs(activity.Baggage).AppendSeparator()
				.AppendParamName("Tags").AppendPairs(activity.TagObjects).AppendSeparator()
				.AppendObjEnd()
				.ToString();

			m_stringBuilderPool.Return(builder);

			return message;
		}
	}

	internal static class ActivityLoggerHelpers
	{
		public static StringBuilder AppendObjStart(this StringBuilder builder) => builder.Append('{');

		public static StringBuilder AppendObjEnd(this StringBuilder builder) => builder.Append('}');

		public static StringBuilder AppendArrayStart(this StringBuilder builder) => builder.Append('[');

		public static StringBuilder AppendArrayEnd(this StringBuilder builder) => builder.Append(']');

		public static StringBuilder AppendSeparator(this StringBuilder builder) => builder.Append(',');

		public static StringBuilder AppendParamName(this StringBuilder builder, string name) =>
			builder.Append(name).Append(':');

		public static StringBuilder AppendPairs<T>(this StringBuilder builder, IEnumerable<KeyValuePair<string, T>> pairs)
		{
			builder.AppendArrayStart();

			bool isFirst = true;
			foreach (KeyValuePair<string, T> pair in pairs)
			{
				if (!isFirst)
				{
					builder.AppendSeparator();
				}

				builder
					.AppendObjStart()
					.AppendParamName(pair.Key)
					.Append(pair.Value)
					.AppendObjEnd();
			}

			builder.AppendArrayEnd();

			return builder;
		}
	}
}
  ```
</details>

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+		public static StringBuilder AppendObjEnd(this StringBuilder builder) => builder.Append('}');
+
+		public static StringBuilder AppendArrayStart(this StringBuilder builder) => builder.Append('[');
+
+		public static StringBuilder AppendArrayEnd(this StringBuilder builder) => builder.Append(']');
+
+		public static StringBuilder AppendSeparator(this StringBuilder builder) => builder.Append(',');
+
+		public static StringBuilder AppendParamName(this StringBuilder builder, string name) =>
+			builder.Append(name).Append(':');
```
K-Cully, line 22: Should there be any sanity check on name here?
I guess we want to keep it as lightweight as possible but worth considering if even a null check is warranted - or is the assumption that anything using Activities will be using the latest compiler flags/version which will perform null guards?

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		public static StringBuilder AppendPairs<T>(this StringBuilder builder, IEnumerable<KeyValuePair<string, T>> pairs)
```
K-Cully, line 24: Suggest documenting how the ToString() of T will be used in this method so that a consumer can understand how to use this correctly for complex types.

K-Cully, line 24: Can any restriction be placed on T here?

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
```diff
@@ -0,0 +1,48 @@
...
+		{
+			builder.AppendArrayStart();
+
+			bool isFirst = true;
+			foreach (KeyValuePair<string, T> pair in pairs)
```
K-Cully, line 29: Do we care if these are unique?

src/Extensions/Activities/Internal/ActivityObserver.cs
```diff
@@ -0,0 +1,72 @@
...
+			m_logger = logger;
+			m_stingBuilderPool = objectPoolProvider.CreateStringBuilderPool();
+		}
+
+		public void OnStart(Activity activity, object? payload = null) =>
+			m_logger.LogInformation(0x696e3739 /* tag_in79 */, $"Starting Activity {activity.OperationName} (Id:{activity.Id}; ParentId:{activity.ParentId})");
+
+		public void OnStop(Activity activity, object? payload = null)
+		{
+			m_eventSender.LogActivityStop(activity);
```
K-Cully, line 31: What does this do and why is it logged before the Ending Activity message?
I'm assuming it is to do with ending the activity, whereas the other logging is just representing it in ULS logs.
If so, it would be a bit of an odd debugging experience to have the log message come after the ending of the activity in time.

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -116,6 +116,8 @@ private string SanitizeString(string value, string name, string activityName)
 			return value;
 		}
 
+		public void OnStop(Activity activity, object? payload = null) => throw new NotImplementedException();
```
K-Cully, line 119: Suggest adding a message to this exception to make the code self-document.
Seems a bit odd to add the interface and then not implement it.

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -13,7 +13,7 @@
 
 namespace Microsoft.Omex.Extensions.Activities
 {
-	internal sealed class ActivityEventSender : IActivitiesEventSender
+	internal sealed class ActivityEventSender : IActivitiesEventSender, IActivityStopObserver
```
K-Cully, line 16: Why use the interface if it is not implemented?

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
AndreyTretyak: Sorry this part committed by mistakes

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
AndreyTretyak: Sorry this part committed by mistakes

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
AndreyTretyak: unfortunately no, it has two values `string?` and `object?`, but it\s not very important here

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
AndreyTretyak: I don't think it's needed for two reasons:
1. we are enforcing nullability so there is no proper way of passing nulls there.
2. `Append` actually fine with `null` it's equivalent of adding an empty string https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRj1UwAJUCA6AFQFMAPYAbjzwDcBDMUgI1IF5SUWgHdSAZWBgAllADmAIQjSANnFpgAFAEoW+ApgD8pDgKEQVKvRQCcmgEQAlWgGcLoe6US9KAQQAO/rRQcJoc2jQA9pIy8jq6QA==

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
AndreyTretyak: Not really, it's possible to get duplicated keys, but we anyway should log all of them

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
AndreyTretyak: The class could not be used outside of the project (its internal). Initially, I wanted to make it a nested class but it does not support extension methods.
I think it terms of one single project it's easier to look into implementation.

src/Extensions/Activities/Internal/ActivityObserver.cs
AndreyTretyak: This probably require a better naming it actually sends an ETW event (or IFx in case of Lennon) that creates metrics.

AndreyTretyak: Renamed

src/Extensions/Activities/Internal/ActivityLoggerHelpers.cs
K-Cully: I'd prefer it was documented (even just a note).
I find the pop-up method prompts in VS really useful for maximizing productivity when there are targeted, useful comments like this - even just as a reminder. However, I don't feel strongly on it.

K-Cully (dismissed):

K-Cully (approved):

eoinobrien (approved):

PR #315, opened 2021-03-03
Create new release

dakanev-msft:
Create new release


AndreyTretyak (dismissed):

AndreyTretyak:
Do you think it would make sense to override the version for extensions to make them start from 1.0.0 (they have separate `Directory.Build.props`)

K-Cully (approved):

dakanev-msft:
I think so.

---
In reply to: [790124208](https://github.com/microsoft/Omex/pull/315#issuecomment-790124208) [](ancestors = 790124208)

AndreyTretyak:
> I think so.
> 
> In reply to: [790124208](https://github.com/microsoft/Omex/pull/315#issuecomment-790124208) [](ancestors = 790124208)

Sorry after thinking about it more, I think it does not worth an effort.

**Possible options:**
1. Public: package version (**change**), dll version (**keep**). Internal: package version (**keep**), dll version (**keep**).
Dll for public packages with version 1.0.0 would have dll version 20.0.3, that might be confusing

2. Public: package version (**change**), dll version (**change**). Internal: package version (**keep**), dll version (**keep**).
Require changing our build and publishing pipeline

3. Public: package version (**change**), dll version (**change**). Internal: package version (**change**), dll version (**keep**).
Would lead to dangerous update of internal services, since we are changing from bigger version to lower and all packages and dependencies would need to be carefully updated.

4. Public: package version (**change**), dll version (**change**). Internal: package version (**change**), dll version (**change**).
Would confuse our automatic package update pipeline since version changed to lower, and would still need careful rollout.
 
Proper way would be version #2 with changing our build pipeline to use totally different version for public and internal packages, but I don't think we have capacity for that now. 

Lets keep it as it is to leave versions consistent with other Omex packages that we already have.

K-Cully (approved):

RobJLeonard (approved):

PR #323, opened 2021-03-24
Dyllew/fhl/settings gen v2

dyllew3:


src/CodeGenerators/SettingsGen/BaseGenerator.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.Omex.CodeGenerators.SettingsGen.Parser;
+using Microsoft.Omex.CodeGenerators.SettingsGen.Validation;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen
+{
+	/// <summary>
+	/// The base generator which can be inherited in order to implement your own version
+	/// </summary>
+	/// <typeparam name="TSettingModel">Settings model type</typeparam>
+	public abstract class BaseGenerator<TSettingModel> : ISourceGenerator
```
AndreyTretyak, line 17: Since we are not generating source code should not it be an analyzer instead of a source generator?
I understand it's essentially the same, but it could help better express intention.

src/CodeGenerators/SettingsGen.Example/Testing.cs
```diff
@@ -0,0 +1,42 @@
...
+		/// 
+		/// </summary>
+		public string Whatever { get; set; } = string.Empty;
+
+
+		/// <summary>
+		/// 
+		/// </summary>
+		[Parameter(Name = "TestingThis", Value = "Hmmmm")]
+		public string? DiffName { get; set; }
```
vlad-ion, line 28: Do these need to be nullable? The settings in the resulting xml have empty string values anyway.

src/CodeGenerators/SettingsGen/BaseGenerator.cs
```diff
@@ -0,0 +1,130 @@
...
+			{
+				context.ReportDiagnostic(Diagnostic.Create(s_failedGeneration, null, ex.Message));
+				throw;
+			}
+		}
+
+		/// <inheritdoc />
+		public void Initialize(GeneratorInitializationContext context)
+		{
+			if (Parser != null)
```
vlad-ion, line 69: What happens if parser is null? I think it would not matter so you can just register without the null check?

src/CodeGenerators/SettingsGen/BaseGenerator.cs
```diff
@@ -0,0 +1,130 @@
...
+		/// </summary>
+		protected virtual IParser<TSettingModel>? Parser => null;
+
+		/// <summary>
+		/// FileGenerator which takes a given model and generates a file from it
+		/// </summary>
+		protected virtual IFileGenerator<TSettingModel>? Filegenerator => null;
+
+		/// <summary>
+		///	Validator determines whether or not the new and existing settings match
```
vlad-ion, line 86: nit: extra tab in the commet

src/CodeGenerators/SettingsGen/Microsoft.Omex.CodeGenerators.SettingsGen.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>$(NetStandardVersions)</TargetFramework>
+
```
vlad-ion, line 5: nit: extra newline

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,93 @@
...
+// Licensed under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Xml.Serialization;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Models
+{
+	/// <summary>
+	/// Parmaeter element used for service fabric settings
```
vlad-ion, line 11:
```suggestion
	/// Parameter element used for service fabric settings
```

src/CodeGenerators/SettingsGen/Parser/ClassesWithAttributeFinder.cs
```diff
@@ -0,0 +1,197 @@
...
+		/// List of section classes and the corresponding with their section name
+		/// </summary>
+		public IList<(string sectionName, INamedTypeSymbol className)> Classes { get; } = new List<(string, INamedTypeSymbol)>();
+
+		/// <summary>
+		/// Constructor
+		/// </summary>
+		public ClassesWithAttributeFinder(ISet<string> attributesToFind)
+		{
+
```
vlad-ion, line 41: nit: extra newline

src/CodeGenerators/SettingsGen/Parser/ClassesWithAttributeFinder.cs
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		/// <inheritdoc />
+		public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
+		{
+			// any field with at least one attribute is a candidate for property generation
+			if (context.Node is ClassDeclarationSyntax classDeclaration
+				&& classDeclaration.AttributeLists.Count > 0)
+			{
+				// Named type symvbol should not be null
```
vlad-ion, line 62:
```suggestion
				// Named type symbol should not be null
```

src/CodeGenerators/SettingsGen/SettingsGenerator.cs
```diff
@@ -0,0 +1,82 @@
...
+			AnalyzerConfigOptions options = context.AnalyzerConfigOptions.GetOptions(filename);
+			if (options.TryGetValue("build_metadata.additionalfiles.Generate", out string? shouldGenerate))
+			{
+				return bool.TryParse(shouldGenerate, out bool result) && result;
+			}
+			return false;
+		}
+
+		/// <summary>
+		///
```
vlad-ion, line 70: Some missing comments here and below

src/CodeGenerators/SettingsGen.Example/Testing.cs
dyllew3: They don't need to be nullable and the settings.xml file will have the value in the value field

src/CodeGenerators/SettingsGen/BaseGenerator.cs
dyllew3: I think I will leave it as a generator as it makes it clearer that its generating files and also current;y the code is using source generator interfaces and structs, so would require a big rewrite to get it working again, so will leave it as a generator

src/CodeGenerators/SettingsGen/Microsoft.Omex.CodeGenerators.SettingsGen.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="3.9.0" PrivateAssets="all" />
```
vlad-ion, line 11: You'll want to replace the version values with variables from dependencies.props, right?

src/CodeGenerators/SettingsGen/FileGeneration/IFileGeneration.cs
```diff
@@ -0,0 +1,25 @@
...
+		/// Generate the file with settings in it
+		/// </summary>
+		/// <param name="settings">Settings to write to a file</param>
+		/// <param name="filename">Name of the file to write to</param>
+		void GenerateFile(TSettings settings, string filename);
+
+		/// <summary>
+		/// File type that is being written
+		/// </summary>
+		string FileType { get; }
```
AndreyTretyak, line 23: Why this property needed?

src/CodeGenerators/SettingsGen/Models/Attributes/IgnoreAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Models.Attributes
+{
+	/// <summary>
+	/// Ignore attribute which indicates the property should not be generated to settings file or used in validation
+	/// </summary>
+	[AttributeUsage(AttributeTargets.Property)]
+	public sealed class IgnoreAttribute : Attribute
```
AndreyTretyak, line 12: Could it be that we would face with other attributes that have the same name?

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+		/// <summary>
+		/// IsEncrypted field of the Parameter element
+		/// </summary>
+		public string? IsEncrypted { get; set; }
+
+		/// <summary>
+		/// Type field of the Parameter element
+		/// </summary>
+		public string? Type { get; set; }
```
AndreyTretyak, line 37: Why this field is needed?

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,93 @@
...
+using System;
+using System.Collections.Generic;
+using System.Xml.Serialization;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Models
+{
+	/// <summary>
+	/// Parameter element used for service fabric settings
+	/// </summary>
+	public sealed class ParameterModel : IEqualityComparer<ParameterModel>, IEquatable<ParameterModel>
```
AndreyTretyak, line 13: Could it be `record`?

src/CodeGenerators/SettingsGen.Example/Microsoft.Omex.SettingsGen.Example.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\SettingsGen\Microsoft.Omex.CodeGenerators.SettingsGen.csproj"  OutputItemType="Analyzer"  ReferenceOutputAssembly="false"  />
```
AndreyTretyak, line 8: Could you try setting `ReferenceOutputAssembly` to `true`, and maybe then you don't need other refentece on the next line

src/CodeGenerators/SettingsGen/FileGeneration/XmlFileGeneration.cs
```diff
@@ -0,0 +1,53 @@
...
+		public string FileType => "XML";
+
+		/// <inheritdoc />
+		public void GenerateFile(TSettings settings, string filename)
+		{
+			string content = XmlFileContents(settings);
+
+			// Serializing the xml file creates a different format of the second line in the file as Service Fabric.
+			// SF changes the header when the application project gets build and creates changes to the file if these are not swapped.
+			content = content.Replace("xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\"", "xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"");
```
AndreyTretyak, line 27: Could the same be achieved by `XmlTextWritter.WriteAttributeString` or something similar to avoid dumping everything to string?

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,93 @@
...
+		/// MustOverride attribute of the Parameter element
+		/// </summary>
+		[XmlAttribute]
+		public string? MustOverride { get; set; }
+
+		/// <summary>
+		/// IsEncrypted attribute of the Parmater element
+		/// </summary>
+		[XmlAttribute]
+		public string? IsEncrypted { get; set; }
```
AndreyTretyak, line 44: What value could we have here, judging from the name I would assume it should be `bool`

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,93 @@
...
+		/// Value attribute of the Parameter element
+		/// </summary>
+		[XmlAttribute]
+		public string Value { get; set; } = string.Empty;
+
+		/// <summary>
+		/// MustOverride attribute of the Parameter element
+		/// </summary>
+		[XmlAttribute]
+		public string? MustOverride { get; set; }
```
AndreyTretyak, line 38: Also should not it be `bool`?

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+		/// <summary>
+		/// Value of the parameter
+		/// </summary>
+		public string Value { get; set; } = string.Empty;
+
+		/// <summary>
+		/// MustOverride field of the Parameter element
+		/// </summary>
+		public string? MustOverride { get; set; }
```
AndreyTretyak, line 27: Sorry, for a lot of duplicated comments, but should not `MustOverride` and `IsEncrypted` be `bool` or `bool?`?

src/CodeGenerators/SettingsGen/Models/SectionModel.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Xml.Serialization;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Models
+{
+	/// <summary>
+	/// Xml Model representing the section in Settings.xml
+	/// </summary>
+	public sealed class SectionModel : IEqualityComparer<SectionModel>, IEquatable<SectionModel>
```
AndreyTretyak, line 14: Again it looks like record would make sense here

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
dyllew3: For the attribute properties they can't be boolean

dyllew3: can't be nullable boolean

src/CodeGenerators/SettingsGen/Wrappers/AttributeWrapper.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Wrappers
+{
+	internal sealed class AttributeWrapper : IAttributeWrapper
```
AndreyTretyak, line 8: Why do you need `IAttributeWrapper` interface? And `IContextWrapper`?

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
dyllew3: Can't be nullable boolean but will convert to bool

src/CodeGenerators/SettingsGen/SettingsGenerator.cs
```diff
@@ -0,0 +1,80 @@
...
+
+						return ShouldGen(context, file);
+					}
+				}
+				settingsFile = null;
+				return false;
+			}
+			catch (Exception ex)
+			{
+				Console.WriteLine($"Encountered exception {ex.Message} when getting settings xml");
```
AndreyTretyak, line 52: Would this error be actually displayed? Should we make it an analyzer error to display in the build result?

src/CodeGenerators/SettingsGen/Wrappers/AttributeWrapper.cs
dyllew3: So the context and IAttribute wrapper were created in order to limit the usage of getting attributes and named symbols directly from GeneratorExecutionContext as its a readonly struct and there aren't many nice ways of testing it

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
dyllew3: This is a valid field in service fabric settings xml

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
dyllew3: And if it has a default value it will always output it

src/CodeGenerators/SettingsGen/Models/Attributes/IgnoreAttribute.cs
dyllew3: So there is a risk of that but this attribute will only be looked for when it is on the property of class which also has the Attribute section on it so the risk of overlap should be 0

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
AndreyTretyak: But why it's needed on an attribute? Could not we get the type from property type?

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
AndreyTretyak: Then what about using bool and skip adding attribute if it `false`. Or use different type for serialization and that would also help with Type property

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.Collections.Generic;
+using System.Xml.Serialization;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.Models
+{
+	/// <summary>
+	/// Parameter element used for service fabric settings
+	/// </summary>
+	public sealed record ParameterModel : IEqualityComparer<ParameterModel>, IEquatable<ParameterModel>
```
AndreyTretyak, line 13: The point for using records is that you don't need to make equality manully and properties

AndreyTretyak: For example https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAIBOBTAxge02HQAUBDTMgW2wBdtMBZfYbAGwAoo4AGdAOWrYANOm58AamTYBXEegBG+fG3SMZAZ1oB5AG4NMAS1YBKANxA=

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
AndreyTretyak: Also, this type not an attribute so you can have `bool?` property

src/CodeGenerators/SettingsGen/Wrappers/AttributeWrapper.cs
AndreyTretyak: Still trying to understand it's usage, but if it's needed only for tests it should be internal

tests/CodeGenerators/SettingsGen.UnitTests/FileGeneration/XmlFileGenerationUnitTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Omex.CodeGenerators.SettingsGen.Models;
+using Microsoft.Omex.System.UnitTests.Shared;
+using Xunit;
+
+namespace Microsoft.Omex.CodeGenerators.SettingsGen.UnitTests.FileGeneration
+{
+	public sealed class XmlFileGenerationUnitTests : UnitTestBase
+	{
+
+		[Fact]
```
AndreyTretyak, line 15: Not sure how it's relevant, but some time ago we were pushing change from XUnit to MSTest

src/CodeGenerators/SettingsGen/Parser/SettingsFromAttributeParser.cs
```diff
@@ -0,0 +1,174 @@
...
+			if (attributeData.Arguments.TryGetValue(nameof(ParameterModel.Name), out string value))
+			{
+				parameterModel.Name = value;
+			}
+
+			// Set must override property
+			if (attributeData.Arguments.TryGetValue(nameof(ParameterModel.MustOverride), out value))
+			{
+				parameterModel.MustOverride = value;
+				}
```
AndreyTretyak, line 153: nit: wrong whitespaces

src/CodeGenerators/SettingsGen.Example/Settings.xml
```diff
@@ -0,0 +1,13 @@
...
+    <Parameter Name="Whatever" Value="" />
+    <Parameter Name="TestingThis" Value="Hmmmm" />
+    <Parameter Name="Overriding" Value="" MustOverride="true" IsEncrypted="true" />
+  </Section>
+  <Section Name="SectionTwo">
+    <Parameter Name="TestingSectionTwo" Value="" />
+    <Parameter Name="TestingSectionThree" Value="" />
+    <Parameter Name="SomethingElse" Value="" MustOverride="true" />
+  </Section>
+</Settings>
```
AndreyTretyak, line 13: nit: missing end line

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
dyllew3: Type field setting and property type are not the same thing, it can only be a number of predefined values, but didn't think it was worth creating an enum for it given its low usage

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
dyllew3: Xml fails to reflect for bool? also gonna keep it as string? because I'm having issues getting it to ignore default value

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
dyllew3: Had issues with equals in section settingsxml models

tests/CodeGenerators/SettingsGen.UnitTests/FileGeneration/XmlFileGenerationUnitTests.cs
dyllew3: Think I will keep it as xunit for now

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
AndreyTretyak: And what's is the use case for this field? when would we need to override it?

src/CodeGenerators/SettingsGen/Models/SettingsXmlModel.cs
```diff
@@ -0,0 +1,44 @@
...
+		/// </summary>
+		[XmlElement("Section")]
+		public List<SectionModel> Sections { get; set; } = new();
+
+		/// <inheritdoc/>
+		public bool Equals(SettingsXmlModel other) => Equals(this, other);
+		
+		/// <inheritdoc/>
+		public bool Equals(SettingsXmlModel x, SettingsXmlModel y) =>
+			x.Sections.Count == y.Sections.Count && x.Sections.All(param => y.Sections.Any(paramTwo => paramTwo.Equals(param)));
```
AndreyTretyak, line 35: This might be better https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.sequenceequal?view=net-5.0

src/CodeGenerators/SettingsGen/Parser/SettingsFromAttributeParser.cs
```diff
@@ -0,0 +1,174 @@
...
+		/// <inheritdoc />
+		public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
+		{
+			(IList<IAttributeWrapper> attributes, INamedTypeSymbol? classSymbol) = m_contextWrapper.GetAttributes(context);
+
+			if (attributes.Count > 0 && classSymbol != null)
+			{
+				foreach (IAttributeWrapper attribute in attributes)
+				{
+					if (ClassAttributes.Any(x => string.Equals(attribute.Name, x, StringComparison.OrdinalIgnoreCase)))
```
AndreyTretyak, line 65: `Contains` might be more efficient here

src/CodeGenerators/SettingsGen/Parser/SettingsFromAttributeParser.cs
```diff
@@ -0,0 +1,174 @@
...
+			parameterModel = new ParameterModel();
+
+			// If it has the ignore attribute don't generate it
+			if (attributes.Any(x => string.Equals(x.Name, AttributeNames.Ignore, StringComparison.OrdinalIgnoreCase)))
+			{
+				return false;
+			}
+
+			// Handle parameter attribute present
+			IEnumerable<IAttributeWrapper> parameters = attributes.Where(
```
AndreyTretyak, line 123: Using `FirstOrDefault` and then checking for `null` could be better than `Where`/`ToList`/`Count`/`First`

src/CodeGenerators/Directory.Build.props
```diff
@@ -0,0 +1,14 @@
...
+  </PropertyGroup>
+  <ItemGroup Label="Make internals visible for the corresponding unit test assemblies and for Moq">
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>$(AssemblyName).UnitTests$(StrongNameSuffix)</_Parameter1>
+    </AssemblyAttribute>
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
+      <_Parameter1>DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
+</Project>
```
AndreyTretyak, line 14: nit: missing new line

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
dyllew3: Basically you would state if its plaintext or encrypted, we haven't had to use it but keeping it in there to allow others to use it

tests/CodeGenerators/SettingsGen.UnitTests/Microsoft.Omex.CodeGenerators.SettingsGen.UnitTests.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
```
AndreyTretyak, line 4: Is there are reason to not enable nullability here?

src/CodeGenerators/SettingsGen/Models/Attributes/ParameterAttribute.cs
dyllew3: https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-service-model-schema-elements#xml-source-158

src/CodeGenerators/SettingsGen/Wrappers/AttributeWrapper.cs
dyllew3: Have made it internal

tests/CodeGenerators/SettingsGen.UnitTests/Microsoft.Omex.CodeGenerators.SettingsGen.UnitTests.csproj
dyllew3: handled by direcrory.build.props

src/CodeGenerators/SettingsGen/Wrappers/AttributeWrapper.cs
dyllew3: Also the issue is mocking generatorsyntaxcontext, so in this case I am just putting the struct in a wrapper so that I can test getting the attributes without relying on the struct

src/CodeGenerators/SettingsGen/FileGeneration/XmlFileGeneration.cs
dyllew3: This was able to removed

src/CodeGenerators/SettingsGen/SettingsGenerator.cs
dyllew3: Have it output as an error now

AndreyTretyak (dismissed):

src/CodeGenerators/SettingsGen/Parser/SettingsFromAttributeParser.cs
```diff
@@ -0,0 +1,173 @@
...
+		/// List of section classes and the corresponding with their section name
+		/// </summary>
+		public IList<(string sectionName, INamedTypeSymbol className)> Classes { get; } = new List<(string, INamedTypeSymbol)>();
+
+		/// <summary>
+		/// Constructor
+		/// </summary>
+		internal SettingsFromAttributeParser(ISet<string> attributesToFind, IContextWrapper contextWrapper)
+		{
+
```
vlad-ion, line 41: nit: extra newline

src/CodeGenerators/SettingsGen/Parser/SettingsFromAttributeParser.cs
```diff
@@ -0,0 +1,173 @@
...
+			// Handle parameter attribute present
+			IAttributeWrapper? parameterAttr = attributes.FirstOrDefault(
+				x => string.Equals(x.Name, AttributeNames.Parameter, StringComparison.OrdinalIgnoreCase));
+			if (parameterAttr is not null)
+			{
+				LoadParameterFromAttribute(parameterModel, parameterAttr, propertyName);
+				return true;
+			}
+
+			// Handle no parameter tag present just assume property name is the setting name
```
vlad-ion, line 131:
```suggestion
			// Handle no parameter tag present: just assume property name is the setting name
```

tests/CodeGenerators/Directory.Build.props
```diff
@@ -0,0 +1,17 @@
...
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(Microsoft_Net_Test_Sdk_Version)" />
+    <PackageReference Include="MSTest.TestAdapter" Version="$(MSTest_TestAdapter_Version)" PrivateAssets="All" />
+    <PackageReference Include="MSTest.TestFramework" Version="$(MSTest_TestFramework_Version)" />
+    <PackageReference Include="Moq" Version="$(Moq_Version)" />
+    <!-- exclude xunit packages used in other test projest -->
```
vlad-ion, line 12:
```suggestion
    <!-- exclude xunit packages used in other test projects -->
```

src/CodeGenerators/SettingsGen.Example/Testing.cs
vlad-ion: What I meant here is if it needs to be "string?" and can't it just be "string DiffName"

src/CodeGenerators/SettingsGen.Example/Microsoft.Omex.SettingsGen.Example.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <!--
```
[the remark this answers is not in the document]
dyllew3: I have it in there to show another way of using the settings generator

src/CodeGenerators/SettingsGen/BaseGenerator.cs
```diff
@@ -0,0 +1,147 @@
...
+
+		/// <summary>
+		/// Parser which parses syntax into a given model format to be used by file generator
+		/// </summary>
+		protected virtual IParser<TSettingModel>? Parser => null;
+
+		/// <summary>
+		/// FileGenerator which takes a given model and generates a file from it
+		/// </summary>
+		protected virtual IFileGenerator<TSettingModel>? Filegenerator => null;
```
vlad-ion, line 78: nit: this could be renamed to FileGenerator

src/CodeGenerators/Directory.Build.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('$(_DirectoryBuildPropsFile)', '$(MSBuildThisFileDirectory)../'))" />
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+  <ItemGroup Label="Make internals visible for the corresponding unit test assemblies and for Moq">
+    <AssemblyAttribute Include="System.Runtime.CompilerServices.InternalsVisibleTo">
```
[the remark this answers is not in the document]
dyllew3: In order to be able to test them and also to mock them

src/CodeGenerators/SettingsGen/Models/ParameterModel.cs
```diff
@@ -0,0 +1,52 @@
...
+		public string Value { get; set; } = string.Empty;
+
+		/// <summary>
+		/// MustOverride attribute of the Parameter element
+		/// </summary>
+		[XmlAttribute]
+		public string? MustOverride { get; set; }
+
+		/// <summary>
+		/// IsEncrypted attribute of the Parmater element
```
vlad-ion, line 41:
```suggestion
		/// IsEncrypted attribute of the Parameter element
```

src/CodeGenerators/SettingsGen/FileGeneration/SettingsAdditionalText.cs
```diff
@@ -0,0 +1,44 @@
...
+		public SettingsAdditionalText(string path)
+		{
+			if (string.IsNullOrEmpty(path))
+			{
+				throw new ArgumentException("Empty filename given");
+			}
+
+			m_path = path;
+
+			string fileContent = File.ReadAllText(m_path);
```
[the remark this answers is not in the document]
dyllew3: Yes this will be caught by the try catch in the main part of the generator

src/CodeGenerators/SettingsGen/BaseGenerator.cs
```diff
@@ -0,0 +1,147 @@
...
+
+		/// <summary>
+		///	Comparer determines whether or not the new and existing settings match
```
vlad-ion, line 81:
```suggestion
		/// Comparer determines whether or not the new and existing settings match
```

src/CodeGenerators/Directory.Build.props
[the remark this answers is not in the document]
vlad-ion: I was thinking more like we shouldn't test internal behaviour :). But if you need it, sure.

src/CodeGenerators/SettingsGen.Example/Microsoft.Omex.SettingsGen.Example.csproj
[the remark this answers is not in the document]
vlad-ion: Might help to add an explanation about this in the commented area itself, like "this is an alternative way to..."

vlad-ion (approved):

AndreyTretyak (approved):

PR #334, opened 2021-04-21
Rest Health Check Publisher

mvxxx:
# Changes
- Implementation of the REST API Health Check Publisher.  API: [link](https://docs.microsoft.com/en-us/rest/api/servicefabric/sfclient-api-reportservicehealth)
- Restructure of current code to make inheritance
- Unit tests for publishers


src/Extensions/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
```diff
@@ -1,15 +1,17 @@
...
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.ServiceFabric" Version="$(Microsoft_ServiceFabric_Version)" />
-    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(Microsoft_Extensions_Diagnostics_HealthChecks_Version)" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="$(Microsoft_Extensions_Http_Version)" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="$(Microsoft_Extensions_ObjectPool_Version)" />
-  </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\Abstractions\Microsoft.Omex.Extensions.Abstractions.csproj" />
+    <PackageReference Include="Microsoft.Omex.Extensions.Abstractions" Version="21.4.10420.11100" />
```
mvxxx, line 10: I will revert these changes with explicite version numbers

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using sfh = System.Fabric.Health;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	/// <summary>
+	///
```
dyllew3, line 15: nit: empty comments

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	/// <summary>
+	/// Just mock
+	/// </summary>
+	public class RestHealthCheckPublisher : HealthCheckPublisher
+	{
+		private readonly ServiceFabricHttpClient m_client;
+		private readonly string m_serviceId;
+		/// <summary>
```
dyllew3, line 26: nit: newline

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+				foreach (KeyValuePair<string, HealthReportEntry> entryPair in report.Entries)
+				{
+					cancellationToken.ThrowIfCancellationRequested();
+					await PublishHealthInfo(BuildSfHealthInformation(entryPair.Key, entryPair.Value), cancellationToken);
+				}
+
+				cancellationToken.ThrowIfCancellationRequested();
+				await PublishHealthInfo(BuildSfHealthInformation(report), cancellationToken);
+			}
+			catch (Exception)
```
dyllew3, line 106: nit: If not using exception can just have empty catch
i.e.
catch
{
    
}

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -45,5 +62,30 @@ public static IServiceCollection AddHealthCheckPublisher<TPublisher>(this IServi
...
+				{
+					return new RestHealthCheckPublisher(
+						clusterEndpoints: new Uri(serviceProvider
+													.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>()
+													.Value
+													.RestHealthPublisherClusterEndpoint),
+						serviceId: serviceProvider
+													.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>()
+													.Value
+													.RestHealthPublisherServiceId
```
AndreyTretyak, line 84:
```suggestion
					RestHealthCheckPublisherOptions options = serviceProvider.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>().Value;
					return new RestHealthCheckPublisher(
						clusterEndpoints: new Uri(options.RestHealthPublisherClusterEndpoint),
						serviceId: options.RestHealthPublisherServiceId
```

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+				).ConfigureAwait(false);
+		}
+
+		/// <summary>
+		/// Just mock
+		/// </summary>
+		/// <param name="report"></param>
+		/// <param name="cancellationToken"></param>
+		/// <returns></returns>
+		public override async Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
```
dimeehan, line 92: I would suggest moving public method above private ones

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+				   description: healthInfo.Description,
+				   sequenceNumber: healthInfo.SequenceNumber.ToString(),
+				   removeWhenExpired: healthInfo.RemoveWhenExpired,
+				   healthReportId: healthInfo.HealthReportId
+		   );
+
+			return sfcHealthInfo;
+		}
+
+		private async Task PublishHealthInfo(sfh.HealthInformation healthInfo, CancellationToken cancellationToken)
```
dimeehan, line 74: nit: we usually add Async to async method names like PublishHealthInfoAsync

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Omex.Extensions.Abstractions.ExecutionContext;
+using Microsoft.ServiceFabric.Client;
+using Microsoft.ServiceFabric.Client.Http;
+using sfc = Microsoft.ServiceFabric.Common;
+using sfh = System.Fabric.Health;
```
dimeehan, line 15: nit: I think we can improve the naming here so it's more readable in code, maybe something like ServiceFabricHealth instead on sfh?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+				.BuildAsync().GetAwaiter().GetResult();
+		}
+
+
+		/// <summary>
+		/// 
+		/// </summary>
+		/// <param name="healthInfo"></param>
+		/// <returns></returns>
+		private sfc.HealthInformation fromSfHealthInformation(sfh.HealthInformation healthInfo)
```
AndreyTretyak, line 48: The naming convention for .NET requires starting method name from capital letter. https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
AndreyTretyak: Also it might be better just get `IOptions<RestHealthCheckPublisherOptions>` as constructer parameter for `RestHealthCheckPublisher`, then you can just register type.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -15,10 +17,7 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 	/// </summary>
 	public static class ServiceCollectionExtensions
 	{
-		/// <summary>
-		/// Register types required for Omex health check
-		/// </summary>
-		public static IHealthChecksBuilder AddServiceFabricHealthChecks(this IServiceCollection serviceCollection)
```
AndreyTretyak, line 21: I think we would need to keep the method signature as it is, to avoid breaking existing clients and just add an alternative method to register new publisher

src/Extensions/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
AndreyTretyak: `Microsoft.Omex.Extensions.Abstractions` project located in this repository, so it should be `ProjectReference` instead.

AndreyTretyak: `Microsoft.Omex.Extensions.Abstractions` project located in this repository, so it should be `ProjectReference instead. It was in this project before (line 13).

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,141 @@
...
+		/// 
+		/// </summary>
+		protected const string HealthReportSummaryProperty = "HealthReportSummary";
+
+		/// <summary>
+		/// 
+		/// </summary>
+		public HealthCheckPublisher()
+		{
+			m_descriptionBuilder = new();
```
AndreyTretyak, line 44: We should not create new instances of `StringBuilder` to avoid allocations. Looks how it was implemented before with `StringBuilderPool`.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+
+		/// <summary>
+		/// 
+		/// </summary>
+		public RestHealthCheckPublisher(Uri clusterEndpoints, string serviceId) : base()
+		{
+			m_serviceId = serviceId;
+			m_client = (ServiceFabricHttpClient)new ServiceFabricClientBuilder()
+				.UseEndpoints(clusterEndpoints)
+				.BuildAsync().GetAwaiter().GetResult();
```
AndreyTretyak, line 39: Calling `.GetAwaiter().GetResult()` considered a bad practice since it could lead to wrong handling of SyncronizationContext and deadlocks.
I understand that you are unable to make constructor async, so probably the only option here would be to build a client during the first call of `PublishAsync` or consider alternative design approaches that I'll describe in the general comment.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -30,12 +29,30 @@ public static IHealthChecksBuilder AddServiceFabricHealthChecks(this IServiceCol
...
+		{
+			serviceCollection.ConfigureService();
+			Type publisherType = typeof(TPublisher);
+			if (publisherType == typeof(RestHealthCheckPublisher))
+			{
+				return serviceCollection.AddRestHealthCheckPublisher().AddHealthChecks();
+
+			}
+
+			return serviceCollection.AddHealthCheckPublisher<TPublisher>().AddHealthChecks();
```
AndreyTretyak, line 50: This method won't be needed at all if you would change `RestHealthCheckPublisher` provided as suggested here https://github.com/microsoft/Omex/pull/334/files#r617671328
Then both publishers could be registered by calling `AddHealthCheckPublisher`

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.ServiceFabric.Client.Http;
+using sfc = Microsoft.ServiceFabric.Common;
+using sfh = System.Fabric.Health;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	/// <summary>
+	/// Just mock
+	/// </summary>
+	public class RestHealthCheckPublisher : HealthCheckPublisher
```
AndreyTretyak, line 22: I would suggest making class `internal` and just provide public method to register it on ServiceCollection. Is this way you would reduce possibility of the wrong usage of the type and remove the need of comments :)

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using sfh = System.Fabric.Health;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public abstract class HealthCheckPublisher : IHealthCheckPublisher
```
AndreyTretyak, line 17: Again I think this class should be internal

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -45,5 +62,30 @@ public static IServiceCollection AddHealthCheckPublisher<TPublisher>(this IServi
...
+						clusterEndpoints: new Uri(serviceProvider
+													.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>()
+													.Value
+													.RestHealthPublisherClusterEndpoint),
+						serviceId: serviceProvider
+													.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>()
+													.Value
+													.RestHealthPublisherServiceId
+					);
+				}
```
AndreyTretyak, line 86: I think it would be better to get `IOptions<RestHealthCheckPublisherOptions>` as a constructor parameter for `RestHealthCheckPublisher` and then it could be resolved as a regular type without an explicit factory.

src/Extensions/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
```diff
@@ -1,15 +1,17 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>$(LibraryTargetFrameworks)</TargetFrameworks>
+    <TargetFrameworks>net5.0</TargetFrameworks>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net5.0|AnyCPU'">
     <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
```
AndreyTretyak, line 8: I think this was not intended change

AndreyTretyak (commented):
Following description provided just to describe the broader goal and it could be addressed later.

The main reason behind creating rest publisher is to avoid using `ServiceContext`, but another added benefit is that we could avoid depending on native SF dll's and building only for x64. 

To do that rest publisher and sf publisher would need to be in separate projects.
`Microsoft.Omex.Extensions.Diagnostics.HealthChecks.Abstractions` (I'm open to better naming) with rest publisher and common functionality
`Microsoft.Omex.Extensions.Diagnostics.HealthChecks` (we would need to keep this naming to avoid breaking existing clients) depending on the previous project and would contain only part that depends on ServiceContext.

The problem here would be the fact that abstract publisher internal as an alternative we could have one `ServiceFabricPublisher` that would accept interface for sending data to SF (with Rest and ServiceContext implementations in different publisher). In this way  to change from rest to service context publisher we would just need to override interface implementation registered in DI. 

Feel free to ping me and we could discuss this approach in more detail.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
mvxxx: Restored old signature

src/Extensions/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
mvxxx: Yeah, there were my local changes only to work (I am developing on 2 repos simultaneously) and committed it by mistake. Will revert soon

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: Added RestClientWrapper

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -30,20 +31,47 @@ public static IHealthChecksBuilder AddServiceFabricHealthChecks(this IServiceCol
...
+		/// <summary>
+		/// Register publisher for processing health check results directly to nodes using REST api
+		/// </summary>
+		public static IHealthChecksBuilder AddRestHealthChecks(this IServiceCollection serviceCollection)
+		{
+			serviceCollection.TryAddSingleton(
+				provider =>
+				{
+					return provider.GetRequiredService<IOptions<RestHealthCheckPublisherOptions>>().Value;
+				});
```
AndreyTretyak, line 53: `IOptions<>` registered as open generic, there is no need to register each option separately.
```suggestion
```

AndreyTretyak, line 53: `IOptions<>` registered as open generic, there is no need to register each option separately.
```suggestion
```

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -30,20 +31,47 @@ public static IHealthChecksBuilder AddServiceFabricHealthChecks(this IServiceCol
...
+			return serviceCollection.AddServiceFabricHealthChecks<RestHealthCheckPublisher>();
 		}
 
 		/// <summary>
 		/// Register publisher for processing health check results
 		/// </summary>
+		public static IHealthChecksBuilder AddServiceFabricHealthChecks<TPublisher>(this IServiceCollection serviceCollection)
+				where TPublisher : class, IHealthCheckPublisher
+		{
+			serviceCollection.ConfigureService();
```
AndreyTretyak, line 63: Why this line needed? If I'm not mistaken this call does not do anything.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -30,20 +31,47 @@ public static IHealthChecksBuilder AddServiceFabricHealthChecks(this IServiceCol
...
+		/// <summary>
+		/// Register publisher using Dependency Injection
+		/// </summary>
 		public static IServiceCollection AddHealthCheckPublisher<TPublisher>(this IServiceCollection serviceCollection)
 			where TPublisher : class, IHealthCheckPublisher
 		{
 			serviceCollection.TryAddEnumerable(ServiceDescriptor.Singleton<IHealthCheckPublisher, TPublisher>());
 			return serviceCollection;
-		}
+		}
```
AndreyTretyak, line 75: nit: Trailing whitespace
```suggestion
		}
```

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+		public RestClientWrapper(Uri clusterEndpoint)
+#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
+		{
+			// m_client is initialized during first publish async in order to avoid possible deadlocks
+			m_clusterEndpoint = clusterEndpoint;
+			m_ready = false;
+		}
+
+		public ServiceFabricHttpClient Get()
```
AndreyTretyak, line 27: This method should have `async` modifier (https://docs.microsoft.com/en-us/dotnet/csharp/async) then you would be able to use `await` keyword to get the result of the `Task` instead of calling `GetAwaiter` directly.

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+	internal class RestClientWrapper
+	{
+		private ServiceFabricHttpClient m_client;
+
+		private readonly Uri m_clusterEndpoint;
+
+		private bool m_ready;
+
+#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```
AndreyTretyak, line 18: Suppressing warnings should be the last resort, I think in this case it would be better to make field `m_client` nullable (with type `ServiceFabricHttpClient?`) and check it for null in `Get` method. Then you also won't need `m_ready` field.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,104 @@
...
+		internal async Task PublishHealthInfoAsync(ServiceFabricHealth.HealthInformation healthInfo, CancellationToken cancellationToken,
+			ServiceFabricHttpClient httpClient)
+		{
+			ServiceFabricCommon.HealthInformation sfcHealthInfo = FromSfHealthInformation(healthInfo);
+			string? nodeName = Environment.GetEnvironmentVariable(FabricNodeNameEnv);
+			if(nodeName == null)
+			{
+				m_logger.LogError(Tag.Create(), "Can't find node name.");
+				return;
+			}
```
AndreyTretyak, line 96: I think it would be better to get the node name once since it would never change.
An even better approach would be just to expose it as another field in `IExecutionContext` since we already getting it there. https://github.com/microsoft/Omex/blob/4732acfb475274e111504098d01332a15d0aa374/src/Extensions/Abstractions/ExecutionContext/BaseExecutionContext.cs#L63

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	internal abstract class HealthCheckPublisher : IHealthCheckPublisher
+	{
+		protected readonly ObjectPool<StringBuilder> StringBuilderPool;
+
+		protected string HealthReportSourceId { get { return HealthReportSourceIdImpl; } }
+
+		protected abstract string HealthReportSourceIdImpl { get; }
```
AndreyTretyak, line 21: Why do we need two separate properties? I think just making `HealthReportSourceId` abstract would give the same result?

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceFabricHealthCheckPublisher.cs
```diff
@@ -12,33 +12,31 @@
...
 
-		private readonly ILogger<ServiceFabricHealthCheckPublisher> m_logger;
+		private readonly IAccessor<StatelessServiceContext> m_serviceCtx;
 
-		private readonly ObjectPool<StringBuilder> m_stringBuilderPool;
+		private readonly ILogger<ServiceFabricHealthCheckPublisher> m_logger;
 
 		public ServiceFabricHealthCheckPublisher(
 			IAccessor<IServicePartition> partitionAccessor,
+			IAccessor<StatelessServiceContext> serviceCtx,
```
AndreyTretyak, line 28: This dependency should not be here, since then Stateful service won't be able to resolve publisher, and also I don't see how are you using it.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,104 @@
...
+
+		public RestHealthCheckPublisher(ILogger<ServiceFabricHealthCheckPublisher> logger,
+										RestHealthCheckPublisherOptions options,
+										ObjectPoolProvider objectPoolProvider) : base(objectPoolProvider)
+		{
+			m_clientWrapper = new(new Uri(options.RestHealthPublisherClusterEndpoint));
+			m_logger = logger;
+		}
+
+		public override async Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
```
AndreyTretyak, line 40: It looks to me as this method very similar to the implementation in `ServiceFabricHealthCheckPublisher` and could be also moved to the base class, instead you can create `private abstract Task ReportHealthAsync(SfHealthInformation info)` with implementation diffirences.

AndreyTretyak:
Looks like test projects are failing to build, you would need to update them also.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
mvxxx: It calls configuration for HTTP communication:
```cs
private static void ConfigureService(this IServiceCollection serviceCollection)
		{
			serviceCollection
				.AddHttpClient(HttpEndpointHealthCheck.HttpClientLogicalName)
				.ConfigurePrimaryHttpMessageHandler(() => new HttpClientHandler
				{
					AllowAutoRedirect = false,
					Credentials = CredentialCache.DefaultCredentials,
					ServerCertificateCustomValidationCallback = (sender, x509Certificate, chain, errors) => true
				});

			serviceCollection.TryAddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
		}
```

It was used also before: [line](https://github.com/microsoft/Omex/blob/master/src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs#L23)

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
mvxxx: OK, done

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: Ok, I assumed that they can dynamically rotate

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
AndreyTretyak: Or I have not realized that it's our method, we would need to think about better naming here, ex `AddPublisherDepenencies()`.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: it was probably some workaround - I dropped this now.

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
```diff
@@ -0,0 +1,38 @@
...
+			{
+				ConfiguredTaskAwaitable < IServiceFabricClient > result = new ServiceFabricClientBuilder()
+					.UseEndpoints(m_clusterEndpoint)
+					.BuildAsync().ConfigureAwait(false);
+
+				m_client = (ServiceFabricHttpClient?)await result;
+			}
+
+			return m_client!;
+		}
```
AndreyTretyak, line 36: I would suggest something like this instead to avoid dangerous casting (line 32) and dismissing nullable warning (line 35), but I have not validated it with the compiler, please let me know if it won't compile and we could search for a solution.
```suggestion
		private IServiceFabricClient? m_client;

		private readonly Uri m_clusterEndpoint;

		public RestClientWrapper(Uri clusterEndpoint)
		{
			// m_client is initialized during first publish async in order to avoid possible deadlocks
			m_clusterEndpoint = clusterEndpoint;
		}

		public async Task<ServiceFabricHttpClient> GetAsync()
		{
			if (m_client == null)
			{
				m_client = await new ServiceFabricClientBuilder()
					.UseEndpoints(m_clusterEndpoint)
					.BuildAsync()
					.ConfigureAwait(false);
			}

			return m_client;
		}
```

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,102 @@
...
+		{
+			if(m_nodeName == null)
+			{
+				m_logger.LogError(Tag.Create(), "Can't find node name.");
+				return;
+			}
+
+			m_client = await m_clientWrapper.GetAsync();
```
AndreyTretyak, line 50: You are already caching the client in a field once in the client wrapper, not need to save it here also, or alternatively remove the wrapper.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		internal static string? HealthReportSourceId => nameof(HealthCheckPublisher);
+
+		internal const string HealthReportSummaryProperty = "HealthReportSummary";
+
+		public HealthCheckPublisher(ObjectPoolProvider objectPoolProvider) => StringBuilderPool = objectPoolProvider.CreateStringBuilderPool();
+
+		public abstract Task PublishAsync(HealthReport report, CancellationToken cancellationToken);
+
+		protected void PublishAllEntries(HealthReport report, Action<ServiceFabricHealth.HealthInformation> publishFunc,
```
AndreyTretyak, line 28: Having a protected abstract method might be better than `Action<ServiceFabricHealth.HealthInformation>`, but it up to you to decide what fit's better.

mvxxx: Yeah, I considered that - however the problem is that publishFunc can be async (as in RestHealthPublisher) or not (as partition publisher) - so functional approach seems to solve problem with that difference

AndreyTretyak: `async` is not a part of function signature so you can define an abstract function like:
```
protected abstract Task ReportHelthInfoAsync(ServiceFabricHealth.HealthInformation information);
```
And then have two implementations, sync:
```
protected override Task ReportHelthInfoAsync(ServiceFabricHealth.HealthInformation information)
{
     // Sync code
     return Task.CompletedTask;
}
```
async:
```
protected override async Task ReportHelthInfoAsync(ServiceFabricHealth.HealthInformation information)
{
     // code with await
}
```

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: afaik it is only reference, right? I am caching this secondly because it will be more efficient than calling it each time in `PublishHealthInfoAsync`

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: Ok, it explains also why `Action` didn't have problem with two versions (with(out) `async`)

mvxxx: I think that I would stay with that functional approach because `reportHealth` can change, so it would be passed each time.

src/Extensions/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -16,9 +16,9 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 	public static class ServiceCollectionExtensions
 	{
 		/// <summary>
-		/// Register types required for Omex health check
+		/// Add fdependencies for a publisher
```
AndreyTretyak, line 19: nit: what does `fdependencies` mean here?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
AndreyTretyak: Then what's the point of having `RestClientWrapper` class?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,108 @@
...
+
+			if(m_client == null)
+			{
+				m_logger.LogInformation(Tag.Create(), "HTTP Client isn't ready yet.");
+				return;
+			}
```
AndreyTretyak: By definition, your function `GetAsync` has not nullable return type, so I don't think a null check needed here.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
AndreyTretyak: Also is there any point in calling `m_clientWrapper.GetAsync()` each time instead of checking if it's already in `m_client` variable?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,108 @@
...
+
+			return sfcHealthInfo;
+		}
+
+		internal async Task PublishHealthInfoAsync(ServiceFabricHealth.HealthInformation healthInfo)
+		{
+			ServiceFabricCommon.HealthInformation sfcHealthInfo = FromSfHealthInformation(healthInfo);
+
+			// We know that during this call m_client is not null
+			await m_client!.Nodes.ReportNodeHealthAsync(
```
AndreyTretyak, line 94: As a rule, we are not using null suppression operator (`!`) in production code, and actually, here it could be null if function call order would change especially since you are putting it into Action. I would suggest just doing a null check here, it would be more reliable.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
AndreyTretyak: I'm not following your reasoning here, what do you mean by reasoning `reportHealth` can change, and how abstract method definition would be different from Activity here? (aside from the fact that Activity creates delegate object and closure each time).

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: RestClientWrapper is just abstraction used for more elegant solution

mvxxx: Ad. https://github.com/microsoft/Omex/pull/334#discussion_r621062629
If we move null checking and recreating from wrapper to publisher, then there is no need for wrapper, but I think that this is more elegant but I'm open to change that if needed.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: I mean that in order to create `reportHealth` from `			Action<SfHealthInformation> reportHealth = partition switch...` we need partition accessor. At this point when I'm looking at this I see possibility to create such override implementation but this need some reimplementation, I think. Should I do this or functional approach is fine?

AndreyTretyak: Okay, I'm fine with keeping it in Action for now, as long as we understand the downsides, we can improve it later.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
AndreyTretyak: As discussed offline this wrapper might be useful for other tasks and would be extracted into the Abstractions package in the follow-up.

AndreyTretyak (approved):

dimeehan (dismissed):

Dependencies.props
```diff
@@ -44,6 +45,7 @@
   </PropertyGroup>
   <PropertyGroup Label="ServiceFabric packages for Omex Extensions">
     <Microsoft_ServiceFabric_Version>7.2.434</Microsoft_ServiceFabric_Version>
+    <Microsoft_ServiceFabric_Client_Http_Version>4.2.0</Microsoft_ServiceFabric_Client_Http_Version>
```
aszczepanski, line 48: You might be able to start with the latest version - 4.4.0 :)

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.ObjectPool;
+using ServiceFabricHealth = System.Fabric.Health;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	internal abstract class HealthCheckPublisher : IHealthCheckPublisher
```
aszczepanski, line 16: Make sure you add unit tests

aszczepanski (changes requested):
Please add a description to the PR

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: HealthCheckPublisher is covered by currently existing tests

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,105 @@
...
+		{
+			ServiceFabricCommon.HealthInformation sfcHealthInfo = FromSfHealthInformation(healthInfo);
+
+			if (m_client == null)
+			{
+				m_logger.LogInformation(Tag.Create(), "HTTP Client isn't ready yet.");
+				return;
+			}
+
+			await m_client.Nodes.ReportNodeHealthAsync(
```
aszczepanski, line 93: Why are we reporting node health here? I believe it's supposed to be used for node-level health, and here we're monitoring service replicas  or services

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceFabricHealthCheckPublisher.cs
```diff
@@ -2,43 +2,35 @@
...
-			ObjectPoolProvider objectPoolProvider)
+			ObjectPoolProvider objectPoolProvider) : base(objectPoolProvider)
 		{
 			m_partitionAccessor = partitionAccessor;
 			m_logger = logger;
-			m_stringBuilderPool = objectPoolProvider.CreateStringBuilderPool();
 		}
 
-		public Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
+		internal static new string HealthReportSourceId => nameof(ServiceFabricHealthCheckPublisher);
```
AndreyTretyak, line 31: As a note for a future fix, If I understand everything correctly we should not use `new` here, instead we need to mark property in the base class as `virtual` and use `override` here.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: Maybe we can add tests for RestPublisher - in that case I think that we should reuse courrent tests, maybe inherit them? Not sure, probably worth checking.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,105 @@
...
+				nodeName: m_nodeName,
```
aszczepanski, line 94: What's the retry logic here? Is it the same as when reporting health using regular fabric client?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: Yes. Currently this is workaround. Probably we are not able to report directly to replicas/instances as long as replicaId and partitionId are not exposed in SF Environment.

aszczepanski, line 93: In terms of sending health reports - could you please check if fabirc client doesn't send health reports in batches?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: I think that this is regular fabric client

Dependencies.props
dimeehan: how exciting!

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		protected ServiceFabricHealth.HealthInformation BuildSfHealthInformation(string healthCheckName, HealthReportEntry reportEntry)
+		{
+			StringBuilder descriptionBuilder = StringBuilderPool.Get();
+
+			if (!string.IsNullOrWhiteSpace(reportEntry.Description))
+			{
+				descriptionBuilder.Append("Description: ").Append(reportEntry.Description).AppendLine();
+			}
+			descriptionBuilder.Append("Duration: ").Append(reportEntry.Duration).Append('.').AppendLine();
```
aszczepanski, line 110: Would it make sense to also include the current correlation id/activity id here? Or is it expected to be part of the description?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,105 @@
...
+				healthInformation: sfcHealthInfo);
+		}
+
+		internal string? FindNodeName() => Environment.GetEnvironmentVariable(NodeNameVariableName);
```
aszczepanski, line 98: How will this variable be set?

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: This part existed before - This PR convers only making rest health publisher and doesn't change old functionality.

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: I'm not sure, what do you mean?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
mvxxx: I don't found logic responsible for that

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,123 @@
...
+
+		internal static string HealthReportSourceId => nameof(HealthCheckPublisher);
+
+		internal const string HealthReportSummaryProperty = "HealthReportSummary";
+
+		public HealthCheckPublisher(ObjectPoolProvider objectPoolProvider) => StringBuilderPool = objectPoolProvider.CreateStringBuilderPool();
+
+		public abstract Task PublishAsync(HealthReport report, CancellationToken cancellationToken);
+
+		protected abstract ServiceFabricHealth.HealthInformation FinalizeHealthReport(string healthReportSummaryProperty, ServiceFabricHealth.HealthState healthState);
```
AndreyTretyak, line 28: Also for my information, could you please explain what's the point of making this method abstract, it looks identical in both cases?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
AndreyTretyak: This variable provided by SF, in the future it would be moved to IExecutionContext https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-environment-variables-reference#:~:text=Service%20Fabric%20environment%20variables%20%20%20%20Environment,%20%208234%20%2011%20more%20rows%20

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
AndreyTretyak: I think retries and batching are done by the `Microsoft.ServiceFabric.Client.Http` library.

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceFabricHealthCheckPublisher.cs
AndreyTretyak: oh sorry, I missed static here, then I think it would be better to just make it not static.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
mvxxx: Yes, sure - it is only for taking proper Publisher Source Id  because static members can't be abstract

AndreyTretyak: In this case, I think just making Publisher Source Id property non-static and abstract would be much better, than splitting this method.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,123 @@
...
+						degradedEntries++;
+						break;
+					case HealthStatus.Unhealthy:
+					default:
+						unhealthyEntries++;
+						break;
+				}
+			}
+
+			StringBuilder descriptionBuilder = StringBuilderPool.Get();
```
artempushkin, line 75: When you are done with the StringBuilder you should return it to the pool.

artempushkin (changes requested):

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,132 @@
...
```
AndreyTretyak, line 77: I don't see this builder returned to the pool

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
AndreyTretyak: I still don't see this builder returned to the pool, why thread resolved?

AndreyTretyak: Sorry, I was looking on old version

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+using Microsoft.ServiceFabric.Client;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+	internal class RestClientWrapper
+	{
+		private IServiceFabricClient? m_client;
+
+		private readonly Uri? m_clusterEndpoint;
```
AndreyTretyak, line 14: Why this one nullable?

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
```diff
@@ -0,0 +1,111 @@
...
+
+		private string? m_nodeName;
+
+		internal override string HealthReportSourceId => nameof(RestHealthCheckPublisher);
+
+		internal const string NodeNameVariableName = "Fabric_NodeName";
+
+		public RestHealthCheckPublisher(ILogger<RestHealthCheckPublisher> logger,
+										IOptions<RestHealthCheckPublisherOptions> options,
+										ObjectPoolProvider objectPoolProvider) : base(objectPoolProvider)
```
AndreyTretyak, line 34: nit: we usually use a single tab for argumetns on a new line, please fix here and in other methods.

src/Extensions/Diagnostics.HealthChecks/Internal/HealthCheckPublisher.cs
```diff
@@ -0,0 +1,141 @@
...
+			string description;
+			try
+			{
+
+
```
dimeehan, line 82: nit: remove empty lines

src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs
mvxxx: beucase there is a ctor which accepts only client

mvxxx: Generally it will be changed in next PR with more generic wrappers

dimeehan: One thing, if something will be addressed in a separate PR, we usually add a TODO comment with the work item ID so it's tracked, for example: https://office.visualstudio.com/OC/_git/OmexMarketplace?path=%2Fsrc%2FOmexAddInsLicensing%2FAddInsLicensing.AcquireExperience%2FExperience%2FOutlookExperience.cs&version=GBmaster&line=106&lineEnd=107&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

mvxxx: Ok, done https://github.com/mvxxx/Omex/blob/rest-publisher/src/Extensions/Diagnostics.HealthChecks/Internal/RestClientWrapper.cs#L10

artempushkin (approved):

aszczepanski (approved):

src/Extensions/Diagnostics.HealthChecks/Internal/RestHealthCheckPublisher.cs
aszczepanski: Might be worth adding a code comment with that info

PR #335, opened 2021-04-23
Using memory for async methods

dyllew3:
Removing the use of Span in async methods and replacing it with memory which allows for the span to be accessed in async methods. Have updated unit tests accordingly, link to docs about memory and span usage https://docs.microsoft.com/en-us/dotnet/standard/memory-and-spans/memory-t-usage-guidelines

### DIFF
--- a/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentityMiddleware.cs
+++ b/src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentityMiddleware.cs
@@ -48,7 +48,8 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 
 		internal async Task<string> CreateUserHashAsync(HttpContext context)
 		{
-			using IMemoryOwner<byte> uidMemoryOwner = MemoryPool<byte>.Shared.Rent(m_maxIdentitySize + m_saltProvider.GetSalt().Length);
+			int len = m_saltProvider.GetSalt().Length;
+			using IMemoryOwner<byte> uidMemoryOwner = MemoryPool<byte>.Shared.Rent((m_maxIdentitySize + m_saltProvider.GetSalt().Length));
 
 			// Done because span cannot be declared in an async function
 			uidMemoryOwner.Memory.Span.Fill(0);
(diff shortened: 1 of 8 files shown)


src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentityMiddleware.cs
```diff
@@ -48,7 +48,8 @@ async Task IMiddleware.InvokeAsync(HttpContext context, RequestDelegate next)
 
 		internal async Task<string> CreateUserHashAsync(HttpContext context)
 		{
-			using IMemoryOwner<byte> uidMemoryOwner = MemoryPool<byte>.Shared.Rent(m_maxIdentitySize + m_saltProvider.GetSalt().Length);
+			int len = m_saltProvider.GetSalt().Length;
```
AndreyTretyak, line 51: I've noticed that we are getting Salt too many times in this method (not the fault of this PR). Could you please try to change it to get salt once and reuse it?

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentityMiddleware.cs
```diff
@@ -57,7 +58,7 @@ internal async Task<string> CreateUserHashAsync(HttpContext context)
 			foreach (IUserIdentityProvider provider in m_userIdentityProviders)
 			{
 				bool success;
-				(success, identityBytesWritten) = await provider.TryWriteBytesAsync(context, uidMemoryOwner.Memory.Span.Slice(0, provider.MaxBytesInIdentity))
+				(success, identityBytesWritten) = await provider.TryWriteBytesAsync(context, uidMemoryOwner.Memory)
```
AndreyTretyak, line 61: Since we how giving identity providers access to the whole memory we probably should call `.Fill(0)` (similar to how it's done above) when `success` is `false`. To avoid the case when identity provider filled memory with something and returned `false`.

src/Extensions/Hosting.Services.Web/Middlewares/UserIdentity/UserIdentityMiddleware.cs
```diff
@@ -57,17 +57,23 @@ internal async Task<string> CreateUserHashAsync(HttpContext context)
...
+				{
+					uidMemoryOwner.Memory.Span.Fill(0);
+				}
 			}
 
-			return GetHashString(m_saltProvider.GetSalt(), uidMemoryOwner.Memory.Span, identityBytesWritten);
+			return GetHashString(m_salt, uidMemoryOwner.Memory.Span, identityBytesWritten);
 		}
 
+		private ReadOnlySpan<byte> m_salt => m_saltProvider.GetSalt();
```
AndreyTretyak, line 75: I think it would be better to expose `MaxBytesInSalt` property on the `ISaltProvider` to avoid double calls to `GetSalt()`.

dyllew3: Have added MaxBytesInSalt

aszczepanski (approved):

AndreyTretyak (approved):

PR #341, opened 2021-05-06
Replace old service fabric client in RestHealthPublisher

mvxxx:
# Description
Removed old RestClientWrapper with service fabric client from `Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions`.

# Tests
Tested manually on previously prepared projects, all unit tests passes, there is no need for additional ones.


tests/Extensions/Diagnostics.HealthChecks.UnitTests/HealthCheckPublisherTests.cs
```diff
@@ -474,7 +475,8 @@ public override void Configure(bool closed = false)
 
 		internal class RestPublisherTestContext : PublisherTestContext
 		{
-			public Mock<IServiceFabricClient>? SFClient { get; protected set; } // TODO: move to RestTestContext
+			public Mock<IServiceFabricClient>? SFClient { get; protected set; }
+			public Mock<IServiceFabricClientWrapper>? SFClientWrapper { get; protected set; }
```
AndreyTretyak, line 479: nit: I don't see any point in making this property public or even having those properties for now, especially since Publisher field internal for some reason.

AndreyTretyak (approved):

AndreyTretyak (approved):

PR #342, opened 2021-05-06
Service Fabric Rest Client Factory

mvxxx:
# Description
Minor change which exposes factory which allows us to create instance of `IServiceFabricClientWrapper`. It is especially useful when we want to get client wrapper without DI.

# Tests
Covered via unit testing

### DIFF
--- a/src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
+++ b/src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions
 		/// Creates service fabric client wrapper instance and returns it
 		/// </summary>
 		/// <param name="options">REST client options</param>
-		/// <returns></returns>
+		/// <returns>Service Fabric Client Wrapper</returns>
 		public static IServiceFabricClientWrapper CreateServiceFabricClientWrapper(IOptions<ServiceFabricRestClientOptions> options)
 		{
 			return new ServiceFabricClientWrapper(options);
--- a/src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
+++ b/src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions
+{
+	/// <summary>
+	/// Factory for service fabric client wrapper
+	/// </summary>
+	public class ServiceFabricClientFactory
+	{
+		/// <summary>
+		/// Creates service fabric client wrapper instance and returns it
+		/// </summary>
+		/// <param name="options">REST client options</param>
+		/// <returns></returns>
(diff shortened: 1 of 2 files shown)


src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// <summary>
+	/// Factory for service fabric client wrapper
+	/// </summary>
+	public class ServiceFabricClientFactory
+	{
+		/// <summary>
+		/// Creates service fabric client wrapper instance and returns it
+		/// </summary>
+		/// <param name="options">REST client options</param>
+		/// <returns></returns>
```
dyllew3, line 17: nit: empty returns

src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+		public static IServiceFabricClientWrapper CreateServiceFabricClientWrapper(IOptions<ServiceFabricRestClientOptions> options)
```
AndreyTretyak, line 18: This method would be used when there is no way of getting client wrapper from DI, it's better to use `ServiceFabricRestClientOptions` directly as a parameter or even ClusterEndpoint directly

src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
AndreyTretyak: nit: if you are not adding description to return then just remove it

src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+	/// Factory for service fabric client wrapper
+	/// </summary>
+	public class ServiceFabricClientFactory
+	{
+		/// <summary>
+		/// Creates service fabric client wrapper instance and returns it
+		/// </summary>
+		/// <param name="options">REST client options</param>
+		/// <returns>Service Fabric Client Wrapper</returns>
+		public static IServiceFabricClientWrapper CreateServiceFabricClientWrapper(IOptions<ServiceFabricRestClientOptions> options)
```
AndreyTretyak, line 18: nit: since your factory already named `ServiceFabricClientFactory` maybe we could name method just `Create`, but it's up to you.

tests/Extensions/ServiceFabricGuest.Abstractions.UnitTests/FactoryTests.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions.UnitTests
+{
+	[TestClass]
+	public class FactoryTests
```
AndreyTretyak, line 12: nit: We usually call UT class with the name of the class we task + Tests sufix

src/Extensions/ServiceFabricGuest.Abstractions/factory/ServiceFabricClientFactory.cs
mvxxx: By passing ServiceFabricRestClientOptions I wanted hide the way how it is already created (Maybe in the future there will be something other needed that URI of cluster endpoint). I replaced that with ServiceFabricRestClientOptions

AndreyTretyak (approved):

dimeehan (approved):

PR #352, opened 2021-05-17
Service Fabric Config Provider

mvxxx:
## Description
- Removed `ServiceContext` dependency in `Microsoft.Omex.Extensions.Diagnostics.HealthChecks`
- Moved `SfConfigurationProvider` and `SfConfigurationException` to `Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions`
- Updated unit tests (old ones were using service context for health checks).


src/Extensions/ServiceFabricGuest.Abstractions/SfConfigurationProvider.cs
```diff
@@ -2,22 +2,35 @@
...
-namespace Microsoft.Omex.Extensions.Hosting.Services.Web
-{
-	internal class SfConfigurationProvider
+
+namespace Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions
+{
+	/// <summary>
+	/// Configuration provider for service fabric environment variables
+	/// </summary>
+	public class SfConfigurationProvider
```
artempushkin, line 11: Is this class designed to be used only from guest executables?

mvxxx: Generally it can be used from any type of service, especially from guest executables due to lack of dependency to service context

AndreyTretyak: We also use it to create WebService with a single host, because we need to provide endpoint port to Kestrel before ServiceContext available. I think in the future we would move all interactions with SF environment variables into it.

tests/Extensions/Hosting.Services.Web.UnitTests/TestHelpers/SfConfigurationProviderHelper.cs
```diff
@@ -2,15 +2,20 @@
...
 using System;
-
+using Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions;
+
 namespace Microsoft.Omex.Extensions.Hosting.Services.Web.UnitTests
 {
 	internal static class SfConfigurationProviderHelper
-	{
+	{
+		/// SF environment variables taken from https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-environment-variables-reference
```
AndreyTretyak, line 11: This comment should be in `SfConfigurationProvider`. And here we should point to that class instead since values should be instead (alternatively we can access internal fields from of SfConfigurationProvider here by moving class to Abstractions UnitTest and then referencing it).

AndreyTretyak (approved):
Looks good, but please fix the comments.

tests/Extensions/Hosting.Services.Web.UnitTests/TestHelpers/SfConfigurationProviderHelper.cs
mvxxx: Done

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
```diff
@@ -4,25 +4,24 @@
...
 using Moq;
-using ServiceFabric.Mocks;
 
 namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks.UnitTests
 {
 	[TestClass]
 	public class HttpEndpointHealthCheckTests
-	{
+	{
+		internal const string PublishAddressEvnVariableName = "Fabric_NodeIPOrFQDN";
```
AndreyTretyak, line 22: Should not you use SfConfigurationProviderHelper here? do avoid duplicating constants?

tests/Extensions/Diagnostics.HealthChecks.UnitTests/HttpEndpointHealthCheckTests.cs
```diff
@@ -215,34 +212,22 @@ public async Task CheckHealthAsync_WithAdditionalCheck_ReturnsOverridenResult()
 		private async Task<(MockClient, HealthCheckResult)> RunHealthCheckAsync(
 			HttpHealthCheckParameters parameters, HttpResponseMessage response, HealthStatus failureStatus = HealthStatus.Unhealthy)
 		{
-			Accessor<ServiceContext> accessor = new Accessor<ServiceContext>();
-			Mock<IHttpClientFactory> factoryMock = new Mock<IHttpClientFactory>();
-			MockClient clientMock = new MockClient(response);
+			Mock<IHttpClientFactory> factoryMock = new();
+			MockClient clientMock = new(response);
+			Environment.SetEnvironmentVariable(SfConfigurationProviderHelper.EndpointPortEvnVariableSuffix + parameters.EndpointName, 6789.ToString(), EnvironmentVariableTarget.Process);
```
AndreyTretyak, line 217: Why not `SfConfigurationProviderHelper.SetPortVariable`?

AndreyTretyak (approved):

dimeehan (approved):

PR #361, opened 2021-05-26
HttpEndpointHealthcheck enhancements

vh-vahan:


src/Extensions/Diagnostics.HealthChecks/Internal/HttpHealthCheckParameters.cs
```diff
@@ -12,56 +11,26 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
...
-			Method = method ?? HttpMethod.Get;
-
-			Scheme = scheme == null
-				? Uri.UriSchemeHttp
-				: Uri.CheckSchemeName(scheme)
-					? scheme
-					: throw new ArgumentException("Invalid uri scheme", nameof(scheme));
-
-			Headers = headers ?? s_emptyHeaders;
+			RequestMessage = httpRequestMessage == null ? throw new ArgumentNullException(nameof(httpRequestMessage)) : httpRequestMessage;
```
AndreyTretyak, line 33: This class is internal so the only usage would be in this project with nullability enabled, so I think just having a non-nullable type is enough. But if you insist then shorter form could be used like `RequestMessage = httpRequestMessage ?? throw new ArgumentNullException(nameof(httpRequestMessage));`.

AndreyTretyak (approved):
Looks great, much more flexible, and shorter. Thanks for doing it.

AndreyTretyak (approved):

src/Extensions/Diagnostics.HealthChecks/HealthChecksBuilderExtensions.cs
```diff
@@ -40,9 +42,73 @@ public static class HealthChecksBuilderExtensions
...
+		/// <param name="expectedStatus">response status code that considered healthy, default to 200(OK)</param>
+		/// <param name="additionalCheck">action that would be called after getting response, function should return new result object that would be reported</param>
+		/// <param name="reportData">additional properties that will be attached to health check result, for example escalation info</param>
+		public static IHealthChecksBuilder AddHttpEndpointCheck(
+			this IHealthChecksBuilder builder,
+			string name,
+			string endpointName,
+			Func<UriBuilder, HttpRequestMessage> httpRequestMessageBuilder,
+			HttpStatusCode? expectedStatus = null,
+			Func<HttpResponseMessage, HealthCheckResult, HealthCheckResult>? additionalCheck = null,
```
MisterBoum, line 102: Is it worth it making this async friendly?

MisterBoum (dismissed):
:shipit:

vskh:
Can you add the description please?

vskh (changes requested):
PR is missing description.

src/Extensions/Diagnostics.HealthChecks/HealthChecksBuilderExtensions.cs
AndreyTretyak: For `httpRequestMessageBuilder` I hope it's not needed since it's mostly for configuring additional properties of the request. And the fact that we are invoking it in this method would force us to make it async also, so no.

But it's a really good idea for `additionalCheck` since you might need to get content there and all of the methods for it are async. It's a bit out of the scope of this PR, but definitely something we need to do. Thank you.

@vh-vahan  I'll live it up to you to include the change in this PR or I'll create an issue and we can handle it separately.

AndreyTretyak: Thanks. Good point about breaking change. I'll send a note about it.

AndreyTretyak (approved):

vskh (approved):

PR #364, opened 2021-06-01
Using the cloned httprequest message in HttpEndpointHealthcheck

vh-vahan:
HttpEndpointHealthcheck will clone the configured HttpRequestMessage from the parameters before sending the request due to the restriction of the same http request message not being allowed to send more then once.

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
@@ -85,9 +85,14 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 #if !NETCOREAPP3_1 && !NETSTANDARD2_0
 			clone.VersionPolicy =  message.VersionPolicy;
 
-			foreach (KeyValuePair<string, object?> item in message.Options)
+			foreach (KeyValuePair<string, object?> option in message.Options)
 			{
-				clone.Options.Set(new HttpRequestOptionsKey<object?>(item.Key), item.Value);
+				clone.Options.Set(new HttpRequestOptionsKey<object?>(option.Key), option.Value);
+			}
+#else
+			foreach (KeyValuePair<string, object> prop in message.Properties)
+			{
+				clone.Properties.Add(prop.Key, prop.Value);
 			}
 #endif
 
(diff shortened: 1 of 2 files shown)


src/Extensions/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
```diff
@@ -65,5 +66,23 @@ protected override async Task<HealthCheckResult> CheckHealthInternalAsync(Health
...
+				Content = message.Content,
+				Version = message.Version
+			};
+
+			foreach (KeyValuePair<string, IEnumerable<string>> header in message.Headers)
+			{
+				clone.Headers.TryAddWithoutValidation(header.Key, header.Value);
+			}
+
+			return clone;
```
AndreyTretyak, line 85: Should we also copy [Options](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httprequestoptions?view=net-5.0) and [VersionPolicy ](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httprequestmessage.versionpolicy?view=net-5.0#System_Net_Http_HttpRequestMessage_VersionPolicy)or we are not planning to use them?

AndreyTretyak (approved):

AndreyTretyak (approved):

src/Extensions/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
```diff
@@ -65,5 +66,32 @@ protected override async Task<HealthCheckResult> CheckHealthInternalAsync(Health
...
+				Content = message.Content,
+				Version = message.Version
+			};
+
+			foreach (KeyValuePair<string, IEnumerable<string>> header in message.Headers)
+			{
+				clone.Headers.TryAddWithoutValidation(header.Key, header.Value);
+			}
+
+#if !NETCOREAPP3_1 && !NETSTANDARD2_0
```
vlad-ion, line 85: Should this be || instead of &&? I am thinking the project can't be compiled for both at the same time but I might miss something here.

vlad-ion (approved):

src/Extensions/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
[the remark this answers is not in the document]
vh-vahan: Adding the code to clone the props as well

vlad-ion (approved):

AndreyTretyak (approved):

PR #372, opened 2021-06-10
Remove task extension for activity

AndreyTretyak:
It looks like they are not useful enough to justify supporting them, especially taking into account simplified using syntax that could handle similar cases without any performance degradation and would also take into account time before the task actually started.


neilr81:
Are we using these anywhere already?

neilr81 (approved):

AndreyTretyak:
> Are we using these anywhere already?

only in bootcamp, it should be easy to update

K-Cully (approved):

dimeehan (approved):

PR #378, opened 2021-06-28
Refactoring before splitting HealthCheck package

AndreyTretyak:
Remove publishers inheritance hierarchy by extracting SF depended logic into a separate interface.

As a side effects of the change:
* Abstraction package now has a polifil to use records in .net framework and .net core
* `IExecutionContext` now has `NodeName`


src/Extensions/Diagnostics.HealthChecks/Internal/OmexHealthCheckPublisher.cs
```diff
@@ -11,32 +11,36 @@
...
 
-		public HealthCheckPublisher(ObjectPoolProvider objectPoolProvider) => StringBuilderPool = objectPoolProvider.CreateStringBuilderPool();
+		public OmexHealthCheckPublisher(IHealthStatusSender statusSender, ObjectPoolProvider objectPoolProvider)
+		{
+			StringBuilderPool = objectPoolProvider.CreateStringBuilderPool();
+			StatusSender = statusSender;
+		}
 
-		public abstract Task PublishAsync(HealthReport report, CancellationToken cancellationToken);
+		public async Task PublishAsync(HealthReport report, CancellationToken cancellationToken)
```
K-Cully, line 28: please decorate public methods

src/Extensions/Diagnostics.HealthChecks/IHealthStatusSender.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+	/// <summary>
+	/// Interface for sending health status from HealthPublisher to extenal consumer (ex. SF Cluster).
+	/// </summary>
+	public interface IHealthStatusSender
+	{
+		/// <summary>
+		/// Initialization method that's called before each health status publishing session.
+		/// </summary>
+		Task IntializeAsync(CancellationToken token);
```
K-Cully, line 18: I take it the token is just added to enforce standard convention and avoid potential future refactoring?
*It's not used in the concrete implementation below.

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceContextHealthStatusSender.cs
```diff
@@ -0,0 +1,85 @@
...
+			}
+
+			return Task.CompletedTask;
+		}
+
+		public Task SendStatusAsync(string checkName, HealthStatus status, string description, CancellationToken token)
+		{
+			if (m_reportHealth == null)
+			{
+				m_logger.LogWarning(Tag.Create(), "Publisher run before health report functions is provided.");
```
K-Cully, line 56: I'm not clear on what this message is trying to convey.
Should it be "The status sender was executed before a report was generated." or something like that?

K-Cully, line 56: Should a cancelled task be returned in this case instead of a completed task?

K-Cully (changes requested):

src/Extensions/Diagnostics.HealthChecks/IHealthStatusSender.cs
AndreyTretyak: Good point, I've found a way to propagate `CancelationToken` into client creation.

src/Extensions/Diagnostics.HealthChecks/Internal/OmexHealthCheckPublisher.cs
AndreyTretyak: Class is Internal and the interface used for accessing it provides documentation. Do you think we should add something more here?

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceContextHealthStatusSender.cs
AndreyTretyak: Updated message and logic to throw exception instead

src/Extensions/Abstractions/SfConfigurationException.cs
```diff
@@ -1,21 +1,21 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license.
-
-using System;
-using System.Globalization;
-
-namespace Microsoft.Omex.Extensions.Abstractions
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
artempushkin, line 1: What are the differences here? Whitespace, invisible chars? It's making the PR hard to review.

AndreyTretyak: Yes, for multiple files, VS forced me to normalize line endings

AndreyTretyak: I assume it would be the same for everybody so I've decided it's better to commit them.

artempushkin: In the future please try to commit cosmetic changes as separate PRs.

src/Extensions/Diagnostics.HealthChecks/Internal/OmexHealthCheckPublisher.cs
K-Cully: Didn't realise the class was internal.
It's probably okay as is.

src/Extensions/Diagnostics.HealthChecks/Internal/ServiceContextHealthStatusSender.cs
```diff
@@ -0,0 +1,80 @@
...
+			else
+			{
+				// Avoiding repeated pattern matching for each report entry.
+				m_reportHealth = partition switch
+				{
+					IStatefulServicePartition statefulPartition => statefulPartition.ReportReplicaHealth,
+					IStatelessServicePartition statelessPartition => statelessPartition.ReportInstanceHealth,
+					_ => throw new ArgumentException($"Service partition type '{partition.GetType()}' is not supported."),
+				};
+			}
```
K-Cully, line 47: I feel like there should be a way to bring the null check into the switch expression...

K-Cully: something like
`
_ when partition is null => Log
`

K-Cully (approved):

artempushkin (approved):

PR #381, opened 2021-07-02
Added ProductVersion constructor with nullable inputs

amandal1810:
In order to make type casting easier and uniform across internal repos of Omex

### DIFF
--- a/src/System/Model/Types/ProductVersion.cs
+++ b/src/System/Model/Types/ProductVersion.cs
@@ -110,6 +110,27 @@ namespace Microsoft.Omex.System.Model.Types
 		}
 
 		/// <summary>
+		/// Initializes a new instance of the ProductVersion class using the specified
+		/// major, minor, build and revision values.
+		/// </summary>
+		/// <param name="major">The major version number.</param>
+		/// <param name="minor">The minor version number.</param>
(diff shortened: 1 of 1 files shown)


src/System/Model/Types/ProductVersion.cs
```diff
@@ -109,6 +109,27 @@ public ProductVersion(int major, int minor, int build, int revision)
...
 
+		/// <summary>
+		/// Initializes a new instance of the ProductVersion class using the specified
+		/// major, minor, build and revision values.
+		/// </summary>
+		/// <param name="major">The major version number.</param>
+		/// <param name="minor">The minor version number.</param>
+		/// <param name="build">The build number (nullable)</param>
+		/// <param name="revision">The revision number (nullable)</param>
+		public ProductVersion(int major, int minor, int? build, int? revision)
```
aszczepanski, line 120: The other constructor with non-nullable parameters should probably be removed

amandal1810: But that might a breaking change for consumers. So i kept it

amandal1810: Removed it

aszczepanski (approved):

vlad-ion (approved):

PR #384, opened 2021-08-03
Add Error for project that depend on Microsoft.Omex.System

AndreyTretyak:
This PR adds a build error for projects that reference `Microsoft.Omex.System` or packages that depend on it.

Idea is that projects that reference them would need to add special flag `<UseUnsupportedOmexSystemPackage>true</UseUnsupportedOmexSystemPackage>` with a comment to PBI about fixing it and plan for removing a dependency. 

Initial PR would create warning instead of error and we would switch after all consumers updated

### DIFF
--- a/src/System/BuildTransitive/Microsoft.Omex.System.props
+++ b/src/System/BuildTransitive/Microsoft.Omex.System.props
@@ -1,0 +1,13 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <!-- Enforce nullability check and language version -->
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <Target Name="RunValidations" AfterTargets="Build">
+    <Error
+        Condition="'$(UseUnsupportedOmexPackage)' != 'true'"
+        Code="OMEX0101"
+        HelpKeyword="DoNotUseUnsupportedPackages"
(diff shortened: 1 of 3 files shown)


src/System/BuildTransitive/Microsoft.Omex.System.props
```diff
@@ -0,0 +1,13 @@
...
+    <!-- Enforce nullability check and language version -->
+    <LangVersion>latest</LangVersion>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <Target Name="RunValidations" AfterTargets="Build">
+    <Error
+        Condition="'$(UseUnsupportedOmexPackage)' != 'true'"
+        Code="OMEX0101"
+        HelpKeyword="DoNotUseUnsupportedPackages"
+        Text="Mirosoft.Omex.System, Microsoft.OMEX.ServiceFabric.Core and packages dependend on them would be out of support after 30 of September 2021. You can temporary use flag &lt;UseUnsupportedOmexPackage&gt;true&lt;/UseUnsupportedOmexPackage&gt;, but add comment with PBI to fix it." />
```
K-Cully, line 12:
```suggestion
        Text="Mirosoft.Omex.System, Microsoft.OMEX.ServiceFabric.Core and packages depending on them will be out of support after the 30th of September 2021. You can temporarily use flag &lt;UseUnsupportedOmexPackage&gt;true&lt;/UseUnsupportedOmexPackage&gt; to suppress, but add a reference to the follow work item to fix it." />
```

K-Cully (changes requested):

K-Cully (dismissed):

neilr81 (approved):

K-Cully (approved):

PR #386, opened 2021-08-06
Add extension method to identify Performance testing on activity

martinpe36:


src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -37,6 +37,24 @@ public static class ActivityExtensions
 				HealthCheckMarkerValue,
 				StringComparison.OrdinalIgnoreCase);
 
+		/// <summary>
+		/// Returns true if activity is marked as Performance test
+		/// </summary>
+		/// <param name="activity"></param>
+		/// <returns></returns>
+		public static bool IsPerformanceTest(this Activity activity) =>
```
AndreyTretyak, line 45: How are you planning to add this flag? Maybe we need the corresponding method `MarkAsPerformanceTest` and it's done for the health check marker?

Also could you please add some explanation of why this marker is needed and how it would be used?

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -37,6 +37,24 @@ public static class ActivityExtensions
...
+			!string.IsNullOrEmpty(activity.GetBaggageItem(PerformanceMarkerKey));
+
+		/// <summary>
+		/// Returns the value of the Performance key if it exists
+		/// </summary>
+		/// <param name="activity"></param>
+		/// <returns></returns>
+		public static string PerformanceValue(this Activity activity)
```
AndreyTretyak, line 53: Why this should be a separate method and why it's public, would marker value have some significance or it's just a flag?

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -129,6 +147,7 @@ public static class ActivityExtensions
 		private const string UserHashKey = "UserHash";
 		private const string HealthCheckMarkerKey = "HealthCheckMarker";
 		private const string HealthCheckMarkerValue = "true";
+		private const string PerformanceMarkerKey = "PerformanceMarker";
```
AndreyTretyak, line 150:
```suggestion
		private const string PerformanceMarkerKey = "PerformanceTestMarker";
```
Maybe naming it `PerformanceTestMarker` would be more clear? (depending on the usage)

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
K-Cully: Yes, please make sure anything going into public-facing repos is documented to the standard you'd expect if you were referencing an external library.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
martinpe36: The idea for the performance marker is that an int value can be passed instead of just a boolean value (true or false).  This int value could be used to mock a response time for a partner call, e.g. PerfomanceMaker=1000 would mock a 1000 millisecond response from a partner service to see how the calling service would handle such a delay.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
martinpe36: This flag will be added by specifying the value "PerformanceMarker=0" or "PerformanceMarker=500" for the Header Correlation-Context.  I don't think we need a method 'MarkAsPerformanceTest' as the header value will be added manually when running performance tests.

This header value will be specified when running performance tests which will by-pass calls to partner services.

martinpe36: @K-Cully  Is there a location where this should be documented?  Or should documentation be added to the XML tag for the method?

jj05y (approved):

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: XML for methods would the best place

AndreyTretyak: Let's join discussions here https://github.com/microsoft/Omex/pull/386#discussion_r691145343

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
AndreyTretyak: I'm not sure that I fully understand the idea. 
1. If this value is intended to be used as configuration settings only for mocks that would emulate delay from external service, why could not this configuration be set in the mock itself without transferring additional parameters?
2. What would be the benefit of having this functionality in shared code? Would it be needed for many services or some specific use case? We would probably need more description and explanation of how other people could use it. I do not fully understand what mocks are we talking about. Maybe you could point to some sample usage, please?
3. If we are going forward with this approach it might be better for `PerformanceValue` method to return `int` or `int?` instead since every consumer would need to parse it. And then we can have a corresponding set method to avoid duplicating string key (and risk of them having different values) in the place where you planning to set the value. And some simple UT would be nice.

@martinpe36  @jj05y  would you like to set up a short call to discuss the best way of doing it?

AndreyTretyak (changes requested):
I would appreciate it if we can discuss the idea behind this change a bit more to improve my understanding.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -37,6 +37,37 @@ public static class ActivityExtensions
...
+			!string.IsNullOrEmpty(activity.GetBaggageItem(PerformanceMarkerKey));
+
+		/// <summary>
+		/// Returns the value of the Performance key if it exists
+		/// </summary>
+		/// <param name="activity"></param>
+		/// <returns>The value passed for PerformanceTestMarker</returns>
+		/// <remarks>Setting the value in the BaggageItems, the interger value can be used to mock a delayed response from a partner call</remarks>
+		/// <remarks>Example: PerformanceTestMarker=500 mocking a 500 millisecond delay</remarks>
+		public static int? PerformanceValue(this Activity activity)
```
AndreyTretyak, line 56: Maybe this name would be more clear:
```suggestion
		public static int? PerformanceTestMakerValue(this Activity activity)
```

AndreyTretyak (dismissed):
I take it you've decided to go with the approach of using maker value instead of DI. 

We'll be discussing a similar problem in the Liffey (hopefully today or tomorrow) about having custom logic in the service for health check, maybe we could find a common solution. 

If possible, please, let me know why have you've decided to use `PerformanceTestMarker` instead of using DI and how are you planning to handle cases when a single request calls multiple partners that need different configurations.

In terms of the code, if you'll decide to keep using those methods I would like to see them covered by unit tests, but I'm fine if you'll do it as a follow-up later.

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -37,6 +37,37 @@ public static class ActivityExtensions
 				HealthCheckMarkerValue,
 				StringComparison.OrdinalIgnoreCase);
 
+		/// <summary>
+		/// Returns true if activity is marked as Performance test
+		/// </summary>
+		/// <param name="activity"></param>
```
K-Cully, line 43: param missing comment

K-Cully (changes requested):
Please add UTs

src/Extensions/Abstractions/Activities/ActivityExtensions.cs
vlad-ion: Same for the next function

K-Cully:
So if I’m understanding correctly, the overall goal driving this change is to identify perf test triggered requests and mock partner calls, right?
Why would enabling the mock have to be done at run-time? - I assume you'd never want to run perf tests alongside real user traffic.
Can't the mocking behaviour just be switched on before a perf test run?

K-Cully (approved):

martinpe36:
@K-Cully The idea behind having the parameter is to pick at runtime so that we don't have to enable an environment to be a perf environment.  As the dependencies are set on startup, the environment would need to be rebooted to be put in 'Perf' mode and then switched back once the performance testing is complete.  Having it set at runtime with a header value, means any environment can be used for perf testing.  Also, if an environment could be put in Perf mode there was a fear this could be done on Production by mistake.

jj05y (approved):

PR #393, opened 2021-08-25
GitHub Reverse Integration - $(GITHUBCOMMIT) 9797533

omex-github-bot-internal:
GitHub reverse integration Pull Request for commit $(GITHUBCOMMIT)
Note: created due to build 'Omex GitHub Official' with id:21.4.10825.11614


dakanev-msft (changes requested):
Title is wrong. I think the pipeline is misconfigured.

K-Cully:
> Title is wrong. I think the pipeline is misconfigured.

Dismissing comment as this is blocking further reverse integrations - If the pipeline has a bug that should be tracked via a separate issue/bug.

PR #397, opened 2021-09-01
Added ThrowIfNull method

amandal1810:
Added `ThrowIfNull` method in `Extensions.Abstractions.Validations`

### DIFF
--- a/src/Extensions/Abstractions/Validation.cs
+++ b/src/Extensions/Abstractions/Validation.cs
@@ -35,5 +35,21 @@ namespace Microsoft.Omex.Extensions.Abstractions
 
 			throw new ArgumentException("String contains only white space characters", name);
 		}
+
+		/// <summary>
+		/// Throw exception if object is null, otherwise returns not-nullable object
+		/// </summary>
+		/// <param name="value">nullable object to check</param>
+		/// <param name="name">name of the object for exception message</param>
+		/// <exception cref="ArgumentNullException">if object is null</exception>
+		[MethodImpl(MethodImplOptions.AggressiveInlining)]
+		public static object ThrowIfNull(object? value, string? name = null)
+		{
+			if (value == null)
+			{
+				throw new ArgumentNullException(message: "Object is null", paramName: name);
+			}
+			return value;
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


aszczepanski (approved):

vlad-ion (approved):

AndreyTretyak (changes requested):
We've explicitly discussed this method with @K-Cully in the original PR and decided not to add it.

You can use `_ = obj ?? throw new ArgumentNullException(nameof(obj));` and in two month there will be `!!` operator in C# for such cases.

Maybe consider adding  `static bool LogIfNull(this ILogger logger, EventId eventId, object? name, string name)` instead, that would log that parameter value is null but won't throw, since we had a similar method in the old Code class. But there still would be some design discussion about it.

K-Cully (changes requested):
Doesn't really align with modern C# coding practices - see Andrey's comment on previous discussion that led to the deliberate omission of this method.

Though I don't really see LogIfNull adding much either. The general case of null values should be caught during continuous integration tests and indicates something critical is wrong. Adding a log vs. catching the exception in an outer exception handler and logging the stack trace, wouldn't add much production debugging value in my opinion - since the stack trace would give as much information.

AndreyTretyak:
> Doesn't really align with modern C# coding practices - see Andrey's comment on previous discussion that led to the deliberate omission of this method.
> 
> Though I don't really see LogIfNull adding much either. The general case of null values should be caught during continuous integration tests and indicates something critical is wrong. Adding a log vs. catching the exception in an outer exception handler and logging the stack trace, wouldn't add much production debugging value in my opinion - since the stack trace would give as much information.

Good point, let's not do `LogIfNull` also, for now, it was needed in some service converting, but I guess it's just an influence of old design, and we should rewrite usage.

amandal1810:
Ok. I think it some what makes sense.

The inspiration for adding this method was a code path which can be described as below:
 - Consider the request model `AppInstallRequest` which contains a property `IdentityProviderType? IdentityType`. 
 - For unauthenticated scenarios, its `null`, for orgid authenticated its `IdentityProviderType.OrgId` and for liveid authenticated its `IdentityProviderType.LiveId`.
 - I validate the `IdentityType` property in the controller class (check if the scenario and value match)
 - Now i need to pass the request object to another method

PR #400, opened 2021-09-16
Use preview of .NET 6 metrics api

AndreyTretyak:
Switch to the release candidate of .NET telemetry package, that allows us:
* Remove reflection hack that we had for .NET 5 to update baggage
* Use new Metrics API to send Activity metrics as a secondary sending for testing (eventually it should replace ETW events)


src/Extensions/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -0,0 +1,96 @@
...
+
+			int tagsCount = s_customTags.Length + activity.TagObjects.Count() + activity.Baggage.Count();
+
+			KeyValuePair<string, object?>[] tags = m_arrayPool.Rent(tagsCount);
+
+			int index = 0;
+
+			foreach (Func<ActivityMetricsSender, Activity, KeyValuePair<string, object?>> getter in s_customTags)
+			{
+				tags[index++] = getter(this, activity);
```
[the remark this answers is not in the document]
AndreyTretyak: it would provide value and then will be incremented, using `++index` would increment and then return.
here demo https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8ACATAAwCwAUEQIwUBuEyABBIwLyMwCmA7owNoBdRgG9GxOIyoBfANx0GjWBk4APKmzFzy9JktWENxLfioBOABQQ+etdmwCAlMbOW+dm4UcygA==

src/Extensions/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -0,0 +1,96 @@
...
+			static (sender, activity) => CreatePair("RegionName", sender.m_context.RegionName),
+			static (sender, activity) => CreatePair("Cluster", sender.m_context.Cluster),
+			static (sender, activity) => CreatePair("ApplicationName", sender.m_context.ApplicationName),
+			static (sender, activity) => CreatePair("ServiceName", sender.m_context.ServiceName),
+			static (sender, activity) => CreatePair("BuildVersion", sender.m_context.BuildVersion),
+			static (sender, activity) => CreatePair("NodeName", sender.m_context.NodeName),
+			static (sender, activity) => CreatePair("MachineId", sender.m_context.MachineId),
+			static (sender, activity) => CreatePair("DeploymentSlice", sender.m_context.DeploymentSlice),
+			static (sender, activity) => CreatePair("IsCanary", sender.m_context.IsCanary),
+			static (sender, activity) => CreatePair("IsPrivateDeployment", sender.m_context.IsPrivateDeployment)
```
artempushkin, line 91: This seems to be Omex-specific, do we need it in the open source libraries?

AndreyTretyak: It's quite generic in the sense that we report all fields from `Activity` and `IExecutionContext` that are also in this repo. It's also more flexible then ETW events that we currently have in the project now.

AndreyTretyak: Having this opensource also brings us closer to n ability to opensource an actual service.

src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -36,9 +36,29 @@ public void SendActivityMetric(Activity activity)
...
+					resultAsString = pair.Value;
+				}
+				else if (string.Equals(ActivityTagKeys.SubType, pair.Key, StringComparison.Ordinal))
+				{
+					subtype = pair.Value;
+				}
+				else if (string.Equals(ActivityTagKeys.Metadata, pair.Key, StringComparison.Ordinal))
+				{
+					metadata = pair.Value;
+				}
```
K-Cully:
```suggestion
				switch (pair.Key)
				{
					case ActivityTagKeys.Result:
						resultAsString = pair.Value;
					case ActivityTagKeys.SubType:
						subtype = pair.Value;
					case ActivityTagKeys.Metadata:
						metadata = pair.Value;
					default:
						break;					
				}

Comment (AndreyTretyak) [reply src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs]: Thanks for the suggestion it definitely makes the code look better, but all changes in this file are reverted now and merging it will create conflict.  
I would like to avoid changing this file. My hope is that we can remove `ActivityEventSender.cs` eventually. Also using switch replaces string compare with regular `Equals`. I don't remember the details but I think `string.Equals(..., ..., StringComparison.Ordinal)` works faster when you compare only ASCII symbols.

File: src/Extensions/Activities/Internal/ActivityMetricsSender.cs [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+			int index = 0;
+
+			foreach (Func<ActivityMetricsSender, Activity, KeyValuePair<string, object?>> getter in s_customTags)
+			{
+				tags[index++] = getter(this, activity);
+			}
+
+			foreach (KeyValuePair<string, string?> baggage in activity.Baggage)
+			{
+				tags[index++] = new KeyValuePair<string, object?>(baggage.Key, baggage.Value);
```
tallalnparis4ev, line 52: Could you use `CreatePair(baggage.Key, baggage.Value)` here?

AndreyTretyak: good point, updated, thanks

artempushkin (approved):

vlad-ion (approved):

PR #408, opened 2021-10-04
Add secure cluster support for guest executables

AndreyTretyak:
Allows to provide cert common name to use for accessing SF Cluster API endpoint


src/Extensions/Hosting/Certificates/InitializationCertificateReader.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Omex.Extensions.Hosting.Certificates
+{
+	/// <summary>
+	/// CertificateReader to use until DI container build.
+	/// </summary>
+	/// <remarks>
+	/// Please prefer using instance resolved from DI container.
+	/// </remarks>
+	public class InitializationCertificateReader
```
aszczepanski, line 14: Could it be a static class?

src/Extensions/ServiceFabricGuest.Abstractions/internal/ServiceFabricClientWrapper.cs
```diff
@@ -2,32 +2,68 @@
...
+			ServiceFabricClientBuilder builder = new ServiceFabricClientBuilder().UseEndpoints(clusterEndpoint);
+
+			if (clusterEndpoint.Scheme == Uri.UriSchemeHttps)
+			{
+				if (string.IsNullOrWhiteSpace(m_options.ClusterCertCommonName))
+				{
+					throw new InvalidOperationException($"{nameof(m_options.ClusterCertCommonName)} could not be empty for {Uri.UriSchemeHttps} cluster.");
+				}
+
+				X509Certificate2? clusterCert = m_certificateReader.GetCertificateByCommonName(m_options.ClusterCertCommonName);
```
aszczepanski, line 53: If there's multiple installed certificates matching the Common Name, which one is used?

AndreyTretyak: Logic is here https://github.com/microsoft/Omex/blob/3ec60e96b5068837302ff51579e2c1cb7ecdadbc/src/Extensions/Hosting/Certificates/CertificateReader.cs#L97

src/Extensions/Hosting/Certificates/InitializationCertificateReader.cs
AndreyTretyak: oh yeah it should, missed it, thanks, good catch

AndreyTretyak: fixed

vlad-ion (approved):

src/Extensions/ServiceFabricGuest.Abstractions/internal/ServiceFabricClientWrapper.cs
artempushkin: I happened to be looking at that as well. I wonder why we even bother to return expired certificates?

AndreyTretyak: Not sure, initially I was trying to duplicate old behavior, but thinking about it I don't see a good reason to have an expired cert, unless we are getting better error message when it's expired compared to when it's just missing.

@aszczepanski maybe you know why are we returning expired certs?

PR #428, opened 2021-11-08
Replace consts with static readonly

AndreyTretyak:
Using `const` results in a new allocation each time it's used. The most common exception would be health check markers we have that used each time we mark activity as healthcheck or check it.

This PR replaces `const` with `static readonly` to avoid additional allocations.

### DIFF
--- a/src/Extensions/Activities/ServiceCollectionExtensions.cs
+++ b/src/Extensions/Activities/ServiceCollectionExtensions.cs
@@ -1,11 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Text;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.ObjectPool;
 using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
 using Microsoft.Omex.Extensions.Abstractions.ExecutionContext;
@@ -18,8 +15,8 @@ namespace Microsoft.Extensions.DependencyInjection
 	/// </summary>
 	public static class ServiceCollectionExtensions
 	{
-		private const string ActivitySourceName = "OmexActivitySource";
-		private const string ActivitySourceVersion = "1.0.0.0";
+		private static readonly string s_activitySourceName = "OmexActivitySource";
+		private static readonly string s_activitySourceVersion = "1.0.0.0";
 
 		/// <summary>
 		/// Add ActivitySource to ServiceCollection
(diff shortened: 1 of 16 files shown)


syphe (approved):
Another reason this is a good idea, is that const is determined at compile time when being referenced elsewhere, where as 'static readonly' will cause the value to be referenced at runtime instead.

artempushkin:
Let's look closer at those extra allocations, I don't see where they would be coming from.

src/Extensions/Activities/ServiceCollectionExtensions.cs
```diff
@@ -18,8 +15,8 @@ namespace Microsoft.Extensions.DependencyInjection
 	/// </summary>
 	public static class ServiceCollectionExtensions
 	{
-		private const string ActivitySourceName = "OmexActivitySource";
-		private const string ActivitySourceVersion = "1.0.0.0";
+		private static readonly string s_activitySourceName = "OmexActivitySource";
```
K-Cully, line 18: this class is already static

src/Extensions/Logging/Internal/Replayable/ReplayableActivityExtensions.cs
```diff
@@ -11,10 +11,10 @@ namespace Microsoft.Omex.Extensions.Logging.Internal.Replayable
 {
 	internal static class ReplayebleActivityExtensions
 	{
-		private const string ReplayableLogKey = "ReplayableLogKey";
+		private static readonly string s_replayableLogKey = "ReplayableLogKey";
```
K-Cully, line 14: class is already static

src/Extensions/Services.Remoting/Internal/Diagnostics.cs
```diff
@@ -14,15 +14,15 @@ internal static class Diagnostics
 		/// <remarks>
 		/// Changing this name will mean that consumers might miss telemetry events
 		/// </remarks>
-		internal const string DiagnosticListenerName = "Microsoft.Omex.Extensions.Services.Remoting";
+		internal static readonly string DiagnosticListenerName = "Microsoft.Omex.Extensions.Services.Remoting";
```
K-Cully, line 17: class is already static

src/Extensions/Services.Remoting/OmexRemotingHeadersExtensions.cs
```diff
@@ -29,7 +29,7 @@ public static class OmexRemotingHeadersExtensions
 		/// W3C standard for trace context parent 'traceparent', so we've added omex to avoid conflicts if support would be added in future
 		/// link: https://www.w3.org/TR/trace-context/
 		/// </remarks>
-		private const string TraceParentHeaderName = "omex-traceparent";
+		private static readonly string s_traceParentHeaderName = "omex-traceparent";
```
K-Cully, line 32: class is static

K-Cully:
> Let's look closer at those extra allocations, I don't see where they would be coming from.

I take it that the issue is that as transient instances get created and destroyed as part of their DI lifecycle, the const values are being repeatedly allocated.
However, this change appears to be over scoped, since that would not affect static classes.

artempushkin:
> > Let's look closer at those extra allocations, I don't see where they would be coming from.
> 
> I take it that the issue is that as transient instances get created and destroyed as part of their DI lifecycle, the const values are being repeatedly allocated. However, this change appears to be over scoped, since that would not affect static classes.

They shouldn't be, CLI ensures that each string constant exists as a singular object (though you can trick it if you try hard). I was talking to Andrey separately, profiler detects multiple allocations of some strings but I'm sure this PR won't fix it.

artempushkin (changes requested):
I'll block this to prevent accidental merging.

PR #437, opened 2022-01-18
Add ability in to set failure status in http healthcheck

AndreyTretyak:
Currently, we use the only default value (`Unhealthy`) but in some cases, we need no blocking health check that returns Degraded in case of failure.

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/HealthChecksBuilderExtensions.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/HealthChecksBuilderExtensions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 		/// <param name="scheme">uri scheme, defaults to http</param>
 		/// <param name="headers">headers to attach to the request</param>
 		/// <param name="expectedStatus">response status code that considered healthy, default to 200(OK)</param>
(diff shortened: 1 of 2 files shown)


src/Extensions/Diagnostics.HealthChecks/HealthChecksBuilderExtensions.cs
```diff
@@ -40,6 +41,7 @@ public static class HealthChecksBuilderExtensions
 			string? scheme = null,
 			IReadOnlyDictionary<string, IEnumerable<string>>? headers = null,
 			HttpStatusCode? expectedStatus = null,
+			HealthStatus failureStatus = HealthStatus.Unhealthy,
```
artempushkin, line 44: failureStatus should be after name to align with other registration APIs, (https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.healthchecksbuilderaddcheckextensions.addtypeactivatedcheck?view=dotnet-plat-ext-6.0)

AndreyTretyak: Do you mean making type nullable? I've changed it to be similar to `AddTypeActivatedCheck`, please have a look once more.

artempushkin (approved):

neilr81 (approved):

PR #438, opened 2022-01-20
Use healthcheck name for wrapping activity

AndreyTretyak:
We are currently using the constant name for the activity that wraps the health check, with this change it would be named the same as healthcheck.

### DIFF
--- a/src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
+++ b/src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
@@ -43,7 +43,8 @@ namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 		/// <inheritdoc />
 		public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken token = default)
 		{
-			using Activity? activity = m_activitySource.StartActivity(context.Registration.Name)
+			string activityName = string.IsNullOrWhiteSpace(context.Registration.Name) ? "HealthCheckActivity" : context.Registration.Name; 
+			using Activity? activity = m_activitySource.StartActivity(activityName)
 				?.MarkAsSystemError()
 				.MarkAsHealthCheck();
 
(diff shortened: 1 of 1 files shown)


src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
```diff
@@ -43,7 +43,7 @@ protected AbstractHealthCheck(TParameters parameters, ILogger logger, ActivitySo
 		/// <inheritdoc />
 		public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken token = default)
 		{
-			using Activity? activity = m_activitySource.StartActivity("HealthCheckActivity")
+			using Activity? activity = m_activitySource.StartActivity(context.Registration.Name)
```
K-Cully, line 46: Suggest adding an IsNullOrWhiteSpace check on this value, falling back to "HealthCheckActivity" if the name is not set correctly.

K-Cully, line 46: Suggest adding an IsNullOrWhiteSpace check on this value, falling back to "HealthCheckActivity" if the name is not set correctly.

AndreyTretyak: Good point added the check.

src/Extensions/Diagnostics.HealthChecks/AbstractHealthCheck.cs
```diff
@@ -43,7 +43,8 @@ protected AbstractHealthCheck(TParameters parameters, ILogger logger, ActivitySo
 		/// <inheritdoc />
 		public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken token = default)
 		{
-			using Activity? activity = m_activitySource.StartActivity("HealthCheckActivity")
+			string activityName = string.IsNullOrWhiteSpace(context.Registration.Name) ? "HealthCheckActivity" : context.Registration.Name;
```
artempushkin, line 46: Can we differentiate between activities with identical names from different services? It's very likely that there will be collisions on registration name.

AndreyTretyak: Activity reports service name as one of the parameters, so it should be fine. https://github.com/microsoft/Omex/blob/4bc5d95363797390b29fb8e024b6952715f4cc7c/src/Extensions/Activities/Internal/EventSource/ActivityEventSender.cs#L84

artempushkin (approved):

K-Cully (approved):

PR #439, opened 2022-01-24
Delete deprecated packages

AndreyTretyak:
Delete `Omex.System` package and all they depend on it, since we are not recommending for usage anymore and that break builds on using the latest version.


artempushkin:
Have you verified that these packages are being published from OmexSharedOsi?

artempushkin:
> Have you verified that these packages are being published from OmexSharedOsi?

Looks like they are.

artempushkin (approved):

neilr81 (approved):

PR #445, opened 2022-02-23
Dimeehan/log scrubber: accept MatchEvaluator for custom regex replacement logic

dimeehan:
Updating this logic to accept matchevaluator for log scrubbing in Omex. Currently, it scrubs the entire key and value with REDACTED which makes it difficult to read and debug. 

Scrub q1 and q2 query params:
Input: https://api.host.com/appinstall?q1=abc&p1=444&p2=555&q2=def
Current post-scrub: https://api.host.com/appinstall?REDACTED&p1=444&p2=555&REDACTED
Wanted post-scrub: https://api.host.com/appinstall?q1=REDACTED&p1=444&p2=555&q2=REDACTED

### DIFF
--- a/src/Logging/Scrubbing/RegexLogScrubbingRule.cs
+++ b/src/Logging/Scrubbing/RegexLogScrubbingRule.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Omex.Extensions.Logging.Scrubbing
 		/// Initializes a new instance of the <see cref="RegexLogScrubbingRule"/> class.
 		/// </summary>
 		/// <param name="regexToReplace">The regular expression specifying the strings to replace.</param>
-		/// <param name="matchEvaluator">The value with which to replace the matching text.</param>
+		/// <param name="matchEvaluator">Custom logic for regex replace.</param>
 		public RegexLogScrubbingRule(string regexToReplace, MatchEvaluator? matchEvaluator)
 		{
 			m_regexToReplace = new Regex(regexToReplace, RegexOptions.Compiled | RegexOptions.CultureInvariant | RegexOptions.IgnoreCase);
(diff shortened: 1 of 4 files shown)


DierckxLuna (dismissed):

src/Logging/Scrubbing/RegexLogScrubbingRule.cs
```diff
@@ -11,25 +11,41 @@ namespace Microsoft.Omex.Extensions.Logging.Scrubbing
...
+		private readonly MatchEvaluator? m_matchEvaluator;
+		private readonly string? m_replacementValue;
 
 		/// <summary>
 		/// Initializes a new instance of the <see cref="RegexLogScrubbingRule"/> class.
 		/// </summary>
 		/// <param name="regexToReplace">The regular expression specifying the strings to replace.</param>
 		/// <param name="replacementValue">The value with which to replace the matching text.</param>
-		public RegexLogScrubbingRule(string regexToReplace, string replacementValue)
+		public RegexLogScrubbingRule(string regexToReplace, string? replacementValue)
```
AndreyTretyak, line 22: I would argue that if user is colling this constructor `replacementValue` input parameter should not be `null`.  While you still need to keep fields as nullable. Same goes for `matchEvaluator` in another constructor.

AndreyTretyak, line 22: Could we implement this constructor by calling a more generic one, like this?
```Csharp
public RegexLogScrubbingRule(string regexToReplace, string replacementValue)
=> this(regexToReplace, new MatchEvaluator(m => replacementValue)) { }
```
This would allow us to avoid duplicated logic and nullable fields? 
Or do you worried that it would be less performant?

dimeehan: Yes it was added due to non-nullable leaving constructor. I updated the code there to remove this part.

dimeehan: As discussed offline, we'll be sticking with current approach as having match evaluator in both scenarios would cause performance degradation in both places.

dimeehan:
**Unit test profiler from VS2022**
image

src/Logging/Scrubbing/RegexLogScrubbingRule.cs
```diff
@@ -24,12 +25,27 @@ public RegexLogScrubbingRule(string regexToReplace, string replacementValue)
 			m_replacementValue = replacementValue;
 		}
 
+		/// <summary>
+		/// Initializes a new instance of the <see cref="RegexLogScrubbingRule"/> class.
+		/// </summary>
+		/// <param name="regexToReplace">The regular expression specifying the strings to replace.</param>
+		/// <param name="matchEvaluator">Custom logic for regex replace.</param>
```
K-Cully, line 32: Suggest adding a remarks tag detailing that this comes with an increased perf overhead.

DierckxLuna (approved):

K-Cully (approved):

aszczepanski (approved):

PR #447, opened 2022-02-28
Workaround for CodeQL build failures

dimeehan:
Sample build failure: https://github.com/microsoft/Omex/runs/5357713253?check_suite_focus=true 

Error: Tracing builds with **CodeQL is currently not supported on Windows 11 and Windows Server 2022.** 
Please modify your Actions workflow to use an earlier version of Windows for this job, for example by **setting `runs-on: windows-2019`.**
Error: Tracing builds with CodeQL is currently not supported on Windows 11 and Windows Server 2022. Please modify your Actions workflow to use an earlier version of Windows for this job, for example by setting `runs-on: windows-2019`.
    at checkNotWindows11 (D:\a\_actions\github\codeql-action\v1\lib\util.js:533:15)
    at run (D:\a\_actions\github\codeql-action\v1\lib\init-action.js:97:42)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async runWrapper (D:\a\_actions\github\codeql-action\v1\lib\init-action.js:155:9)

https://github.com/actions/virtual-environments/issues/4856 Migration here mentions the following:
1. File an issue in this repository
2. Switch back to windows-2019 by specifying the windows-2019 label in your workflow. We will continue to support Windows Server 2019 image.

PR #448, opened 2022-02-28
Use Windows 2019 for CodeQL

neilr81:
Windows 2022 is currently not supported

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,7 +17,8 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: windows-latest
+    # Move to Windows 2022 when CodeQL is supported there
+    runs-on: windows-2019
 
     strategy:
       fail-fast: false


.github/workflows/codeql-analysis.yml
```diff
@@ -17,7 +17,8 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: windows-latest
+    # Move to Windows 2022 when CodeQL is supported there
+    runs-on: windows-2019
```
dimeehan, line 21: I was also going to make this change earlier https://github.com/microsoft/Omex/pull/447 as I thought it was blocking changes :) if we make this change, would the plan be to move back to windows-latest at a later stage?

artempushkin: According to this (https://github.com/actions/virtual-environments/issues/4856) server 2022 is being rolled out. As a naturally lazy engineer, I'd just wait, if it's not blocking.

K-Cully: Yes, server 2022 is being rolled out as the windows-latest target, which is why this change is needed.
CodeQL doesn't support server 2022 - this is explicitly called out in that issue.
We should change it to 2019 and log an issue to track changing it back to latest when https://github.com/github/codeql-action/issues/850 is fixed.

K-Cully: Note it is currently blocking package updates - https://github.com/microsoft/Omex/pull/444

K-Cully (approved):

dimeehan (approved):

.github/workflows/codeql-analysis.yml
artempushkin: Ah, I've misread the error message completely, thanks for noticing this.

PR #455, opened 2022-04-12
Updating to use Packages.Props

neilr81:


Directory.Build.targets
```diff
@@ -0,0 +1,3 @@
+<Project>
+  <Sdk Name="Microsoft.Build.CentralPackageVersions" Version="2.0.1" />
```
AndreyTretyak, line 2: The version of SDK should be provided only in `global.json` file. I think we can remove it from here.

AndreyTretyak (approved):

marc-hayek (approved):

artempushkin (approved):

PR #475, opened 2022-06-22
Removing the console logger when running on Service Fabric

neilr81:
Check for running on SF used from example here:
https://stackoverflow.com/questions/39781817/how-to-see-if-running-under-service-fabric

This is a suspected issue causing performance issues on apps when running high request counts.

### DIFF
--- a/src/Logging/InitializationLogger.cs
+++ b/src/Logging/InitializationLogger.cs
@@ -26,9 +26,9 @@ namespace Microsoft.Omex.Extensions.Logging
 		{
 			// Check if we are running on Service Fabric, if we are not on Service Fabric, add console logging
 			string sfAppName = Environment.GetEnvironmentVariable("Fabric_ApplicationName");
-			bool isSf = sfAppName != null;
+			bool isServiceFabric = sfAppName != null;
 			
-			if (!isSf)
+			if (!isServiceFabric)
 			{
 				builder.AddConsole();
 			}
--- a/src/Logging/InitializationLogger.cs
+++ b/src/Logging/InitializationLogger.cs
@@ -24,7 +24,15 @@ namespace Microsoft.Omex.Extensions.Logging
 
 		private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder builder)
 		{
-			builder.AddConsole();
+			// Check if we are running on Service Fabric, if we are not on Service Fabric, add console logging
+			string sfAppName = Environment.GetEnvironmentVariable("Fabric_ApplicationName");
+			bool isSf = sfAppName != null;
+			
+			if (!isSf)
+			{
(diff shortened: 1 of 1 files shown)


src/Logging/InitializationLogger.cs
```diff
@@ -24,7 +24,15 @@ public static class InitializationLogger
 
 		private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder builder)
 		{
-			builder.AddConsole();
+			// Check if we are running on Service Fabric, if we are not on Service Fabric, add console logging
+			string sfAppName = Environment.GetEnvironmentVariable("Fabric_ApplicationName");
```
artempushkin, line 28: I think nullability is enabled in these libs, so you have to use explicitly nullable types.

artempushkin:
If I understand correctly, this logger is only used during the app initialization, it won't improve runtime performance. Thinking more about it, there is no way to run our services outside of SF, and we typically run them in windowless mode, so having a console logger makes no sense.

K-Cully:
> If I understand correctly, this logger is only used during the app initialization, it won't improve runtime performance. Thinking more about it, there is no way to run our services outside of SF, and we typically run them in windowless mode, so having a console logger makes no sense.

I believe Andre added this for 2 reasons:
1. When we eventually containerised services it would ease debugging services in isolation.
2. It aligns closely with the default ASP.NET Core behaviour in CreateDefaultBuilder -  https://cs.github.com/dotnet/aspnetcore/blob/442a380854c25b85252a743c3ba2e21f08fa6879/src/DefaultBuilder/src/WebHost.cs#L201

However, if we are keeping it, it should never be ran anywhere outside of development so it would make more sense to only register it if the environment is development, rather than if it is executing outside of SF.

dimeehan:
> > If I understand correctly, this logger is only used during the app initialization, it won't improve runtime performance. Thinking more about it, there is no way to run our services outside of SF, and we typically run them in windowless mode, so having a console logger makes no sense.
> 
> I believe Andre added this for 2 reasons:
> 
> 1. When we eventually containerised services it would ease debugging services in isolation.
> 2. It aligns closely with the default ASP.NET Core behaviour in CreateDefaultBuilder -  https://cs.github.com/dotnet/aspnetcore/blob/442a380854c25b85252a743c3ba2e21f08fa6879/src/DefaultBuilder/src/WebHost.cs#L201
> 
> However, if we are keeping it, it should never be ran anywhere outside of development so it would make more sense to only register it if the environment is development, rather than if it is executing outside of SF.

Created a separate PR to disable outside of dev environment: https://github.com/microsoft/Omex/pull/476

K-Cully:
Superseded by https://github.com/microsoft/Omex/pull/476

PR #476, opened 2022-06-28
Improving OmexExternal performance - disabling console logger outside of development environment

dimeehan:
**PR is to remove the console logger outside of a development environment in the InitializationLogger.**

After our recent investigations around the requests being queued, 502s, latencies, etc. we believe that the console logger is still being used across the board. The logger only has a small buffer to handle requests and looks to be blocking calls to threads which we believe leads to delays, queuing of requests, AG 5xx responses.

Although the console logger is disabled in our appsettings.json, we also need to clear the logging providers during startup and only add the console logger in dev environment. TMS change here: https://office.visualstudio.com/OC/_git/TargetedMessagingService/pullrequest/1387670 
This has been tested in production with higher traffic and we've seen a big overall performance  improvement. However, the real test will be when the en-US traffic comes back to External cluster this week (Wednesday 29th). If successful, the change will be added to our Extension libraries for the rest of OMEX to use in their services. We believe everyone could be getting a performance hit here, but may not be as noticeable as TMS due to lower amounts of traffic. 

This PR is updating the console logger in the initialization logger. We don't think it's causing any contention, however it's not needed outside of the dev environment so it's better to disable.


src/Logging/InitializationLogger.cs
```diff
@@ -24,7 +24,11 @@ public static class InitializationLogger
 
 		private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder builder)
 		{
-			builder.AddConsole();
+			if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == "Development")
```
dakanev-msft, line 27: >Development [](http://example.com/codeflow?start=72&length=11)

do we have data if anyone is actually using that in Dev environment?

dakanev-msft: I would be in favor in removing it in all environments and re-adding it if someone screams

dimeehan: No I don't have that data, do you know how we would get that? Wouldn't it be safer to just keep it in dev and it shouldn't cause any problems/issues like it does in other environments?

dakanev-msft: I think the more unused functionality we have, the more difficult it is to maintain it. Also - what if people see some logs in Dev and expect to see them in PROD? You have to explain about the console logger not being enabled.

I would just add OMEX to this PR.

neilr81: The console logger would be where people see logs during development? I think it should be OK to keep it in such circumstances, I don't think many people would be logging onto the fixed environments and expect to see the console for their 1 app running.

neilr81 (approved):

src/Logging/InitializationLogger.cs
dakanev-msft: I am checking the Diagnostic events during development, it would be good to check locally what happens when this logger is disabled - it might be the case that the console logger just adds some logging before initializing the other loggers.

dakanev-msft (approved):

PR #478, opened 2022-07-20
Updating transitive version of NewtonSoft being used

neilr81:
Minimum version 10 is being picked up instead of 13, adding specific references to force the higher version. This is despite     key="dependencyVersion" value="Highest" being specified in the Nuget.config.
Interested if there is a better way to do this.


marc-hayek (approved):

K-Cully:
Does every project here have Newtonsoft as a transitive dependency?
What is the source dependency that is pulling it in?

dimeehan (approved):

K-Cully (approved):

Issue #453, opened 2022-03-16
Process components are not managed correctly


K-Cully:
In some logging classes Process.GetCurrentProcess is used incorrectly.
See https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1837 regarding the issues.

For initialization logger, the code should be updated to Environment.ProcessId (https://docs.microsoft.com/en-us/dotnet/api/system.environment.processid?view=net-6.0).
For OmexLogEventSender, the process should be created in a using scope to ensure it is disposed correctly.

See https://github.com/microsoft/Omex/search?q=GetCurrentProcess

artempushkin:
Good catch.

PR #486, opened 2022-08-31
Replace inefficient uses of Process

K-Cully:
In some logging classes Process.GetCurrentProcess is used incorrectly.
This change fixes these use cases.

- Replace Process.GetCurrentProcess.Id in initialization logger, with Environment.ProcessId to prevent unnecessary Process initialization. (https://docs.microsoft.com/en-us/dotnet/api/system.environment.processid?view=net-6.0).
- Add using scope around Process in OmexLogEventSender to ensure it is disposed correctly.

Fixes https://github.com/microsoft/Omex/issues/453

### DIFF
--- a/src/Logging/InitializationLogger.cs
+++ b/src/Logging/InitializationLogger.cs
@@ -41,11 +41,11 @@ namespace Microsoft.Omex.Extensions.Logging
 		public static void LogInitializationSucceed(string serviceNameForLogging, string message = "")
 		{
 			string logMessage = $"Initialization successful for {serviceNameForLogging}, {message}";
-#if Net_Standard
+#if NET5_0_OR_GREATER
+			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Environment.ProcessId, serviceNameForLogging, logMessage);
+#else
 			using Process process = Process.GetCurrentProcess();
 			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(process.Id, serviceNameForLogging, logMessage);
-#else
-			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Environment.ProcessId, serviceNameForLogging, logMessage);
 #endif
 			Instance.LogInformation(Tag.Create(), logMessage);
 		}
--- a/src/Logging/InitializationLogger.cs
+++ b/src/Logging/InitializationLogger.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Omex.Extensions.Logging
 		public static void LogInitializationSucceed(string serviceNameForLogging, string message = "")
 		{
 			string logMessage = $"Initialization successful for {serviceNameForLogging}, {message}";
-			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging, logMessage);
+			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Environment.ProcessId, serviceNameForLogging, logMessage);
 			Instance.LogInformation(Tag.Create(), logMessage);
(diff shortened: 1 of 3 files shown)


src/Logging/InitializationLogger.cs
```diff
@@ -41,7 +41,7 @@ private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder bui
 		public static void LogInitializationSucceed(string serviceNameForLogging, string message = "")
 		{
 			string logMessage = $"Initialization successful for {serviceNameForLogging}, {message}";
-			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Process.GetCurrentProcess().Id, serviceNameForLogging, logMessage);
+			ServiceInitializationEventSource.Instance.LogHostBuildSucceeded(Environment.ProcessId, serviceNameForLogging, logMessage);
```
K-Cully, line 44: Environment.ProcessId doesn't exist in NetStandard 2.0

dimeehan, line 44: Is Environment.ProcessId correct? I see some failures below saying there is no definition for 'ProcessId' ?

K-Cully: @artempushkin - is there any reason the Microsoft.Omex.Extensions.Logging should support NetStandard 2.0?
I'd assume at this stage we wouldn't want to support FullFramework in downstream consumers, right?

src/Logging/Microsoft.Omex.Extensions.Logging.csproj
```diff
@@ -2,6 +2,9 @@
   <PropertyGroup>
     <TargetFrameworks>$(LibraryTargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
+  <PropertyGroup Condition="$(TargetFramework.StartsWith(`netstandard`))">
```
AndreyTretyak, line 5: MSBuild defines the default constant `NETSTANDARD` no need to create another one https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework

src/Logging/InitializationLogger.cs
artempushkin: We do target netframework in a few services, mostly in OmexAddInCertification. I suppose there can be partner dependencies forcing full .net.

K-Cully: I went ahead with using conditional compilation, but I feel like we should move away from net standard in future.

K-Cully: Yes, as per my comment above Environment.ProcessId doesn't exist in NetStandard 2.0.
It was only added in Net5. (https://docs.microsoft.com/en-us/dotnet/api/system.environment.processid?view=net-6.0)
The main benefits are that it:
- Caches the ID for any future calls.
- Prevents creation of a new process, just to get the ID.
- Avoids memory leaks due to developers forgetting to dispose of the Process instance that is created by GetCurrentProcess()

To get around the Net5+ requirement, I've specified conditional inclusion of this logic, and wrapped the NetStandard code in a using, to ensure the process gets disposed.

src/Logging/Microsoft.Omex.Extensions.Logging.csproj
K-Cully: Thanks Andrey I was looking for something like that but didn't come across that doc. :)

src/Logging/InitializationLogger.cs
K-Cully: Changed the variables used based on the doc Andrey linked.
I think the use of NET5_0_OR_GREATER adds more context.

neilr81 (approved):

PR #504, opened 2022-10-04
Moved the FabricClient wrapper in the open source project.

MaxDac:
The interface `IFabricClientWrapper` existed previously in the shared private package. It was stripped of CCC dependencies and repurposed as an interface and relative implementation that would allow testing by mock instead of a direct use of the class `FabricClient`.

An extension method was added to facilitate the registration of the IFabricClientWrapper instance by default.

### DIFF
--- a/src/Hosting.Services/IFabricClientWrapper.cs
+++ b/src/Hosting.Services/IFabricClientWrapper.cs
@@ -1,0 +1,143 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+using System.Fabric;
+using System.Fabric.Health;
+using System.Fabric.Query;
+using System.Threading.Tasks;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services
+{
+	/// <summary>
+	/// Fabric client wrapper interface
+	/// </summary>
+	public interface IFabricClientWrapper
+	{
+		/// <summary>
+		/// Gets service replicas
+		/// </summary>
+		/// <param name="partitionId">Partition id</param>
+		/// <param name="continuationToken">Continuation token</param>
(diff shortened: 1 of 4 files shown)


src/Hosting.Services/IFabricClientWrapper.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Fabric.Health;
+using System.Fabric.Query;
+using System.Threading.Tasks;
+
+namespace Microsoft.Omex.Extensions.Hosting.Services
+{
+	/// <summary>
+	/// Fabric client wrapper interface
+	/// </summary>
+	public interface IFabricClientWrapper
```
AndreyTretyak, line 15: In my opinion, having this interface as is goes against Omex.Extensions attempt to encapsulate SF dependencies and simplify future migration to AKS or other platforms. 
If you absolutely need a client wrapper it should not just repeat the methods of SF client, but provide the functionality of calling other services without exposing SF guts to the consumer.
If this one only needed to get SF internal topology and health for services like SOS and ClusterDiagnostcs I would argue that it should not be here, since it would have to be deprecated along with them or updated in the way works not only for SF.

MaxDac: The interface and its implementation are a simple wrapper around FabricClient only for testing purpose only, it would solve the complications of mocking the FabricClient in Omex internal projects: it was already implemented in the Shared internal project, but we were looking for a way of refactoring it to make it available for every project.
I agree with you that if this project effort is to abstract the platform in which the microservice is executed, though, this is not the right place to put it. @K-Cully what do you think?

K-Cully: Agree. It shouldn't go in extensions.

MaxDac:
The class in this PR will be moved to the Shared internal package.

PR #509, opened 2022-10-12
StringBuilder Changes for message string

walidaslam7:
Converting message string to String builder before Loop.

### DIFF
--- a/src/Logging/Internal/OmexLogger.cs
+++ b/src/Logging/Internal/OmexLogger.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Microsoft.Extensions.Logging;
 using Microsoft.Omex.Extensions.Logging.Replayable;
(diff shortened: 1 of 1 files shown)


artempushkin (changes requested):
This change breaks the logger logic

src/Logging/Internal/OmexLogger.cs
```diff
@@ -39,27 +40,27 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
 			if (exception != null)
 			{
-				message = string.Concat(message, Environment.NewLine, exception); // We need to concatenate with exception since the default formatter ignores it https://github.com/aspnet/Logging/issues/442
+				message.Append(string.Concat(message, Environment.NewLine, exception)); // We need to concatenate with exception since the default formatter ignores it https://github.com/aspnet/Logging/issues/442
 			}
 
 			foreach (ILogScrubbingRule textScrubber in m_textScrubbers)
 			{
-				message = textScrubber.Scrub(message);
+				message.Append(textScrubber.Scrub(message.ToString()));
```
AndreyTretyak, line 51: Thank you for your contribution, but I think that using `StringBuilder` wouldn't provide better performance at this moment, it would degrade it instead.
The most common use case is that `m_textScrubbers` is an empty collection, having exceptions also relatively rare occasions, compared to the log calls without exception (and single `string.Concat` should still be more performant than the creation of `StringBuilder`), so in the majority of cases we return string as is, and creating `StringBuilder` would be a waste of resources.

Besides that, the point of scrubbers is to replace existing log messages with scrabbed values. Your change breaks this behavior and just appends scrabbed messages to the original defeating the point of scrubbing.

PR #513, opened 2022-10-22
Remove unused packages

AndreyTretyak:
Remove mentions of the NuGet packages that are not referenced/used by current projects.


artempushkin:
Could you add a brief description? Can this change potentially break consumers with transitive dependencies?

AndreyTretyak:
> Could you add a brief description? Can this change potentially break consumers with transitive dependencies?

This change only cleans up the mentioning of the packages that were not used at all (excluding xunit that was referenced by UT project but not used). 
Packages where there as leftovers of old projects deleted before and does not have effect on the build process of current project.
Purely maintenance improvement change from my point of view.

neilr81 (approved):

artempushkin (approved):

PR #514, opened 2022-10-27
Added IOptions extension methods

MaxDac:
## Summary

Introduces two new extension methods for `IOptions` and `IOptionsMonitor` interfaces named `SafeGetValue`.

### Motivation

In .NET, apart from `IOptions` interface that provides a configuration value as it was during startup binding, there are two other interfaces, `IOptionsSnapshot` (that inherits from `IOption`) and `IOptionsMonitor` interfaces provides access to updated settings; the former is set as a **scoped** dependency, while the latter is a **singleton** dependency that provides a configuration value that gets updated in time.

If the option settings bind was defined during Startup to include validation of the bounded object by calling the method `ValidateDataAnnotations`, the three interfaces will throw a `OptionsValidationException` when the configuration does not correspond to the defined constraints. This means that, if the configuration changes at runtime to values that do not satisfy the defined condition, accessing the getter method `IOptionsSnapshot.Value` or `IOptionsMonitor.CurrentValue` will throw the exception in the service, causing abrupt interruption of execution.

More information can be found [here](https://learn.microsoft.com/en-us/dotnet/core/extensions/options).

### Implementation

The two extension methods simply wrap the call to the getter methods `IOptionsSnapshot.Value` or `IOptionsMonitor.CurrentValue` in a `try ... catch` clause, handling the aforementioned exception only. If the call to the getter methods succeeds, the configuration will be returned, otherwise the method will return a null value.

### Reasoning

Other extensions were considered:

- A method that would return the validation error was considered, but the signature of the method has not been deemed sufficiently integrable with the coding style of the project and in .NET in general (more on that in the notes below), and the error details would be easily loggable by passing the `ILogger` instance into the method, or by handling the exception manually in the code.
- Other IOptions interfaces were considered to provide a more elaborate way of getting the settings, for instance returning the initial configuration set in the service at Startup, but this would have incremented the amount of opinionated code about how to handle edge cases that would have been better implemented directly in the client code to respond to the particular needs.

### Examples

Currently, to handle the validation error of a change in configuration at runtime, the `OptionsValidationException` would have to be handled. 
Be `Configuration` the class into which the configuration is bound, and `Executor` a class that receives an instance of `IOptions<Configuration>` in the constructor, in the hypothetic `ExecuteAction` method that needs an option from `Confiiguration` it can be written:

```csharp
public void ExecuteAction()
{
    try
    {
        string optionValue = m_optionsMonitor.CurrentValue.OptionValue;
        /* Execution of the action using optionValue */
    }
    catch (OptionsValidationException ex)
    {
        m_logger.LogError(/* logging exception */);
        throw;
    }
}
```

With these new extension methods, the handling of the exception will be automatic:

```csharp
public void ExecuteAction()
{
    // for IOptions and IOptionsSnapshot interface
    string? optionValue = m_options.SafeGetValue();

    // for IOptionsMonitor interface
    string? optionValue = m_optionsMonitor.SafeGetCurrentValue();

    // with Logger, with m_logger being an instance of the ILogger<Executor>
    string? optionValue = m_optionsMonitor.SafeGetCurrentValue(m_logger)?.OptionValue;

    /* Execution of the action using optionValue, although in this case the null check would be necessary */
    if (optionValue != null)
    {
        /* Execution Method */
    }

    // Imagining that the Configuration class is bundled with a series of constant values representing
    // default values, the null check can be avoided:
    string? optionValue = m_optionsMonitor.SafeGetCurrentValue(m_logger)?.OptionValue ?? ConfigurationDefaults.OptionValue;
}
```

### Notes

About the other signatures for extension methods considered, there were these two:

- `bool TryGetValue<T>(IOptions<T> options, out T? settings)`: this signature would not have added more information than the current methods, that forces the client code to check for nullability anyway.
- `(T?, IEnumerable<string>?) GetValue<T>(IOptions<T> options)`: this signature would have added error information through a "monad-like" return type, but was not deemed easily integrable within the coding style context.

### Final Considerations

These extension methods can be used to get configuration entries without catching an exception, but whenever it makes more sense to manually catch the exception or let the service fail as soon as it recognises that the configuration is incorrect, the user should rely on that mechanism.

The second purpose of these extension methods is to highlight to the user the fact that the getter methods of the `IOptions` interfaces throw a particular exception of type `OptionsValidationException` that should be handled if called directly.


artempushkin:
Thanks for the examples, however they don't demonstrate further handling of configuration absence. Try writing a complete snippet that tries to get a value, fails and takes recovery action. I suspect the difference between the two approaches will be less noticeable.

MaxDac:
> Can you provide usage examples with and without these extensions? If the application is misconfigured, silently falling back to safe defaults is not necessarily the best approach, it might be in fact better to explode and rely on monitoring to catch and mitigate that condition.

I agree, these extension methods would introduce a way of handling errors as null value automatically, and by giving an option to do so document in a specific way what could happen if the option interface getter method is called without proper handling, they are not meant to substitute an exception handling done in place

I added some documentation over the method, we can tweak it to suggest which handling method would be best in the client code.

MaxDac:
> Thanks for the examples, however they don't demonstrate further handling of configuration absence. Try writing a complete snippet that tries to get a value, fails and takes recovery action. I suspect the difference between the two approaches will be less noticeable.

Added more examples, and you're right: the handling of the nullable could lead to the exact amount of code, and worse, it can obscure the error rather than let it crash the application, and by thus notify us to handle the case.

But for configuration that has default values, like the example in the snippet of code, the extension method could be a more concise solution.

In my opinion, these methods could provide a better documentation and warning about how to get configuration natively. What I mean is that, for example, if one investigates which accessor grants access to the configuration and they notice that there is a method that starts with `Safe`, they will tend to investigate more on the matter, discovering that the accessor may throw an exception in case of configuration update.

src/Configuration/IOptionsExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+		/// <para>
+		/// If the options value does not pass the validation set while configuring the configuration at startup,
+		/// it will return null instead of throwing <seealso cref="OptionsValidationException"/>.
+		/// </para>
+		/// <para><em><strong>Note</strong> - In certain context, handling directly the exception, or let the service fail if the configuration
+		/// is wrong, would be a better solution.</em></para>
+		/// <para><em>Example:</em></para>
+		/// <para>
+		/// <code>
+		///	try
```
Adnovac, line 26: NIT: Isn't it wrongly formatted?

tests/Configuration.UnitTests/Support/ConfigurationExample.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Omex.Extensions.Configuration.UnitTests.Support
+{
+	public class ConfigurationExample
+	{
+		[Required(AllowEmptyStrings = false)]
+		public string? ValidatedConfiguration { get; set; }
```
Adnovac, line 12: Why is it nullable when its Required?

MaxDac: Unfortuntely, the data annotations do not have any influence on the compiler warning (in our case, errors) triggered when it recognises possible nullable reference. Hence, the alternatives are two:
- Declaring the property as nullable.
- Declaring the property as not nullable and equalise it to a predefined value.

src/Configuration/IOptionsExtensions.cs
MaxDac: Yes, there was a tab instead of a space, thanks for spotting that

neilr81 (dismissed):

AndreyTretyak:
I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies).
It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.

Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).

The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.

Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.  

In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behavior.

tests/Configuration.UnitTests/Support/ConfigurationExample.cs
Adnovac: I would go with the default value to make it more visible that it shouldn't be null but feel free to ignore that

K-Cully:
> I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> 
> Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> 
> The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> 
> Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> 
> In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.

At deployment time, I agree that a service should crash/fail to start due to incorrect configuration. However, when it comes to runtime configuration of a live service, forcing a service to break is a terrible idea as this leads to customer impact.
The choice is between:
- Having only deployment time config and breaking on config failures or
- Allowing for human errors in run-time configuration (emergency overrides) and failing gracefully.
In the latter model, the fallback behaviour should be reported in metrics so that alerts are fired to the service owners to address.

MaxDac:
> I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> 
> Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> 
> The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> 
> Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> 
> In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.

I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.

I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.

Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.

src/Configuration/IOptionsExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+		public static TOption? SafeGetCurrentValue<TOption>(this IOptionsMonitor<TOption> options, ILogger? logger = default)
+			where TOption : class
+		{
+			try
+			{
+				return options.CurrentValue;
+			}
+			catch (OptionsValidationException ex)
+			{
+				logger?.LogError(Tag.ReserveTag(0), ex, "The configuration validation threw an exception: '{ExceptionMessage}'.", ex.Message);
```
K-Cully, line 54: Tag.ReserveTag cannot be used in the OSS repo.
There is a Tag.CreateTag() function that uses the class name and line number which should be used instead.

src/Configuration/IOptionsExtensions.cs
```diff
@@ -0,0 +1,103 @@
...
+		/// </para>
+		/// </summary>
+		/// <typeparam name="TOption">The type that parses the configuration.</typeparam>
+		/// <param name="options">The <seealso cref="IOptions{TOptions}"/> instance.</param>
+		/// <param name="logger">
+		/// An <seealso cref="ILogger"/> instance that, if passed in input, will be used
+		/// to log the details of the validation exception if thrown.
+		/// </param>
+		/// <returns>The configuration value if the validation passes, <c>null</c> otherwise.</returns>
+		public static TOption? SafeGetValue<TOption>(this IOptions<TOption> options, ILogger? logger = default)
```
K-Cully, line 89: Shouldn't this extension apply to IOptionsSnapshot, rather than IOptions?

tests/Configuration.UnitTests/Support/LogCollectorLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Omex.Extensions.Configuration.UnitTests.Support
+{
+	/// <summary>
+	/// This implementation of the ILogger instace collect all the log registered within the instance
+	/// to return them for test purposes.
+	/// </summary>
+	/// <typeparam name="TInstance">The class for which the logger instance will log.</typeparam>
+	public class LogCollectorLogger<TInstance> : ILogger<TInstance>
```
K-Cully, line 16: Should this be internal?

tests/Configuration.UnitTests/Support/ConfigurationExample.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Omex.Extensions.Configuration.UnitTests.Support
+{
+	public class ConfigurationExample
```
K-Cully, line 9: should this be internal?

tests/Configuration.UnitTests/Support/ConfigurationExample.cs
MaxDac: You're right, expressing the type as a non-nullable string is semantically correct. Thanks!

AndreyTretyak:
> > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> 
> At deployment time, I agree that a service should crash/fail to start due to incorrect configuration. However, when it comes to runtime configuration of a live service, forcing a service to break is a bad idea as this leads to more profound customer impact. The choice is between:
> 
> * Having only deployment time config and breaking on config failures or
> * Allowing for human errors in run-time configuration (emergency overrides) and failing gracefully in the event of human error.
>   In the latter model, the fallback behaviour should be reported in metrics so that alerts are fired to the service owners to address.

I would argue we are providing not a way to fail gracefully, but to hide an error.
You've listed two ways of configuration updates. At deployment time and in run time.
*For deployment time configuration error* it looks like we agree that it's a bad idea to hide configuration errors.
*For runtime configuration error* I think there are better and more precise ways of guarding against those by enforcing that settings override should be applied not for all instances of the service and rollout gradually instead when first service failure would stop configuration update instead of logs that would be just ignored and everybody would be fooled that configuration applied. Especially since most of the override done by OCE or Shield and not by the person who created the service and knows what logs to look for. The point of having multiple instances of small services is to be resilient to those kinds of failures.
Even better options (but harder to implement) in creating a way for the service to provide validation rules to the settings override system so they could be verified before applying. 

My worry with this approach is that it's not selective enough, it would hide any configuration error and make everyone think that you are using new configuration. It essentially makes useless all the tools we have now to check for configuration (like SOS page or ClusterExplorer) because they won't be reliable anymore.

As soon as those methods will become available everybody would start using them, because they look safer, and it would push Omex down the rabbit hole of the defensive programming that we struggled so much to get away from.

Also, I would like to point out that having a service that runs with an incorrect configuration is usually a much bigger problem than a service that failed because failed service does nothing when a service with an incorrect configuration can do harm like writing into the wrong storage or doing incorrect request to other service showing false response to a customer that could be hard or impossible to fix afterward.

AndreyTretyak:
> > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> 
> I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.
> 
> I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.
> 
> Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.

My view on this is that if service is critical: 
1. We should not have a single instance of it.
2. We should not apply configuration changes to all instances of the service at once. (That I believe already in guidelines and even might be enforced by SOS)
Also, as I've pointed out in [reply to Keith](https://github.com/microsoft/Omex/pull/514#issuecomment-1295458505) having crucial service running in with incorrect configuration (and lying to all tools that it runs on other configuration) is even more dangerous.

Besides that, methods in this PR does not provide any distinction between critical and non-critical services, so since using them looks safer everybody would use them and all configuration becomes unreliable.

About the operator checking for logs errors during applying settings override. To me, it looks like a very optimistic scenario. It's a bad idea to hope that people follow the steps that are not enforced. Especially since most of the overrides  are done by people on-call or Shilelds that know nothing about the internals of this particular service and what methods are used to get the settings. 
If we would work on enforcing this procedure, I think, there are much better practices to enforce, for example, gradual configuration rollout with fallback in case of failure or the ability for the service to provide validation rules to the override tool.
Also, added methods allow not passing the logger or using NullLogger and in combination with constant log loss by our log aggregator could result in undetectable misconfiguration (both in runtime and deploy time). And if there is a possibility for that it will happen eventually.

Omex's previous way of getting configuration provided customers with the same choice of getting configuration value or getting it safely. As a result, all of the configuration requests were done using safe methods (and PR's were blocked because people not using safe ones with default values in case of failure). As a result, it caused multiple issues when the configuration used was different from the one everybody expected and made the code more convoluted obscuring the main logic.
This made sense before when everything was a single service and failing in one part could break totally unrelated things, but with smaller that each has their own area of responsibility we should not fall back to the same bad practices. It was intentional design behind not giving people tools to default back to the bad practices, creating discrepancies between reported and used configuration and obscure logic.

About compilation warnings, the way to do that is by creating code analyzers that ship with the package and could provide warnings, info messages, or even code fixes on incorrect code usage, but that's much more work. The benefit is that you can do that not only for your code, create analyzers about incorrect usages of standard classes. This repo might be a good sample for that https://github.com/xunit/xunit.analyzers

MaxDac:
> > > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> > 
> > 
> > I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.
> > I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.
> > Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.
> 
> My view on this is that if service is critical:
> 
> 1. We should not have a single instance of it.
> 2. We should not apply configuration changes to all instances of the service at once. (That I believe already in guidelines and even might be enforced by SOS)
> 
> Also, as I've pointed out in [reply to Keith](https://github.com/microsoft/Omex/pull/514#issuecomment-1295458505) having crucial service running with incorrect configuration (and lying to all tools that it runs on other configuration) is even more dangerous.
> 
> Besides that, methods in this PR does not provide any distinction between critical and non-critical services, so since using them looks safer everybody would use them and all configuration becomes unreliable.
> 
> About the operator checking for logs errors during applying settings override. To me, it looks like a very optimistic scenario. It's a bad idea to hope that people follow the steps that are not enforced. Especially since most of the overrides are done by people on-call or Shilelds that know nothing about the internals of this particular service and what methods are used to get the settings. If we would work on enforcing this procedure, I think, there are much better practices to enforce, for example, gradual configuration rollout with fallback in case of failure or the ability for the service to provide validation rules to the override tool. Also, added methods allow not passing the logger or using NullLogger and in combination with constant log loss by our log aggregator could result in undetectable misconfiguration (both in runtime and deploy time). And if there is a possibility for that it will happen eventually.
> 
> Omex's previous way of getting configuration provided customers with the same choice of getting configuration value or getting it safely. As a result, all of the configuration requests were done using safe methods (and PR's were blocked because people not using safe ones with default values in case of failure). As a result, it caused multiple issues when the configuration used was different from the one everybody expected and made the code more convoluted obscuring the main logic. This made sense before when everything was a single service and failing in one part could break totally unrelated things, but with smaller that each has their own area of responsibility we should not fall back to the same bad practices. It was intentional design behind not giving people tools to default back to the bad practices, creating discrepancies between reported and used configuration and obscure logic.
> 
> About compilation warnings, the way to do that is by creating code analyzers that ship with the package and could provide warnings, info messages, or even code fixes on incorrect code usage, but that's much more work. The benefit is that you can do that not only for your code, create analyzers about incorrect usages of standard classes. This repo might be a good sample for that https://github.com/xunit/xunit.analyzers
> 
> About attracting attention to the fact that Options could throw, I think it would instead push people to the bad practice since as practice shows it's easier for people to use "safe" method and don't thinking too much about it than accept the fact that the service should fail if it configured incorrectly.

I will add some more thoughts about some of the points you raised:

1. It is true that, semantically speaking, these methods give a wrong idea of safely getting configuration for a service, while instead they are actually handling the exception internally, logging it, and return null, but:
  a. The logger can be made a mandatory parameter in input.
  b. The method currently returns a nullable object that will have to be handled accordingly; although this could result in an increase of defensive programming, or to simply overlook the null object in projects that do not have nullability check enabled, the alternative would simply be uglier in case a fallback value is indeed available, having that the exception would have to be manually handled.
2. It is true that the person on Shield for a particular team can be unaware of the logging mechanism implemented for this kind of error, but by centralising the handling of the exception, further action can be performed to, for instance, trigger the creation of a ticket to report faulty configuration with details.

To avoid inspiring semantic misconceptions regarding these methods, I propose the following changes to their implementation:

1. Change the signature of the methods and/or their description to express the specific case in which they should be used.
2. Make the logger instance mandatory.
3. Allow for a mandatory definition of a fallback value, or a function that gets the value from a valid configuration instance; this should provide a clearer idea of which cases the method should be used for, and would return a non-null instance, solving the problem of projects with nullable check still not enabled.

As an alternative, a good idea to document the behaviour of the existing getter methods would be to ask for the inclusion of the `OptionsValidationException` in the methods documentation, [as it is currently absent](https://cs.github.com/dotnet/runtime/blob/bd175dae3a6fca2986db6135318add78867eceab/src/libraries/Microsoft.Extensions.Options/src/IOptionsMonitor.cs?q=org%3Adotnet+interface+IOptionsMonitor+language%3AC%23), but this would not trigger an error or a warning, as far as I am aware, if the method is called without proper exception handling.

AndreyTretyak:
> > > > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > > > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > > > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > > > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > > > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> > > 
> > > 
> > > I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.
> > > I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.
> > > Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.
> > 
> > 
> > My view on this is that if service is critical:
> > 
> > 1. We should not have a single instance of it.
> > 2. We should not apply configuration changes to all instances of the service at once. (That I believe already in guidelines and even might be enforced by SOS)
> > 
> > Also, as I've pointed out in [reply to Keith](https://github.com/microsoft/Omex/pull/514#issuecomment-1295458505) having crucial service running with incorrect configuration (and lying to all tools that it runs on other configuration) is even more dangerous.
> > Besides that, methods in this PR does not provide any distinction between critical and non-critical services, so since using them looks safer everybody would use them and all configuration becomes unreliable.
> > About the operator checking for logs errors during applying settings override. To me, it looks like a very optimistic scenario. It's a bad idea to hope that people follow the steps that are not enforced. Especially since most of the overrides are done by people on-call or Shilelds that know nothing about the internals of this particular service and what methods are used to get the settings. If we would work on enforcing this procedure, I think, there are much better practices to enforce, for example, gradual configuration rollout with fallback in case of failure or the ability for the service to provide validation rules to the override tool. Also, added methods allow not passing the logger or using NullLogger and in combination with constant log loss by our log aggregator could result in undetectable misconfiguration (both in runtime and deploy time). And if there is a possibility for that it will happen eventually.
> > Omex's previous way of getting configuration provided customers with the same choice of getting configuration value or getting it safely. As a result, all of the configuration requests were done using safe methods (and PR's were blocked because people not using safe ones with default values in case of failure). As a result, it caused multiple issues when the configuration used was different from the one everybody expected and made the code more convoluted obscuring the main logic. This made sense before when everything was a single service and failing in one part could break totally unrelated things, but with smaller that each has their own area of responsibility we should not fall back to the same bad practices. It was intentional design behind not giving people tools to default back to the bad practices, creating discrepancies between reported and used configuration and obscure logic.
> > About compilation warnings, the way to do that is by creating code analyzers that ship with the package and could provide warnings, info messages, or even code fixes on incorrect code usage, but that's much more work. The benefit is that you can do that not only for your code, create analyzers about incorrect usages of standard classes. This repo might be a good sample for that https://github.com/xunit/xunit.analyzers
> > About attracting attention to the fact that Options could throw, I think it would instead push people to the bad practice since as practice shows it's easier for people to use "safe" method and don't thinking too much about it than accept the fact that the service should fail if it configured incorrectly.
> 
> I will add some more thoughts about some of the points you raised:
> 
> 1. It is true that, semantically speaking, these methods give a wrong idea of safely getting configuration for a service, while instead they are actually handling the exception internally, logging it, and return null, but:
>    
>    * The logger can be made a mandatory parameter in input.
>    * The method currently returns a nullable object that will have to be handled accordingly; although this could result in an increase of defensive programming, or to simply overlook the null object in projects that do not have nullability check enabled, the alternative would simply be uglier in case a fallback value is indeed available, having that the exception would have to be manually handled.
> 2. It is true that the person on Shield for a particular team can be unaware of the logging mechanism implemented for this kind of error, but by centralising the handling of the exception, further action can be performed to, for instance, trigger the creation of a ticket to report faulty configuration with details.
> 
> To avoid inspiring semantic misconceptions regarding these methods, I propose the following changes to their implementation:
> 
> 1. Change the signature of the methods and/or their description to express the specific case in which they should be used.
> 2. Make the logger instance mandatory.
> 3. Allow for a mandatory definition of a fallback value, or a function that gets the value from a valid configuration instance; this should provide a clearer idea of which cases the method should be used for, and would return a non-null instance, solving the problem of projects with nullable check still not enabled.
> 
> As an alternative, a good idea to document the behaviour of the existing getter methods would be to ask for the inclusion of the `OptionsValidationException` in the methods documentation, [as it is currently absent](https://cs.github.com/dotnet/runtime/blob/bd175dae3a6fca2986db6135318add78867eceab/src/libraries/Microsoft.Extensions.Options/src/IOptionsMonitor.cs?q=org%3Adotnet+interface+IOptionsMonitor+language%3AC%23), but this would not trigger an error or a warning, as far as I am aware, if the method is called without proper exception handling.

About the point that you've added:
1.1 Making a logger mandatory is an improvement but does not guarantee that logs will be reported. Our log aggregator constantly loses significant about of logs (Artem can tell more about it) and we had multiple cases of logger disabled during misconfiguration or using incorrect logger.
1.2 My point is that ideally there should not be fallback for configuration errors, so by hiding ugliness of exception catching we are making it harder for users to understand that they are doing something wrong. If you absolutely insist that configuration value should be optional it could be done by providing default value in the Option class itself (it won't protect from parsing error, but SOS service should be doing that).
2. As I've pointed out before if we will put an effort into automating detection of such exception it would be much better to spend those resources on not allowing applying overrides to all instances in a single change or adding better validation. Ticket creating is too slow mechanism for reacting to misconfiguration problems (especially when it depends on log parsing that could be missing), since couple seconds of running service with incorrect settings could lead to unrecoverable data corruption. Failing a service is much more reliable in this case.

The steps that you've suggested in API changes are nice but having an API pushes library users in the wrong direction.
Adding a warning to the existing option methods could be done using Roslyn Analyzer, but my main point is that there are no reasonable actions for service when the configuration is incorrect, it should only fail.
It is similar to how each object creation can throw out of memory exception, but we are not wrapping each allocation with try/catch because we can't do anything meaningful to resolve the issue. Same here, a service that received incorrect configuration can't and should not try to fix itself (especially in a way that will create an impression that config change applied), running with misconfiguration is dangerous. We should just fail the service until the proper configuration is provided.

MaxDac:
> > > > > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > > > > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > > > > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > > > > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > > > > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> > > > 
> > > > 
> > > > I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.
> > > > I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.
> > > > Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.
> > > 
> > > 
> > > My view on this is that if service is critical:
> > > 
> > > 1. We should not have a single instance of it.
> > > 2. We should not apply configuration changes to all instances of the service at once. (That I believe already in guidelines and even might be enforced by SOS)
> > > 
> > > Also, as I've pointed out in [reply to Keith](https://github.com/microsoft/Omex/pull/514#issuecomment-1295458505) having crucial service running with incorrect configuration (and lying to all tools that it runs on other configuration) is even more dangerous.
> > > Besides that, methods in this PR does not provide any distinction between critical and non-critical services, so since using them looks safer everybody would use them and all configuration becomes unreliable.
> > > About the operator checking for logs errors during applying settings override. To me, it looks like a very optimistic scenario. It's a bad idea to hope that people follow the steps that are not enforced. Especially since most of the overrides are done by people on-call or Shilelds that know nothing about the internals of this particular service and what methods are used to get the settings. If we would work on enforcing this procedure, I think, there are much better practices to enforce, for example, gradual configuration rollout with fallback in case of failure or the ability for the service to provide validation rules to the override tool. Also, added methods allow not passing the logger or using NullLogger and in combination with constant log loss by our log aggregator could result in undetectable misconfiguration (both in runtime and deploy time). And if there is a possibility for that it will happen eventually.
> > > Omex's previous way of getting configuration provided customers with the same choice of getting configuration value or getting it safely. As a result, all of the configuration requests were done using safe methods (and PR's were blocked because people not using safe ones with default values in case of failure). As a result, it caused multiple issues when the configuration used was different from the one everybody expected and made the code more convoluted obscuring the main logic. This made sense before when everything was a single service and failing in one part could break totally unrelated things, but with smaller that each has their own area of responsibility we should not fall back to the same bad practices. It was intentional design behind not giving people tools to default back to the bad practices, creating discrepancies between reported and used configuration and obscure logic.
> > > About compilation warnings, the way to do that is by creating code analyzers that ship with the package and could provide warnings, info messages, or even code fixes on incorrect code usage, but that's much more work. The benefit is that you can do that not only for your code, create analyzers about incorrect usages of standard classes. This repo might be a good sample for that https://github.com/xunit/xunit.analyzers
> > > About attracting attention to the fact that Options could throw, I think it would instead push people to the bad practice since as practice shows it's easier for people to use "safe" method and don't thinking too much about it than accept the fact that the service should fail if it configured incorrectly.
> > 
> > 
> > I will add some more thoughts about some of the points you raised:
> > 
> > 1. It is true that, semantically speaking, these methods give a wrong idea of safely getting configuration for a service, while instead they are actually handling the exception internally, logging it, and return null, but:
> >    
> >    * The logger can be made a mandatory parameter in input.
> >    * The method currently returns a nullable object that will have to be handled accordingly; although this could result in an increase of defensive programming, or to simply overlook the null object in projects that do not have nullability check enabled, the alternative would simply be uglier in case a fallback value is indeed available, having that the exception would have to be manually handled.
> > 2. It is true that the person on Shield for a particular team can be unaware of the logging mechanism implemented for this kind of error, but by centralising the handling of the exception, further action can be performed to, for instance, trigger the creation of a ticket to report faulty configuration with details.
> > 
> > To avoid inspiring semantic misconceptions regarding these methods, I propose the following changes to their implementation:
> > 
> > 1. Change the signature of the methods and/or their description to express the specific case in which they should be used.
> > 2. Make the logger instance mandatory.
> > 3. Allow for a mandatory definition of a fallback value, or a function that gets the value from a valid configuration instance; this should provide a clearer idea of which cases the method should be used for, and would return a non-null instance, solving the problem of projects with nullable check still not enabled.
> > 
> > As an alternative, a good idea to document the behaviour of the existing getter methods would be to ask for the inclusion of the `OptionsValidationException` in the methods documentation, [as it is currently absent](https://cs.github.com/dotnet/runtime/blob/bd175dae3a6fca2986db6135318add78867eceab/src/libraries/Microsoft.Extensions.Options/src/IOptionsMonitor.cs?q=org%3Adotnet+interface+IOptionsMonitor+language%3AC%23), but this would not trigger an error or a warning, as far as I am aware, if the method is called without proper exception handling.
> 
> About the point that you've added: 1.1 Making a logger mandatory is an improvement but does not guarantee that logs will be reported. Our log aggregator constantly loses significant about of logs (Artem can tell more about it) and we had multiple cases of logger disabled during misconfiguration or using incorrect logger. 1.2 My point is that ideally there should not be fallback for configuration errors, so by hiding ugliness of exception catching we are making it harder for users to understand that they are doing something wrong. If you absolutely insist that configuration value should be optional it could be done by providing default value in the Option class itself (it won't protect from parsing error, but SOS service should be doing that). 2. As I've pointed out before if we will put an effort into automating detection of such exception it would be much better to spend those resources on not allowing applying overrides to all instances in a single change or adding better validation. Ticket creating is too slow mechanism for reacting to misconfiguration problems (especially when it depends on log parsing that could be missing), since couple seconds of running service with incorrect settings could lead to unrecoverable data corruption. Failing a service is much more reliable in this case.
> 
> The steps that you've suggested in API changes are nice but having an API pushes library users in the wrong direction. Adding a warning to the existing option methods could be done using Roslyn Analyzer, but my main point is that there are no reasonable actions for service when the configuration is incorrect, it should only fail. It is similar to how each object creation can throw out of memory exception, but we are not wrapping each allocation with try/catch because we can't do anything meaningful to resolve the issue. Same here, a service that received incorrect configuration can't and should not try to fix itself (especially in a way that will create an impression that config change applied), running with misconfiguration is dangerous. We should just fail the service until the proper configuration is provided.

I understand your point of view, and I understand that the problem is not actually the implementation itself, but rather **how to treat a service with invalid configuration**, and by that which API to provide to the developer consuming it. The options are the following:

1. Let the service fail, so do not do anything when the `Value` or `CurrentValue` getters are called by the `IOptions`es interfaces.

    - **Pro**: it gives immediate feedback on the failed service, avoiding unwanted behaviours due to incorrect configuration.
    - **Con**: the client's user experience would suffer, having that the service would actually not work.

2. Provide fallback manually (i.e. catching the exception manually in the project code).

    - **Pro**: it gives more control to the developer; the client's user experience would not suffer, if the fallback is provided correctly.
    - **Cons**: the fallback could be implemented in a way that will cause the service to eventually fail or behave unexpectedly; no centralised way of handling the configuration validation errors, that may easily result in hiding the error completely.

3. Provide fallback through API methods.

    - **Pro**: the client's user experience would not suffer.
    - **Cons**: again, the fallback (in this case, the null handling) could be done wrong, or not at all (e.g., in projects that do not have null checking enabled); potentially hiding the error indefinitely; causing the application to potentially work unexpectedly.

I can understand the problem you are highlighting, but I am unsure about how to avoid blocking the execution of the service. We should put in place the following:

- Add validation to configuration overrides in SOS.
- Implement a behaviour for microservices installed as multiple services by which only an instance of those service handles the configuration change, potentially failing, while allowing the other instances to work normally in the meantime.

But as far as my knowledge goes, I don't think any of this is actually currently implemented, and I cannot estimate how much time it would require. 
In the meantime, providing a well-documented API could force the developers consuming it to be aware of the problem and provide fallback themselves, for instance by falling back to the configuration parsed at startup using `IOptions` interface, while creating error tickets, even though they are not currently instantaneous.

To resume, possible API alternatives to the ones implemented here could be:

- Changing the API to allow users to define the fallback value directly as I mentioned before by a lambda, thus solving the problem with projects with nullability disabled, and improving documentation.
- Another interface inheriting from `IOptionsMonitor` providing a getter with the last valid configuration, that would have to be implemented thread-safely. I didn't want to implement that because it would be a serious deviation from the standard, rather than an extension of it.
- Documenting strategies to handle configuration validation errors using the Option patter, asking the team handling the `Microsoft.Extensions.Options` maintainer to add documentation about the exception thrown by the configuration value getters.

AndreyTretyak:
> > > > > > I would argue that the service should fail if the configuration is incorrect because otherwise it really hard to find out that your configuration change is not used (and some default applies). It essentially hides the serious error and changes the behavior of the app to the one you don't expect. SOS/ECS or ClusterExlorer would show you the newly applied configuration by service would be running using default value that could lead to all kinds of problems.
> > > > > > Imagine you applying configuration override to take a service out of rotation (or put it into rotation) and making error in the configuration. Instead of knowing right away that something would is wrong, the service would switch to the default value obscuring the error and either all versions would be serving traffic or all of them (both cases are bad).
> > > > > > The same goes for the refactoring hidden behind the configuration flag, you think you've enabled it but made an error, and the service defaults to not using the new code. If the developer misses that incorrect decision could be done about refactoring functionality.
> > > > > > Using default if the configuration is faulty also means that integration testing (ClioudTest) would be unable to detect configuration errors since in most cases it checks that the service is running and healthy, we don't look into logs for it. So this would increase the cases when faulty configurations could be merged unnoticed and deployed to production environments.
> > > > > > In all scenarios that I could think of hiding configuration errors is a bad idea and I think we should not encourage this behaviour.
> > > > > 
> > > > > 
> > > > > I see your point, and I agree that these extension methods should be seldom used, but in case the service is critically important and should work even in case of invalid configuration parsed, they could be a good way of handling human mistakes in a centralised way. I would imagine, for instance, that the operator that changes a configuration at runtime should pay attention to the application logs, and provide a unified tag for searching the error logs could be of help.
> > > > > I would have added a compilation warning associated with these extension methods, but the only one I could find is the `ObsoleteAttribute`, that fails to express semantically what the warning would be. I would gladly accept suggestions on this.
> > > > > Another purpose of the methods would be didactical, they should pose the adequate attention to the fact that the `IOptionsMonitor` and `IOptionsSnapshot` accessor could throw an exception that should be correctly handled.
> > > > 
> > > > 
> > > > My view on this is that if service is critical:
> > > > 
> > > > 1. We should not have a single instance of it.
> > > > 2. We should not apply configuration changes to all instances of the service at once. (That I believe already in guidelines and even might be enforced by SOS)
> > > > 
> > > > Also, as I've pointed out in [reply to Keith](https://github.com/microsoft/Omex/pull/514#issuecomment-1295458505) having crucial service running with incorrect configuration (and lying to all tools that it runs on other configuration) is even more dangerous.
> > > > Besides that, methods in this PR does not provide any distinction between critical and non-critical services, so since using them looks safer everybody would use them and all configuration becomes unreliable.
> > > > About the operator checking for logs errors during applying settings override. To me, it looks like a very optimistic scenario. It's a bad idea to hope that people follow the steps that are not enforced. Especially since most of the overrides are done by people on-call or Shilelds that know nothing about the internals of this particular service and what methods are used to get the settings. If we would work on enforcing this procedure, I think, there are much better practices to enforce, for example, gradual configuration rollout with fallback in case of failure or the ability for the service to provide validation rules to the override tool. Also, added methods allow not passing the logger or using NullLogger and in combination with constant log loss by our log aggregator could result in undetectable misconfiguration (both in runtime and deploy time). And if there is a possibility for that it will happen eventually.
> > > > Omex's previous way of getting configuration provided customers with the same choice of getting configuration value or getting it safely. As a result, all of the configuration requests were done using safe methods (and PR's were blocked because people not using safe ones with default values in case of failure). As a result, it caused multiple issues when the configuration used was different from the one everybody expected and made the code more convoluted obscuring the main logic. This made sense before when everything was a single service and failing in one part could break totally unrelated things, but with smaller that each has their own area of responsibility we should not fall back to the same bad practices. It was intentional design behind not giving people tools to default back to the bad practices, creating discrepancies between reported and used configuration and obscure logic.
> > > > About compilation warnings, the way to do that is by creating code analyzers that ship with the package and could provide warnings, info messages, or even code fixes on incorrect code usage, but that's much more work. The benefit is that you can do that not only for your code, create analyzers about incorrect usages of standard classes. This repo might be a good sample for that https://github.com/xunit/xunit.analyzers
> > > > About attracting attention to the fact that Options could throw, I think it would instead push people to the bad practice since as practice shows it's easier for people to use "safe" method and don't thinking too much about it than accept the fact that the service should fail if it configured incorrectly.
> > > 
> > > 
> > > I will add some more thoughts about some of the points you raised:
> > > 
> > > 1. It is true that, semantically speaking, these methods give a wrong idea of safely getting configuration for a service, while instead they are actually handling the exception internally, logging it, and return null, but:
> > >    
> > >    * The logger can be made a mandatory parameter in input.
> > >    * The method currently returns a nullable object that will have to be handled accordingly; although this could result in an increase of defensive programming, or to simply overlook the null object in projects that do not have nullability check enabled, the alternative would simply be uglier in case a fallback value is indeed available, having that the exception would have to be manually handled.
> > > 2. It is true that the person on Shield for a particular team can be unaware of the logging mechanism implemented for this kind of error, but by centralising the handling of the exception, further action can be performed to, for instance, trigger the creation of a ticket to report faulty configuration with details.
> > > 
> > > To avoid inspiring semantic misconceptions regarding these methods, I propose the following changes to their implementation:
> > > 
> > > 1. Change the signature of the methods and/or their description to express the specific case in which they should be used.
> > > 2. Make the logger instance mandatory.
> > > 3. Allow for a mandatory definition of a fallback value, or a function that gets the value from a valid configuration instance; this should provide a clearer idea of which cases the method should be used for, and would return a non-null instance, solving the problem of projects with nullable check still not enabled.
> > > 
> > > As an alternative, a good idea to document the behaviour of the existing getter methods would be to ask for the inclusion of the `OptionsValidationException` in the methods documentation, [as it is currently absent](https://cs.github.com/dotnet/runtime/blob/bd175dae3a6fca2986db6135318add78867eceab/src/libraries/Microsoft.Extensions.Options/src/IOptionsMonitor.cs?q=org%3Adotnet+interface+IOptionsMonitor+language%3AC%23), but this would not trigger an error or a warning, as far as I am aware, if the method is called without proper exception handling.
> > 
> > 
> > About the point that you've added: 1.1 Making a logger mandatory is an improvement but does not guarantee that logs will be reported. Our log aggregator constantly loses significant about of logs (Artem can tell more about it) and we had multiple cases of logger disabled during misconfiguration or using incorrect logger. 1.2 My point is that ideally there should not be fallback for configuration errors, so by hiding ugliness of exception catching we are making it harder for users to understand that they are doing something wrong. If you absolutely insist that configuration value should be optional it could be done by providing default value in the Option class itself (it won't protect from parsing error, but SOS service should be doing that). 2. As I've pointed out before if we will put an effort into automating detection of such exception it would be much better to spend those resources on not allowing applying overrides to all instances in a single change or adding better validation. Ticket creating is too slow mechanism for reacting to misconfiguration problems (especially when it depends on log parsing that could be missing), since couple seconds of running service with incorrect settings could lead to unrecoverable data corruption. Failing a service is much more reliable in this case.
> > The steps that you've suggested in API changes are nice but having an API pushes library users in the wrong direction. Adding a warning to the existing option methods could be done using Roslyn Analyzer, but my main point is that there are no reasonable actions for service when the configuration is incorrect, it should only fail. It is similar to how each object creation can throw out of memory exception, but we are not wrapping each allocation with try/catch because we can't do anything meaningful to resolve the issue. Same here, a service that received incorrect configuration can't and should not try to fix itself (especially in a way that will create an impression that config change applied), running with misconfiguration is dangerous. We should just fail the service until the proper configuration is provided.
> 
> I understand your point of view, and I understand that the problem is not actually the implementation itself, but rather **how to treat a service with invalid configuration**, and by that which API to provide to the developer consuming it. The options are the following:
> 
> 1. Let the service fail, so do not do anything when the `Value` or `CurrentValue` getters are called by the `IOptions`es interfaces.
>    
>    * **Pro**: it gives immediate feedback on the failed service, avoiding unwanted behaviours due to incorrect configuration.
>    * **Con**: the client's user experience would suffer, having that the service would actually not work.
> 2. Provide fallback manually (i.e. catching the exception manually in the project code).
>    
>    * **Pro**: it gives more control to the developer; the client's user experience would not suffer, if the fallback is provided correctly.
>    * **Cons**: the fallback could be implemented in a way that will cause the service to eventually fail or behave unexpectedly; no centralised way of handling the configuration validation errors, that may easily result in hiding the error completely.
> 3. Provide fallback through API methods.
>    
>    * **Pro**: the client's user experience would not suffer.
>    * **Cons**: again, the fallback (in this case, the null handling) could be done wrong, or not at all (e.g., in projects that do not have null checking enabled); potentially hiding the error indefinitely; causing the application to potentially work unexpectedly.
> 
> I can understand the problem you are highlighting, but I am unsure about how to avoid blocking the execution of the service. We should put in place the following:
> 
> * Add validation to configuration overrides in SOS.
> * Implement a behaviour for microservices installed as multiple services by which only an instance of those service handles the configuration change, potentially failing, while allowing the other instances to work normally in the meantime.
> 
> But as far as my knowledge goes, I don't think any of this is actually currently implemented, and I cannot estimate how much time it would require. In the meantime, providing a well-documented API could force the developers consuming it to be aware of the problem and provide fallback themselves, for instance by falling back to the configuration parsed at startup using `IOptions` interface, while creating error tickets, even though they are not currently instantaneous. The cost in this case could be that, by implementing the definitive strategy at service-orchestration level, we would have to mark these extension methods as `Obsolete` and rely on the developer consuming them to get rid of it.
> 
> To resume, possible API alternatives to the ones implemented here could be:
> 
> * Changing the API to allow users to define the fallback value directly as I mentioned before by a lambda, thus solving the problem with projects with nullability disabled, and improving documentation.
> * Another interface inheriting from `IOptionsMonitor` providing a getter with the last valid configuration, that would have to be implemented thread-safely. I didn't want to implement that because it would be a serious deviation from the standard, rather than an extension of it.
> * Documenting strategies to handle configuration validation errors using the Option patter, asking the team handling the `Microsoft.Extensions.Options` maintainer to add documentation about the exception thrown by the configuration value getters.

**Option 1. Cons:** 
Client experience should not suffer. Even before SOS was created there was a rule that every person applying override should follow: "Do not apply overrides to all instances of the service at once". So if you applied incorrect override to some instances client would be redirected to the one that are guaranty to have correct configuration. On a contrary client calling misconfigured service could result in bad experience or data corruption.

**Option 2/3. Cons:** 
1. Service failing is not the worth consequences. Imagine misconfigured service calling incorrect database/endpoint or passing incorrect information to the partners corrupting data that is impossible to recover. Fallback means that it would take longer for the person doing overrides to notice an error and as a result incorect override actually could be applied to more or all instances before problem discovered. So all instances won't work and clinent experience would suffer.
2. Another important point that you are missing is that SOS, ClusterExplorer, configuration dump and any other standard tool would show developer incorect configuration in this case. That's one of my main problems with with this approach. Service is lying to us and unless you are intimately familiar with its internals it's very hard to detect, when couple seconds of misconfiguration could cause unrecoverable damage.
3. Creating correct fallback is hard. I don't trust service developers doing it right and don't want to add this complexity for them when it's not needed. And they unable to change the fact service reporting one configuration but running with another.

**About API alternatives that you are listing:**
They all discuss different ways service instance should handle configuration error. My main point is that it can't and shouldn't be handled at instance level it's for service orchestrator to deal with.

**In terms of other possible improvements to the process:**
Just fooling the rule "Do not apply overrides to all instances of the service at once" already a protection, unless we would make failure harder to detect :). 
If you are interested in improving SOS very easy UI change is to restrict applying override to all instances at once, requiring some wait period after override applied the half of instances.
More advance options could be check with diagnostic service that instances with this override are healthy before allowing user to apply override. Or doing gradual override automatically when SOS applies changes to single instance waits for some period and then applies to the next one. Ideally it should check service health between overrides and do automatic rollbacks, but just adding wait time between overrides give opportunity for user to notice error and act.
Ideal long term solution is moving to configuration deployments that SF (and hopefully AKS also) knows about and could handle proper rollout strategy for us automatically.

**In terms of using this methods as temporary solution:**
I'm not a big fan of temporary decisions. Removing code is much harder then adding it. Creation of those libraries took two years instead of couple month due to the layers of temporary decisions that where added but never revisited. I don't think that we should raise awareness about the problem service developers unable to solve properly and force them two write more protection/fallback code and then hopefully ask to remove it. It's easy ask person to add try/catch or call a method with "Safe" in the name, but hard to convince that it's safe to remove this protection.
Our job as platform developers, the way I see it, is not pushing more complexity and decisions on service developers, but proving them with the simplest as possible way of writhing business logic and handling all infrastructure related issues for them especially in cases when they have no way of handling it correctly.

neilr81:
There has been a lot written so far, but I would summarise as following:

The absolute worst thing to do is for an application to stop functioning when someone makes a live setting change. The setting override systems are usually disconnected from the service code that defines what values they can be set at making it not feasible to limit setting values on input. Relying on slow rollout systems to catch setting problems is also dangerous as the issue may not be realised until a critical % of users are being impacted.

However, hiding the change is also not a great option as developers/engineers/live site support do not get a clear view on what changes have been made and why and the current state of the system is indeterminate (although I'd argue this is always a case as secret settings are also completely hidden).

The problem then becomes one of education of the developer (i.e. clear guidelines as to what type of Option pattern to use) in combination with clear messaging when something bad occurs (i.e. monitoring when a setting change is picked up, is invalid, and then an alert created).

The best fix is that the systems that change the settings are tightly coupled with the setting values themselves to disallow invalid values up front. As most system are moving away from tight coupling, this isn't a great option.

Which makes the next best fix to handle the expected error and alert on it, understanding that the alerting could be implemented differently per service.

We are currently looking at the R9 settings implementation to see what happens there, but most of these systems are starting to use safe defaults when errors occur and expect engineers/service owners to design accordingly.

AndreyTretyak:
> There has been a lot written so far, but I would summarise as following:
> 
> The absolute worst thing to do is for an application to stop functioning when someone makes a live setting change. The setting override systems are usually disconnected from the service code that defines what values they can be set at making it not feasible to limit setting values on input. Relying on slow rollout systems to catch setting problems is also dangerous as the issue may not be realised until a critical % of users are being impacted.
> 
> However, hiding the change is also not a great option as developers/engineers/live site support do not get a clear view on what changes have been made and why and the current state of the system is indeterminate (although I'd argue this is always a case as secret settings are also completely hidden).
> 
> The problem then becomes one of education of the developer (i.e. clear guidelines as to what type of Option pattern to use) in combination with clear messaging when something bad occurs (i.e. monitoring when a setting change is picked up, is invalid, and then an alert created).
> 
> The best fix is that the systems that change the settings are tightly coupled with the setting values themselves to disallow invalid values up front. As most system are moving away from tight coupling, this isn't a great option.
> 
> Which makes the next best fix to handle the expected error and alert on it, understanding that the alerting could be implemented differently per service.
> 
> We are currently looking at the R9 settings implementation to see what happens there, but most of these systems are starting to use safe defaults when errors occur and expect engineers/service owners to design accordingly.

"The absolute worst thing to do is for an application to stop functioning when someone makes a live setting change."
Imagine a case when the default value for blob storage location is set to test storage (reasonable default). A customer works with this service to save data in production and it looks okay, but when we finally find out mistake in configuration all data saved meanwhile disappears. I would argue that it would've been better for the service to fail right away.

"issue may not be realised until a critical % of users are being impacted" I would argue that this happens when developers catch configuration errors and try to use default :) And it's the job of the override system to make sure that rollout could be done safely.

" I'd argue this is always a case as secret settings are also completely hidden" secrets just hidden and we unable to see them, with this change SOS, ClusterExplorer, Config dump would show us value1 but in reality value2 would be used.

"The problem then becomes one of education of the developer" In my experience educating developers never works until you make incorrect way impossible, only then do they look in docs or search for guidance.

Also, I would like to note that this change would bring us back to the state when CloutTest and INT environments are useless for configuration validation, since ever when none of the configurations is available or correct service would continue working using defaults (that usually points to INT values). For me ephemeral protection from typos during config overrides does not worth it, it's better to try educating developers to override safely or ensure those safe rollout rules in SOS.

artempushkin:
Looks like PR comments is not the best place for such discussions, we should talk about it separately. Also, don't forget that it's a public repo and we might accidentally disclose too much about our environments.

I suggest to take some real life scenarios and discuss them elsewhere. In their current form these methods don't provide a dev experience improvement.

PR #515, opened 2022-10-27
Histogram for passive monitoring

Gnol-VN:
Goal: to support Percentile sampling type for metrics. For example, 99th percentile of Duration metric for an activity. This article explains motivation why percentile is important: https://opentelemetry.io/blog/2022/exponential-histograms/

Task: 
1. Add Histogram instrument type
2. Add a config for consumer to enable Histogram instead of Counter
3. Unit test

How to use: in your project, you create a boolean config section "Monitoring:UseHistogramForActivityMonitoring", set it to `true/True/TRUE` to enable Percentile sampling type

Example of appsettings.json:
```
{
  "Monitoring": {
    "UseHistogramForActivityMonitoring": true
  }
}
```
Example of Service Fabric settings.xml
```
<Settings xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schemas.microsoft.com/2011/01/fabric">
  <Section Name="Monitoring">
    <Parameter Name="UseHistogramForActivityMonitoring" Value="true" />
  </Section>
</Settings>
```


src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -18,17 +19,31 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
 		private readonly Counter<double> m_healthCheckActivityCounter;
+		private readonly Histogram<double> m_activityHistogram;
+		private readonly Histogram<double> m_healthCheckActivityHistogram;
+		private readonly bool m_useHistogramForActivity;
 		private readonly IExecutionContext m_context;
 		private readonly IHostEnvironment m_hostEnvironment;
 		private readonly ArrayPool<KeyValuePair<string, object?>> m_arrayPool;
 
-		public ActivityMetricsSender(IExecutionContext executionContext, IHostEnvironment hostEnvironment)
+		public ActivityMetricsSender(IExecutionContext executionContext, IHostEnvironment hostEnvironment, IConfiguration configuration)
```
AndreyTretyak, line 29: I would advise against using `IConfiguration` directly please use `IOptions` pattern instead to inject only the section you are interested in and avoid doing parsing manually. https://learn.microsoft.com/en-us/dotnet/core/extensions/options

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -61,11 +76,25 @@ public void SendActivityMetric(Activity activity)
...
-				m_activityCounter.Add(durationMs, tagsSpan);
+				if (m_useHistogramForActivity)
+				{
+					m_activityHistogram.Record(durationMs, tagsSpan);
+				}
+				else
+				{
+					m_activityCounter.Add(durationMs, tagsSpan);
+				}
 			}
```
AndreyTretyak, line 98: Rewriting it this way might reduce duplication.
```CSharp
Histogram<double> histogram = m_activityHistogram;
Counter<double> counter = m_activityCounter;
if (activity.IsHealthCheck())
{
	histogram = m_healthCheckActivityHistogram;
	counter = m_healthCheckActivityCounter;
}

if (m_useHistogramForActivity)
{
	histogram.Record(durationMs, tagsSpan);
}
else 
{
	counter.Add(durationMs, tagsSpan);
}
```

src/Activities/Internal/ActivityMetricsSender.cs
AndreyTretyak: You need to use the following method to register the option:
```Csharp
collection
	.AddOptions<ServiceProviderOptions>()
	.BindConfiguration(UseHistogramForActivityMonitoringConfigurationPath);
```
It would bind it to the configuration obtained from DI without you actually having an instance.

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -18,17 +19,31 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
+		public ActivityMetricsSender(IExecutionContext executionContext, IHostEnvironment hostEnvironment, IConfiguration configuration)
 		{
 			m_context = executionContext;
 			m_hostEnvironment = hostEnvironment;
 			m_meter = new Meter("Microsoft.Omex.Activities", "1.0.0");
 			m_activityCounter = m_meter.CreateCounter<double>("Activities");
 			m_healthCheckActivityCounter = m_meter.CreateCounter<double>("HealthCheckActivities");
+			m_activityHistogram = m_meter.CreateHistogram<double>("Activities");
+			m_healthCheckActivityHistogram = m_meter.CreateHistogram<double>("HealthCheckActivities");
+			if (!string.IsNullOrEmpty(configuration.GetSection(UseHistogramForActivityMonitoringConfigurationPath).Value))
```
MaxDac, line 38: nit: maybe leave a line between the if and the assignments.

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -18,17 +19,31 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
+			{
+				bool isConfigParsable = bool.TryParse(configuration.GetSection(UseHistogramForActivityMonitoringConfigurationPath).Value, out m_useHistogramForActivity);
+				if (!isConfigParsable)
```
MaxDac, line 41: nit: same as previous comment

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -59,13 +74,21 @@ public void SendActivityMetric(Activity activity)
 
 			ReadOnlySpan<KeyValuePair<string, object?>> tagsSpan = MemoryExtensions.AsSpan(tags, 0, tagsCount);
 
+			Histogram<double> histogram = m_activityHistogram;
+			Counter<double> counter = m_activityCounter;
 			if (activity.IsHealthCheck())
```
MaxDac, line 79: nit: maybe leave a space between the if and the assignment in the line above.

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -59,13 +74,21 @@ public void SendActivityMetric(Activity activity)
...
 			{
-				m_healthCheckActivityCounter.Add(durationMs, tagsSpan);
+				histogram = m_healthCheckActivityHistogram;
+				counter = m_healthCheckActivityCounter;
+			}
```
MaxDac, line 83: Consider using a ternary operator to assign these two variables, or an `if-else` statement, to avoid double assignment of the same variable.

tests/Activities.UnitTests/Internal/ActivityMetricsSenderTests.cs
```diff
@@ -17,7 +19,70 @@ namespace Microsoft.Omex.Extensions.Activities.UnitTests.Internal
...
+		[DataRow(null!)]
+		public void ActivityMetricsSender_Construction_NonExistConfigSection_ReturnNotNull(string nonExistValue)
+		{
+			Mock<IExecutionContext> contextMock = new();
+			Mock<IHostEnvironment> environmentMock = new();
+			Mock<IConfiguration> mockConfig = new();
+			Mock<IConfigurationSection> mockConfigSection = new();
+
+			mockConfig.Setup(configuration => configuration.GetSection(ActivityMetricsSender.UseHistogramForActivityMonitoringConfigurationPath))
+				.Returns(mockConfigSection.Object);
```
MaxDac, line 32: nit: I would chain the extension method calls vertically to increase readability (in the specific case, put `.GetSection` call one line below, indented with `.Returns`, and then leave a line below the `.Returns`.

tests/Activities.UnitTests/Internal/ActivityMetricsSenderTests.cs
```diff
@@ -17,7 +19,70 @@ namespace Microsoft.Omex.Extensions.Activities.UnitTests.Internal
...
+			mockConfigSection.Setup(section => section.Value).Returns(nonExistValue);
+
+			ActivityMetricsSender sender = new(contextMock.Object, environmentMock.Object, mockConfig.Object);
+			Assert.IsNotNull(sender);
```
MaxDac, line 36: nit: maybe add a line above the `Assert`, to make it clear that the `Assert` is the final part of the *Arrange-Act-Assert* test pattern.

tests/Activities.UnitTests/Internal/ActivityMetricsSenderTests.cs
```diff
@@ -17,7 +19,70 @@ namespace Microsoft.Omex.Extensions.Activities.UnitTests.Internal
...
+			Mock<IHostEnvironment> environmentMock = new();
+			Mock<IConfiguration> mockConfig = new();
+			Mock<IConfigurationSection> mockConfigSection = new();
+
+			mockConfig.Setup(configuration => configuration.GetSection(ActivityMetricsSender.UseHistogramForActivityMonitoringConfigurationPath))
+				.Returns(mockConfigSection.Object);
+			mockConfigSection.Setup(section => section.Value).Returns(parsableConfigValue);
+
+			ActivityMetricsSender sender = new(contextMock.Object, environmentMock.Object, mockConfig.Object);
+			Assert.IsNotNull(sender);
```
MaxDac, line 54: nit: same as comment above

tests/Activities.UnitTests/Internal/ActivityMetricsSenderTests.cs
```diff
@@ -33,13 +98,19 @@ public void SendActivityMetric_ProduceMetrics()
 			contextMock.Setup(e => e.DeploymentSlice).Returns("002");
 			contextMock.Setup(e => e.IsCanary).Returns(true);
 			contextMock.Setup(e => e.IsPrivateDeployment).Returns(false);
-			IExecutionContext context= contextMock.Object;
+			IExecutionContext context = contextMock.Object;
```
MaxDac, line 101: nit: having that the definition of `contextMock` spans multiple lines, I would add a line above the assignment of the `context` variable to increase readability.

tests/Hosting.UnitTests/HostBuilderExtensionsTests.cs
```diff
@@ -21,6 +22,7 @@ public class HostBuilderExtensionsTests
 		public void AddOmexServices_TypesRegistered(Type type)
 		{
 			object collectionObj = new ServiceCollection()
+				.AddSingleton<IConfiguration>(new ConfigurationBuilder().Build())
```
MaxDac, line 25: As specified before, consider using the Option pattern. You're also defining it as a singleton, so you can use the `IOptions` interface, but be aware that using this dependency configuration the code will not be able to parse the updated configuration.

MaxDac (changes requested):

tests/Hosting.UnitTests/HostBuilderExtensionsTests.cs
MaxDac: Is this line still necessary after having implemented the Options pattern?

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -18,17 +21,23 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
 		{
 			m_context = executionContext;
 			m_hostEnvironment = hostEnvironment;
+			m_monitoringOption = monitoringOption;
 			m_meter = new Meter("Microsoft.Omex.Activities", "1.0.0");
 			m_activityCounter = m_meter.CreateCounter<double>("Activities");
 			m_healthCheckActivityCounter = m_meter.CreateCounter<double>("HealthCheckActivities");
+			m_activityHistogram = m_meter.CreateHistogram<double>("Activities");
+			m_healthCheckActivityHistogram = m_meter.CreateHistogram<double>("HealthCheckActivities");
 			m_arrayPool = ArrayPool<KeyValuePair<string, object?>>.Create();
```
MaxDac, line 41: I may be mistaken, but I would avoid calling method in the constructor. They may throw exception or implement complex logic, and the constructor is not the best place for that.
Feel free to ignore this comment if that's not the case, or if you want to address this issue in another PR.

Gnol-VN: It is safe to do so as this basically creates a new object and doesn't change any data. 
I also want to keep them here because Counter/Histogram should not be newly created everytime a request come as it is expensive to do

MaxDac (approved):

dimeehan (approved):

Adnovac (approved):

PR #521, opened 2022-11-15
Added .NET 6 to targets

gajdusep:
After bumping dotnet version to 7.0, some teams started having problems with builds.
This PR is to make the transition smoother by readding .NET 6 to targets (will be completely removed only after an announcement in the future)


K-Cully (approved):

K-Cully:
@gajdusep  Please add a description to the PR

PR #526, opened 2022-11-23
ConfigurationValidation - wrapper for IConfiguration nullable checks

gajdusep:
After changes of nullability of some types in Microsoft.Extensions 7.0.0
- added to simplify and unify nullable checks of IConfiguration objects

### DIFF
--- a/src/Abstractions/ConfigurationValidation.cs
+++ b/src/Abstractions/ConfigurationValidation.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Omex.Extensions.Abstractions
+{
+	/// <summary>
+	/// Class provides common validation methods for Microsoft.Extensions.IConfiguration
+	/// </summary>
+	public static class ConfigurationValidation
+	{
+		/// <summary>
+		/// Omex workaround around IConfigurationSection.Get() returning null variables.
+		/// </summary>
+		/// <typeparam name="T">Type needed from the configuration section</typeparam>
+		/// <param name="configuration">IConfiguration object</param>
+		/// <param name="sectionName">Name of configuration section to get the object from</param>
+		/// <returns>Non-nullable object of type T</returns>
+		/// <exception cref="ArgumentNullException">If the type T is not found in the given section of the configuration</exception>
+		public static T GetNonNullableOrThrow<T>(IConfiguration configuration, string sectionName)
+		{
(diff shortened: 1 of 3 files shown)


K-Cully (approved):

src/Abstractions/ConfigurationValidation.cs
```diff
@@ -0,0 +1,28 @@
...
+	{
+		/// <summary>
+		/// Omex workaround around IConfigurationSection.Get() returning null variables.
+		/// </summary>
+		/// <typeparam name="T">Type needed from the configuration section</typeparam>
+		/// <param name="configuration">IConfiguration object</param>
+		/// <param name="sectionName">Name of configuration section to get the object from</param>
+		/// <returns>Non-nullable object of type T</returns>
+		/// <exception cref="ArgumentNullException">If the type T is not found in the given section of the configuration</exception>
+		public static T GetNonNullableOrThrow<T>(IConfiguration configuration, string sectionName)
```
AndreyTretyak, line 22: It looks like [`GetRequiredSection`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getrequiredsection?view=dotnet-plat-ext-7.0#microsoft-extensions-configuration-configurationextensions-getrequiredsection(microsoft-extensions-configuration-iconfiguration-system-string))
doing the same thing.
And if we need for `Get<T>()` to throw we can probably set [`BinderOptions.ErrorOnUnknownConfiguration`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.binderoptions.erroronunknownconfiguration?view=dotnet-plat-ext-7.0#microsoft-extensions-configuration-binderoptions-erroronunknownconfiguration) globally

gajdusep: The problem with nullables was not on `GetSection` (returns always not-null) but on `Get<T>`. 
Would the `BinderOptions.ErrorOnUnknownConfiguration` get us rid of the nullable warnings? That's why this PR - to have the validation in one place and not repeat the code.

AndreyTretyak: I don't know if `Get<T>` provides any options of ensuring compiler that return type is not nullable. If you won't be able to find one and decide to go ahead with adding custom wrapper I would suggest following:
* Use method name that aligns with other methods of the API, ex 'GetRequired<T>`.
* In the implementation unitize methods that I've mentioned above to get proper explanation if what failed to bind instead of generic exception we are creating now.
* First of all create an API suggestion for ,NET explaining your use case and making sure that they don't have better suggestions with .NET 7 and possibly convincing them to add 'GetRequired<T>` into next version.

PR #528, opened 2022-12-05
Added explicit dependency for health checks

artempushkin:
Microsoft.Extensions.Diagnostics.HealthChecks 7.0.0 targets only net7 and netstandard20. Netstandard20 version (which is consumed by any non-net7 project) fails to reference a required dependency Microsoft.Bcl.AsyncInterfaces, so it has to be added explicitly.

### DIFF
--- a/src/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
+++ b/src/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
@@ -21,6 +21,9 @@
     <PackageReference Include="System.ComponentModel.Annotations" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Condition="'$(TargetFramework)' != '$(DefaultNetCoreVersion)'" Include="Microsoft.Bcl.AsyncInterfaces" />
+  </ItemGroup>
+  <ItemGroup>
     <ProjectReference Include="..\Abstractions\Microsoft.Omex.Extensions.Abstractions.csproj" />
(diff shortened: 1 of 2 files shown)


src/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
```diff
@@ -20,6 +20,9 @@
     <PackageReference Include="Newtonsoft.Json" /> <!-- Bump version of transitive dependency -->
     <PackageReference Include="System.ComponentModel.Annotations" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Condition="'$(TargetFramework)' != '$(DefaultNetCoreVersion)'" Include="Microsoft.Bcl.AsyncInterfaces" />
```
AndreyTretyak, line 24: Any chance healthcheck project could stop target netstandard2.0 I assumed Omex does not have any services that target full framework anymore?

artempushkin: Services target net6 so they get netstandard20 version of dotnet package. We can't upgrade to net7 yet.

neilr81 (approved):

AndreyTretyak (approved):

src/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
AndreyTretyak: Maybe leaving net6.0 and net7.0 would be easier that still supporting netstandard2.0 or at least removing them after full move to net7.0.

PR #535, opened 2023-01-04
Made OmexHealthCheckPublisherHostedService to be used by default

Vshishkar:
Made ```OmexHealthCheckPublisherHostedService``` to be used by default. 
When health checks were added using ```AddServiceFabricHealthChecks``` library used default health check publisher hosted service instead of our own.


neilr81 (approved):

K-Cully:
Can you add more detail here?
I take it the intention is that all health checks registered using this library are using the same publisher.
Is that correct, and if so, what are the advantages and disadvantages?
Please add more detail to the summary.

Vshishkar:
@K-Cully The main advantage is that our health publisher is in development now (we are adding custom periods for publishing reports) and we have control over it.

francedot (approved):

PR #536, opened 2023-01-04
Health Check service exception handing were added

Vshishkar:
There is a bug that if health check could not be initiated (for example missing some dependencies in container) exception is unhandled and health reports are not published. 
This bug was missed due to assumption, that ```DefaultHealthCheckService``` from ASP.Net Core repository does not throw Exceptions. 
There will be later an opened issue and proposed fix to .net core team addressing this, but for now it's better add simple workaround for OMEX.

Here is a picture of how it will look like in SF Health Report

image

### DIFF
--- a/src/Diagnostics.HealthChecks/OmexHealthCheckPublisherHostedService.cs
+++ b/src/Diagnostics.HealthChecks/OmexHealthCheckPublisherHostedService.cs
@@ -243,8 +243,26 @@ internal sealed partial class OmexHealthCheckPublisherHostedService : IHostedSer
 			return _healthCheckPublisherOptions.Value.Predicate(r);
 		};
 
-		// The health checks service does it's own logging, and doesn't throw exceptions.
-		var report = await _healthCheckService.CheckHealthAsync(withOptionsPredicate, cancellationToken).ConfigureAwait(false);
+		HealthReport report;
+		try
+		{
+			// The health checks service does it's own logging
+			report = await _healthCheckService.CheckHealthAsync(withOptionsPredicate, cancellationToken).ConfigureAwait(false);
+		}
+		catch (Exception e)
+		{
+			report = new HealthReport(new Dictionary<string, HealthReportEntry>()
+			{
+				{ "Error", new HealthReportEntry(
+					HealthStatus.Unhealthy,
+					"Exception been thrown while trying to execute Health Check",
+					TimeSpan.Zero,
+					e,
+					null)
+				}
+			}, TimeSpan.Zero);
+		}
+		
 
 		var publishers = _publishers;
 		var tasks = new Task[publishers.Length];


Vshishkar:
Here is a ASP.NET Core issue: https://github.com/dotnet/aspnetcore/issues/45873

src/Diagnostics.HealthChecks/OmexHealthCheckPublisherHostedService.cs
```diff
@@ -243,8 +243,26 @@ private async Task RunAsyncCore((TimeSpan Delay, TimeSpan Period, TimeSpan Timeo
 			return _healthCheckPublisherOptions.Value.Predicate(r);
 		};
 
-		// The health checks service does it's own logging, and doesn't throw exceptions.
-		var report = await _healthCheckService.CheckHealthAsync(withOptionsPredicate, cancellationToken).ConfigureAwait(false);
```
AndreyTretyak, line 247: What's current behavior? I would expect that missing DI dependency in any class should fail the service, since it's an error that happens in 100% of cases and could only be fixed by code change, so PR that introduces an error should not pass-through Validation/CloudTest and failing the service is simplest way of achieving this.

What would be the motivation of suppressing the exception? if you would like to log it should not we at least rethrow it after that?

Vshishkar: You are absolutely right in case if HealthCheck dependencies are all comming from DI, but due to their design there is another way when some dependencies are simply parameters. In this case health check will be created like: `[ActivatorUtilities.CreateInstance<HealthCheckType>(serviceProvider, params: object[])](https://github.com/microsoft/Omex/blob/main/src/Diagnostics.HealthChecks/DependencyInjection/OmexHealthChecksBuilderAddCheckExtensions.cs#L171)`.

Ideal scenario would be to add something lile `validateScopes` in DI, but it will be quite big change and it's not quite clear for me when to validate, because on the moment of registration of helath check there is no quarantee that all services required for it's initialisation are regisrtered in DI container. Also this change sound like something that should be done at asp.net core side. 

Answering your second question it's a qeneral behavoir of health checks. When they throw exception, HealthCheckService cathes them, log them and generate report using this exception. Here I'm trying to do something similar. If I would just throw exception further health report would never be published and node would be reported to Service fabric cluster as healthy.

AndreyTretyak: Understood, thanks for the detailed explanation.

From my perspective the fact that some dependencies are passed explicitly does not change the fact that it's DI configuration error, which could be fixed only by code change, and ideally, I would prefer service to fail when this happens.
Since current behavior for health publisher to swallow exception, your change makes sense, we should at least report failure if we are unable to fail service completely, but I would argue that we should still re-throw exception after reporting failure. This would allow default health publisher to handle exceptions also (log it or do something else in the future).
In addition, sending faulty health report might be also useful to log exception.

Vshishkar: I agree that better would be to have an exception right away (thow it upon creation of Host) but I'm not quite sure how to do it. 

Regarding your remark unfortunatelly current health publisher doesn't swallow exception but just throws it and we don't have any report. 

Could you please elaborate on why should we re-throw excpetion because for me it's not quite clear?

AndreyTretyak: My idea was that we should report failed state (via health report or logger or both) and then re-throw exception for health publisher to handle it in whatever way it sees fighting. If it hides an exception it's their design decisions if it fails the service that's even better.

Essentially, I don't think our publisher should be the one that stops exceptions, even if we processing it.

My only concern is that exception could somehow stop health report from publishing, but I don't think that should happen unless it fails whole service in which case we would've needed report at all.

I'm sorry if I am misunderstanding something, my point is mostly suggestion and I'm not insisting on blocking PR due to it.

AndreyTretyak: Oh, sorry the fact I've missed is that we are now have our custom publisher that replaces ASP .NET Core default one, I assumed this code actually runs inside default one. And I've missed that you are reusing report after error handling. 

So, if we decided to replace publisher how about just adding re-throw here, instead of another try/catch: https://github.com/microsoft/Omex/blob/29776f07c9ba842ec3a76042280452ed0e0f3f43/src/Diagnostics.HealthChecks/OmexHealthCheckPublisherHostedService.cs#L207 It would fail service instead of just health report (since it's code structure issue not service health problem) that should increase chase of finding healthcheck registration error before merging PR.

francedot: My 2 cents: I would minimize any changes to the `OmexHealthCheckPublisherHostedService` - as this is hopefully going away once https://github.com/dotnet/aspnetcore/pull/42646 is merged. This is better addressed as a PR to the aspnetcore repo. Note that in case it is not a public API contract change it could be merged sooner than .NET 8..

Vshishkar: @AndreyTretyak  @francedot well, if we are getting rid of `OmexHealthCheckPublisherHostedService ` then my PR doesn't make much sense. I have an open issue at dotnet, so I'll try to sort it our there

Vshishkar: https://github.com/dotnet/aspnetcore/issues/45873

PR #556, opened 2023-03-27
Add support for attaching transport settings to OmexServiceProxyFactory

### LINKED ISSUE #555: OmexServiceProxyFactory does not support secure remoting
https://github.com/microsoft/Omex/blob/main/src/Services.Remoting/Client/OmexServiceProxyFactory.cs does not support secure remoting.
This prevents full adoption of the library for service using secure S2S communication through SF proxies.

The static interface is not ideal for extension to support secure remoting, as the transport settings must be provided to the ServiceProxyFactory instance.
See: https://learn.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-services-secure-communication

Where multiple secure tunnels are required, sharing transport settings on a static factory mam not be desirable, though it would at least support some scenarios.

K-Cully:
Fixes #555 

Adds a fluent interface to OmexServiceProxyFactory so that transport settings can be provided to the underlying service proxy.
This unblocks using secure remoting with Extensions.

### DIFF
--- a/src/Services.Remoting/Client/OmexServiceProxyFactory.cs
+++ b/src/Services.Remoting/Client/OmexServiceProxyFactory.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 using Microsoft.ServiceFabric.Services.Remoting.Client;
+using Microsoft.ServiceFabric.Services.Remoting.FabricTransport;
 using Microsoft.ServiceFabric.Services.Remoting.V2.FabricTransport.Client;
 
 namespace Microsoft.Omex.Extensions.Services.Remoting.Client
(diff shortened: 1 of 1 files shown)


neilr81 (approved):

K-Cully:
Closing out as is, to enable some support for secure remoting.
Please feel free to review though as I don't believe this is the best way of doing this and am open to creating a follow-on improvement/refactor PR, if others feel the same.

src/Services.Remoting/Client/OmexServiceProxyFactory.cs
```diff
@@ -11,12 +12,30 @@ namespace Microsoft.Omex.Extensions.Services.Remoting.Client
...
-		public static ServiceProxyFactory Instance { get; } = new ServiceProxyFactory(handler =>
+		private static ServiceProxyFactory s_serviceProxyFactory = new ServiceProxyFactory(handler =>
 			new OmexServiceRemotingClientFactory(
 				new FabricTransportServiceRemotingClientFactory(
 					remotingCallbackMessageHandler: handler)));
+
+		/// <summary>
+		/// Binds transport settings to the service proxy factory, for use in secure remoting communication.
+		/// </summary>
+		public static void WithTransportSettings(FabricTransportRemotingSettings transportSettings)
```
artempushkin, line 23: Usage will be awkward if you need to make remoting calls from several places. It might make more sense to implement IServiceProxyFactory and register in DI. Static insecure instance can be preserved for compatibility and obsoleted later. It would be good to review the whole remoting approach as well, see what is the value added and what we can potentially improve.

PR #567, opened 2023-05-03
Updating packages feed to Office Public feed

neilr81:


NuGet.Config
```diff
@@ -5,7 +5,7 @@
   </config>
   <packageSources>
     <clear />
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="oc_PublicPackages" value="https://pkgs.dev.azure.com/office/oc/_packaging/oc_PublicPackages/nuget/v3/index.json" />
```
artempushkin, line 8: What is the motivation behind this change? Is this feed accessible to non-msft users?

PR #573, opened 2023-05-24
Added consistency health check marker for Activities

MaxDac:
## Summary

This change introduces a new marker for Activities, to track **consistency health checks** for an application.

A consistency health check differentiate itself from a normal health check in that it performs consistency checks on the response returned by the endpoint it targets, so contrary to the normal health check it cannot be short-circuited. Having that it is a health check, though, it should not be considered as part of the normal application traffic, meaning that passive monitoring should not mark these calls as user traffic, or consider these for load feedback.

### Further improvements

The [`AbstractHealthCheck` class](https://github.com/microsoft/Omex/blob/dc2abe5f1da79b88f7ceea0fbb1af7826331e64f/src/Diagnostics.HealthChecks/AbstractHealthCheck.cs#L18) will not be changed in this PR. Intervention on this class like abstracting the marking of the activity created in the overridden method can be discussed in another PR.

### DIFF
--- a/src/Abstractions/Activities/ActivityExtensions.cs
+++ b/src/Abstractions/Activities/ActivityExtensions.cs
@@ -38,6 +38,15 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 				StringComparison.OrdinalIgnoreCase);
 
 		/// <summary>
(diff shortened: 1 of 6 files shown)


MaxDac:
Closed because the health check problem needs an architectural rethinking

PR #580, opened 2023-06-23
Removing unnecessary references

neilr81:
Tidying up unused references across projects. Also moving to the new() format.

Some of the newtonsoft overrides might need to be added back but I'm happy to do that in a separate change.


neilr81:
@AndreyTretyak is there any tool to detect unused entries in Directory.Packages.props ?

AndreyTretyak:
> Directory.Packages.props

You can try `dotnet nuget prune` command https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions#dotnet-nuget-prune

src/Diagnostics.HealthChecks/Microsoft.Omex.Extensions.Diagnostics.HealthChecks.csproj
```diff
@@ -15,10 +15,6 @@
     <PackageReference Include="Microsoft.ServiceFabric" />
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" />
     <PackageReference Include="Microsoft.Extensions.Http" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" />
```
artempushkin, line 18: Not necessarily related to this package, but in general projects consuming Omex.Extensions might be consuming transitive dependencies, are we discouraging that?

neilr81: I don't believe this was done on purpose (otherwise it should be commented in some way), is that something we'd want to do?

Gnol-VN (approved):

PR #589, opened 2023-08-02
IOption to Disable Activity Event sender

Gnol-VN:
## Context
Metrics is now sent to both EventSource and OpenTelemetry, which results in duplication.
We want to use OpenTelemetry metrics sender instead of Event sender by EventSource

## Why to make this PR?
To provide an option to disable sending metric by EventSource 

## What I do?
Create IOption to disable ActivityEventSender

## Unit tested

### DIFF
--- a/src/Abstractions/Option/MonitoringOption.cs
+++ b/src/Abstractions/Option/MonitoringOption.cs
@@ -12,10 +12,5 @@ namespace Microsoft.Omex.Extensions.Abstractions.Option
 		/// Path to the setting
 		/// </summary>
 		public static string MonitoringPath = "Monitoring";
-
-		/// <summary>
-		/// Setting to determine whether using Histogram ar Counter for metrics
-		/// </summary>
-		public bool UseHistogramForActivityMonitoring { get; set; }
 	}
 }
--- a/src/Abstractions/Option/MonitoringOption.cs
+++ b/src/Abstractions/Option/MonitoringOption.cs
@@ -16,6 +16,6 @@ namespace Microsoft.Omex.Extensions.Abstractions.Option
 		/// <summary>
 		/// Disable ActivityEventSender so Activity metric is only sent via ActivityMetricsSender
 		/// </summary>
-		public bool DisableEventSender { get; set; }
+		public bool DisableActivityEventSender { get; set; }
 	}
 }
--- a/src/Abstractions/Option/MonitoringOption.cs
+++ b/src/Abstractions/Option/MonitoringOption.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿	// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
 namespace Microsoft.Omex.Extensions.Abstractions.Option
@@ -12,5 +12,10 @@ namespace Microsoft.Omex.Extensions.Abstractions.Option
 		/// Path to the setting
 		/// </summary>
 		public static string MonitoringPath = "Monitoring";
+
+		/// <summary>
+		/// Disable ActivityEventSender so Activity metric is only sent via ActivityMetricsSender
+		/// </summary>
+		public bool DisableEventSender { get; set; }
 	}
 }
--- a/src/Abstractions/Option/MonitoringOption.cs
+++ b/src/Abstractions/Option/MonitoringOption.cs
@@ -1,4 +1,4 @@
-﻿	// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
 namespace Microsoft.Omex.Extensions.Abstractions.Option
--- a/src/Activities/Internal/EventSource/ActivityEventSender.cs
+++ b/src/Activities/Internal/EventSource/ActivityEventSender.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Omex.Extensions.Activities
 
 		public void SendActivityMetric(Activity activity)
 		{
-			if (m_options.Value.DisableEventSender || !m_eventSource.IsEnabled())
+			if (m_options.Value.DisableActivityEventSender || !m_eventSource.IsEnabled())
 			{
 				return;
 			}
--- a/src/Activities/Internal/EventSource/ActivityEventSender.cs
+++ b/src/Activities/Internal/EventSource/ActivityEventSender.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Omex.Extensions.Activities
 
 		public void SendActivityMetric(Activity activity)
 		{
-			if (m_options.Value.DisableActivityEventSender || !m_eventSource.IsEnabled())
+			if (!m_options.Value.ActivityEventSenderEnabled || !m_eventSource.IsEnabled())
 			{
 				return;
 			}
--- a/src/Activities/Internal/EventSource/ActivityEventSender.cs
+++ b/src/Activities/Internal/EventSource/ActivityEventSender.cs
@@ -11,7 +11,7 @@ using Microsoft.Omex.Extensions.Abstractions;
 using Microsoft.Omex.Extensions.Abstractions.Activities;
 using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
 using Microsoft.Omex.Extensions.Abstractions.ExecutionContext;
-using Microsoft.Omex.Extensions.Activities.Option;
+using Microsoft.Omex.Extensions.Activities;
 
 namespace Microsoft.Omex.Extensions.Activities
 {
--- a/src/Activities/Internal/EventSource/ActivityEventSender.cs
+++ b/src/Activities/Internal/EventSource/ActivityEventSender.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using Microsoft.Omex.Extensions.Abstractions;
 using Microsoft.Omex.Extensions.Abstractions.Activities;
 using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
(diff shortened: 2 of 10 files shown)


src/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -6,25 +6,28 @@
...
 			m_eventSource = eventSource;
 			m_serviceName = executionContext.ServiceName;
 			m_logger = logger;
+			m_options = options;
 		}
 
 		public void SendActivityMetric(Activity activity)
 		{
-			if (!m_eventSource.IsEnabled())
+			if (m_options.Value.DisableActivityEventSender || !m_eventSource.IsEnabled())
```
artempushkin, line 30: Is there a way to simply disable (or not enable) the event source?

src/Abstractions/Option/MonitoringOption.cs
artempushkin, line 1: This class should be in the project that uses it. As it stands, you are forcing the consumer to depend on two packages instead of one.

src/Activities/Internal/EventSource/ActivityEventSender.cs
K-Cully: https://github.com/Microsoft/referencesource/blob/master/mscorlib/system/diagnostics/eventing/eventsource.cs

K-Cully: As far as I recall, we generally configure ETW providers at the cluster level, which would be the usual switch for an event source.

K-Cully: But, based on the above source reference, if you can get a handle on the Event Source instance, you could probably get away with simply calling .Dispose() on it in a specific .NET App to disable it locally.

Gnol-VN: I decice to do this because I dont want to disable the EventSource in case this can be used by other functions. It is safer and no harm in doing it through this setting.

K-Cully: The EventSource is an ActivityEventSource that was created explicitly for emitting activities through this Activity Event Sender.
Disabling the Activity Event Sender and Disabling the Activity Event Source have the same net effect of preventing emission of activities through the ActivityEventSource.

K-Cully: Another alternative may be to update SendActivityMetric in [AggregatedActivitiesEventSender](https://github.com/microsoft/Omex/blob/main/src/Activities/Internal/EventSource/AggregatedActivitiesEventSender.cs) to make sending through ActivityEventSender conditional on some overridable configuration.

src/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -6,25 +6,28 @@
...
 using Microsoft.Omex.Extensions.Abstractions.Activities.Processing;
 using Microsoft.Omex.Extensions.Abstractions.ExecutionContext;
+using Microsoft.Omex.Extensions.Abstractions.Option;
 
 namespace Microsoft.Omex.Extensions.Activities
 {
 	internal sealed class ActivityEventSender : IActivitiesEventSender
 	{
-		public ActivityEventSender(ActivityEventSource eventSource, IExecutionContext executionContext, ILogger<ActivityEventSender> logger)
+		public ActivityEventSender(ActivityEventSource eventSource, IExecutionContext executionContext, ILogger<ActivityEventSender> logger, IOptions<MonitoringOption> options)
```
K-Cully, line 20: Why not just remove the call to AddOmexActivitySource from the apps/services that are being updated?
https://github.com/microsoft/Omex/blob/32b76da25bcff08cb8ea7c6882371ce65ad92257/src/Activities/ServiceCollectionExtensions.cs#L25C36-L25C57

src/Activities/Internal/EventSource/ActivityEventSender.cs
artempushkin: I've skimmed a few docs, it seems EventSource is enabled when there is a subscriber listening to it. I don't mind the IOption approach if we understand that it's a temporary hack. Proper solution is to make the entire data flow conditional, it will take quite some digging and refactoring.

Gnol-VN: @K-Cully there is no easy way to disable EventSoruce that I can find, in terms of changing a setting or updating C# code. This way is the easiest and cleanest that can be done

src/Abstractions/Option/MonitoringOption.cs
Gnol-VN: @artempushkin 
I agree with the usual way if the option class is not shared. However, the MonitoringOption here is used for two separate projects:
1) DisableActivityEventSender in Microsoft.Omex.Extensions.Activities.csproj
2) **Disable OmexLogger in Microsoft.Omex.Extensions.Logging** (Which will be in other PR)

Since Extensions.Logging has already imported Abstraction, and since MonitoringOption are shared between two projects, isn't it good to avoid duplication of MonitoringOption? 
Or do you think it is better to have `ActivityEventSenderOption.cs` and `OmexLoggerOption.cs`? I opt for using Abstraction project because it looks cleaner

artempushkin: Yes, I think independent classes should use independent Options classes.

Gnol-VN: I have moved according to your suggestion

src/Activities/Internal/EventSource/ActivityEventSender.cs
K-Cully: Resolving ActivityEventSource and calling .Dispose on it would work.
Either way it is a dirty hack, so I'm fine with whichever, so long as the follow-on removal is completed.

src/Activities/Internal/EventSource/ActivityEventSender.cs
K-Cully: Given that it is a dirty hack in general, an alternative approach, which pollutes less code would be to read IConfiguration directly in AddOmexServices/AddOmexActivitySource and decide whether to call AddOmexActivitySource or not.

src/Activities/Option/ActivityOption.cs
```diff
@@ -1,16 +1,21 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-namespace Microsoft.Omex.Extensions.Abstractions.Option
+namespace Microsoft.Omex.Extensions.Activities.Option
```
artempushkin, line 4:
```suggestion
namespace Microsoft.Omex.Extensions.Activities
```

src/Activities/Option/ActivityOption.cs
```diff
@@ -1,16 +1,21 @@
...
 	/// <summary>
 	/// Monitoring option
 	/// </summary>
-	public class MonitoringOption
+	public class ActivityOption
 	{
 		/// <summary>
 		/// Path to the setting
 		/// </summary>
 		public static string MonitoringPath = "Monitoring";
```
artempushkin, line 14: Is this used anywhere?

src/Activities/Option/ActivityOption.cs
```diff
@@ -1,16 +1,21 @@
...
 // Licensed under the MIT license.
 
-namespace Microsoft.Omex.Extensions.Abstractions.Option
+namespace Microsoft.Omex.Extensions.Activities.Option
 {
 	/// <summary>
 	/// Monitoring option
 	/// </summary>
-	public class MonitoringOption
+	public class ActivityOption
```
artempushkin, line 9:
```suggestion
	public class ActivityEventSenderOptions
```

src/Activities/Option/ActivityOption.cs
```diff
@@ -1,16 +1,21 @@
...
 	{
 		/// <summary>
 		/// Path to the setting
 		/// </summary>
 		public static string MonitoringPath = "Monitoring";
+
+		/// <summary>
+		/// Disable ActivityEventSender so Activity metric is only sent via ActivityMetricsSender
+		/// </summary>
+		public bool DisableActivityEventSender { get; set; }
```
artempushkin, line 19:
```suggestion
		public bool Enabled { get; set; } = true;
```
This will follow the convention that we have for config providers.

src/Activities/Option/ActivityOption.cs
artempushkin, line 1: In dotnet Options classes are put together with classes using them, no need for another level of nesting.

src/Activities/Option/ActivityOption.cs
Gnol-VN: Yes 
image

src/Activities/Option/ActivityOption.cs
Gnol-VN: The answer for this and the above comments is: I want to have all settings are inside "Monitoring" parent, like this 
image

K-Cully: Would be better to name it ActivityEventSenderEnabled or ActivityEventSourceEnabled or something.
Enabled ties in with convention elsewhere and improves readability, while ActivityEventSource rather than ActivityEventSender would make more sense in context of the service that it is configured.
As a service owner, I want to disable event source emission. I don't want to have to know the specific class name in the abstraction layers.

K-Cully (changes requested):

src/Activities/Option/ActivityOption.cs
Gnol-VN: I answer in the bottom comment

src/Activities/Option/ActivityOption.cs
Gnol-VN: I answer in the bottom comment

src/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -6,25 +6,28 @@
...
 			m_eventSource = eventSource;
 			m_serviceName = executionContext.ServiceName;
 			m_logger = logger;
+			m_options = options;
 		}
 
 		public void SendActivityMetric(Activity activity)
 		{
-			if (!m_eventSource.IsEnabled())
+			if (m_options.Value.DisableActivityEventSender || !m_eventSource.IsEnabled())
```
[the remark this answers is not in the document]
Gnol-VN: Updated

PR #591, opened 2023-08-11
Porting of composable Health Checks

MaxDac:
## Summary

Implements the porting of the composable Health Check classes implementation for .NET Health Checks.

Each composable class implements a different behaviour, specialising in only one responsibility, thus granting more granular maintenance and flexibility compared to an implementation that leverages a chain of inheritance. For instance, using the composition it will be theoretically possible to invert the wrapping order of the classes, or remove a composable class in the middle of the "chain".

These composable classes are the following:

- **`ObserverHealthCheck`**: this is the outermost wrapper, as it implements the exception handling and the activity definition.
- **`StartupHealthCheck`**: implements the behaviour by which the health check is executed only at start-up time, to prevent the deployment of a faulty version of the application.
- **`HttpEndpointHealthCheck`**: this class is the core of the HTTP Endpoint health checks, in that it performs the call to the endpoints that must be checked by the active monitoring and offers extension by the injection of a `IHealthCheckResponseChecker` interface, which exposes a method that checks the response returned from the HTTP call.

There are two example of different composition strategies in the project, both pointing to the load feedback endpoint:

#### Liveness Health Check
This health checks simply checks whether the related endpoint is reachable, and the request handling should be short circuited by the server, returning 200. This is accomplished by specifying the `MarkAsLivenessCheck` activity marker.
#### Execution Health Check
In this case, the activity marker selected is simply the `HealthCheckActivityMarker`, which marks the activity as belonging to a health check without dictating the short-circuiting of the request by the server. 
This strategy should be used when the execution of the endpoint is equally important as its availability.

The short-circuiting behaviour in the server endpoint has been implemented using the Action Filter -`LivenessCheckActionFilterAttribute`, that checks whether the endpoint current activity (if existent) has been marked accordingly, as mentioned above.

### Implementation details

Another diagnostic package has been created to avoid adding .NET 7 specific dependencies to the project. In particular, the implementation of the activity markers for request short-circuiting and the `LivenessCheckActionFilterAttribute` dependent on ASP.NET Core dependencies have been implemented in the new package.

Another Unit Test project has been created for the same purpose, as the controller endpoint filter implementation is heavily dependent on .NET 7.

### DIFF
--- a/src/Diagnostics.HealthChecks/Composables/HealthCheckComposablesExtensions.cs
+++ b/src/Diagnostics.HealthChecks/Composables/HealthCheckComposablesExtensions.cs
@@ -1,0 +1,113 @@
+﻿// Copyright (C) Microsoft Corporation. All rights reserved.
+
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks.Composables;
+
+/// <summary>
+/// Extension methods for health check decorators configuration.
+/// </summary>
+public static class HealthCheckComposablesExtensions
+{
+	/// <summary>
+	/// Makes the Health check executable only at startup time, adding also observability to it.
+	/// If the consumer health check of this extension method does not want automatic observability,
+	/// the developer is encouraged to wrap it directly around the <see cref="StartupHealthCheck"/> class.
+	/// </summary>
+	/// <param name="healthCheck">The health check.</param>
+	/// <param name="cache">The memory cache. It will be used to memorise the latest health check result.</param>
+	/// <param name="activitySource">The activity source.</param>
+	/// <param name="logger">The logger.</param>
+	/// <param name="parameters">The health check parameters, set at DI registration time.</param>
+	/// <returns>The health check executable only at deploy time.</returns>
+	public static IHealthCheck AsObservableStartupHealthCheck(
+		this IHealthCheck healthCheck,
+		IMemoryCache cache,
+		ActivitySource activitySource,
+		ILogger logger,
+		HealthCheckParameters? parameters = null)
+	{
+		IHealthCheck startupHealthCheck = new StartupHealthCheck(healthCheck, cache);
+		return new ObservableHealthCheck(parameters ?? new(), startupHealthCheck, activitySource, logger);
+	}
+
+	/// <summary>
+	/// Applies the default error handling and activity around the current health check.
+	/// </summary>
+	/// The instance of the Activity marker. By selecting the implementor, the health check will determine how
+	/// the related endpoint will process the health check request.
+	/// <param name="healthCheck">The health check.</param>
+	/// <param name="activitySource">The activity source.</param>
+	/// <param name="logger">The logger.</param>
+	/// <param name="parameters">The health check parameters, set at DI registration time.</param>
+	/// <returns>The health check with the default configuration and error handling.</returns>
+	public static IHealthCheck AsObservableHealthCheck(
+		this IHealthCheck healthCheck,
+		ActivitySource activitySource,
+		ILogger logger,
+		HealthCheckParameters? parameters = null) =>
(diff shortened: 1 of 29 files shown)


src/Diagnostics.HealthChecks/Internal/HttpEndpointHealthCheck.cs
```diff
@@ -12,6 +13,7 @@
 
 namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
 {
+	[Obsolete("The usage of this class is deprecated, please use composable classes in Microsoft.Omex.Extensions.Diagnostics.HealthChecks.Composables namespace to build health checks.")]
```
K-Cully, line 16:
```suggestion
	[Obsolete("The usage of this class is deprecated and will be removed in a later release, please use composable classes in Microsoft.Omex.Extensions.Diagnostics.HealthChecks.Composables namespace to build health checks.")]
```

src/Diagnostics.HealthChecks/ServiceCollectionExtensions.cs
```diff
@@ -20,6 +21,7 @@ public static class ServiceCollectionExtensions
 		/// <summary>
 		/// Add dependencies for a publisher
 		/// </summary>
+		[Obsolete("This method is deprecated, please use HealthCheckComposablesExtensions class extension methods to compose health checks.")]
```
K-Cully, line 24:
```suggestion
		[Obsolete("This method is deprecated and will be removed in a later release, please use HealthCheckComposablesExtensions class extension methods to compose health checks.")]
```

K-Cully: Throughout

tests/Diagnostics.HealthChecks.UnitTests/HealthChecksBuilderExtensionsTests.cs
```diff
@@ -25,6 +25,7 @@ public class HealthChecksBuilderExtensionsTests
 	{
 		[DataTestMethod]
 		[DynamicData(nameof(GetHeaders), DynamicDataSourceType.Method)]
+		[Obsolete("The health check implementation based on AbstractHealthCheck is obsolete.")]
```
K-Cully, line 28: Why are the unit test methods being marked as obsolete?

K-Cully: elsewhere too

src/Diagnostics.HealthChecks/Composables/HealthCheckComposablesExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+				activityMarker: activityMarker);
+
+	/// <summary>
+	/// Checks whether the response from the endpoint is successful or not by checking its HTTP Status Code.
+	/// </summary>
+	/// <param name="context">The Health Check Context.</param>
+	/// <param name="response">The endpoint response.</param>
+	/// <param name="allowedStatusCodes">The allowed status codes for the response..</param>
+	/// <returns>The health check result.</returns>
+	public static Task<HealthCheckResult> CheckResponseStatusCodeAsync(
```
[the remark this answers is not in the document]
K-Cully: Goes for any other "Async" methods that are using Task.FromResult without async await too

K-Cully (changes requested):
Besides the comments already present, this is OSS; please add documentation.

tests/Diagnostics.HealthChecks.UnitTests/HealthChecksBuilderExtensionsTests.cs
MaxDac: It wouldn't compile otherwise because the projects are set to treat warnings directly as errors.
Another possibility would be to ignore that kind of warnings, but in that case we would risk not getting other obsolete warnings from other packages.

src/Diagnostics.HealthChecks/Composables/HealthCheckComposablesExtensions.cs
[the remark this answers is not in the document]
MaxDac: That actually surprised me, at this point in time in .NET I think that every time a method returns a `Task`, it's best to simply add `async`, because the overhead is nothing compared to these kind of risks.

neilr81 (approved):

K-Cully (approved):

PR #594, opened 2023-08-14
Composable Health Checks documentation

MaxDac:
## Summary

This PR includes the documentation to support the composable health check implementation merged [with this PR](https://github.com/microsoft/Omex/pull/591).

### DIFF
--- a/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
+++ b/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
@@ -1,6 +1,6 @@
 ## Health Check ASP.NET Core helpers and extensions
 
-This project introduces features specifically dependent from the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
+This project introduces features specifically designed for the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
 
 #### Liveness Health Check
 
--- a/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
+++ b/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
@@ -7,5 +7,5 @@ This project introduces features specifically dependent from the ASP.NET Core We
 The liveness health check is a special type of health check that limits its probing inspection to the reachability of the endpoint. This means that it will not execute the endpoint logic, that will be short-circuited by the `LivenessCheckActionFilterAttribute`. The mechanic of this interaction is the following:
 
 - The Health Check implementor will call the `CreateLivenessHttpHealthCheck` method in the [`HealthCheckComposablesExtensions`](./HealthCheckComposablesExtensions.cs) class. This factory method will create a default Http Health Check, but it will tweak the Activity by injecting the `activityMarker` function in the constructor, adding a key to the `Activity` baggage.
-- The Activity will be send to the endpoint.
+- The Activity will be sent to the endpoint.
 - The endpoint will have to be marked with the [`LivenessCheckActionFilterAttribute`](./LivenessCheckActionFilterAttribute.cs). This filter will check whether there's an `Activity` injected in the request, and whether the `Activity` has the required key in its baggage: if that is the case, the endpoint execution will be short-circuited, returning an HTTP 200 status code, otherwise the method will be executed normally.
--- a/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
+++ b/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
@@ -1,0 +1,3 @@
+## Health Check ASP.NET Core helpers and extensions
+
+#### Liveness Health Check
--- a/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
+++ b/src/Diagnostics.HealthChecks.AspNetCore/Readme.md
@@ -1,3 +1,11 @@
 ## Health Check ASP.NET Core helpers and extensions
 
+This project introduces features specifically dependent from the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
+
 #### Liveness Health Check
+
+The liveness health check is a special type of health check that limits its probing inspection to the reachability of the endpoint. This means that it will not execute the endpoint logic, that will be short-circuited by the `LivenessCheckActionFilterAttribute`. The mechanic of this interaction is the following:
+
+- The Health Check implementor will call the `CreateLivenessHttpHealthCheck` method in the [`HealthCheckComposablesExtensions`](./HealthCheckComposablesExtensions.cs) class. This factory method will create a default Http Health Check, but it will tweak the Activity by injecting the `activityMarker` function in the constructor, adding a key to the `Activity` baggage.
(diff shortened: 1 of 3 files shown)


src/Diagnostics.HealthChecks.AspNetCore/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+This project introduces features specifically dependent from the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
+
+#### Liveness Health Check
+
+The liveness health check is a special type of health check that limits its probing inspection to the reachability of the endpoint. This means that it will not execute the endpoint logic, that will be short-circuited by the `LivenessCheckActionFilterAttribute`. The mechanic of this interaction is the following:
+
+- The Health Check implementor will call the `CreateLivenessHttpHealthCheck` method in the [`HealthCheckComposablesExtensions`](./HealthCheckComposablesExtensions.cs) class. This factory method will create a default Http Health Check, but it will tweak the Activity by injecting the `activityMarker` function in the constructor, adding a key to the `Activity` baggage.
+- The Activity will be send to the endpoint.
```
K-Cully, line 10:
```suggestion
- The Activity will be sent to the endpoint.
```

src/Diagnostics.HealthChecks.AspNetCore/Readme.md
```diff
@@ -0,0 +1,11 @@
+## Health Check ASP.NET Core helpers and extensions
+
+This project introduces features specifically dependent from the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
```
K-Cully, line 3: I can't interpret the intent from the first line. Does this make sense?
```suggestion
This project introduces features specifically designed for the ASP.NET Core WebAPI Framework. This will not introduce other composable classes, but it will leverage them to provide a new factory method to create a **Liveness Health Check**, an health check that will only verify that the endpoint is reachable without letting it execute its logic, short-circuiting it using an implementatin of the `ActionFilterAttribute`.
```

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
+## Omex Health Checks
+
+This project and the [`Diagnostics.HealthChecks.AspNetCore`](../Diagnostics.HealthChecks.AspNetCore/Readme.md) offers a range of composable classes, all of which implements the .NET `IHealthCheck` interface, to help the developer build their own Health Checks by reusing different behaviours with the help of extension methods.
```
K-Cully, line 3:
```suggestion
This project and the [`Diagnostics.HealthChecks.AspNetCore`](../Diagnostics.HealthChecks.AspNetCore/Readme.md) offer a range of composable classes, all of which implement the .NET `IHealthCheck` interface. Developers can build their own Health Checks by reusing these classes behaviors' with the help of extension methods.
```

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+This approach enables the ability to create new composable objects to enrich the single Health Check with different behaviour, or substitute an existing one in the composition chain, offering more flexibility to the
```
K-Cully, line 5: flexibility to the what?
*the sentence does not appear to be complete.

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
...
+The `ObservableHealthCheck` will mark the `Activity` using the `MarkAsHealthCheck` method in the [`ActivityExtensions`](../Abstractions/Activities/ActivityExtensions.cs) class.
+
+#### [`StartupHealthCheck`](./Composables/StartupHealthCheck.cs)
+
+This implementation offers the capability of executing the health check until it succeeds, then it will re-use the last cached successful result without repeating the Health Check execution. This behaviour is useful when an health check must be executed only at startup, to check for instance for configuration errors, reporting them back to Service Fabric engine while deploying to mark the new service as faulty, enabling the orchestrator to rollback the deployment.
```
K-Cully, line 22:
```suggestion
This implementation offers the capability of executing the health check until it succeeds, then it will re-use the last cached successful result without repeating the Health Check execution. This behaviour is useful when a health check must be executed only at startup, to check for instance for configuration errors, reporting them back to Service Fabric engine while deploying to mark the new service as faulty, enabling the orchestrator to rollback the deployment.
```
'an' used before a word starting with h only applies where the h is silent (honor, etc.)

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+The developer must keep in mind that an health check that includes the startup health check in its implementation will keep executing if the health check result results in a warning or in an error state; this can happen when the service the health check it's implemented into is created by Service Fabric in response to an automatic scale-up.
```
K-Cully, line 24:
```suggestion
The developer must keep in mind that a health check that includes the startup health check in its implementation will keep executing if the health check result results in a warning or in an error state; this can happen when the service the health check it's implemented into is created by Service Fabric in response to an automatic scale-up.
```

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Implementation examples
+
+Implementing an Health Check using the composable classes provided by default can be done following two patterns:
+
+- **Creating the classes manually**: this approach consists on creating the classes manually inside the constructor of the Health Check that will be registered for the application. The resulting Health Check will have to declare the join set of dependencies needed by all the default classes used, along with its own.
+- **Leveraging the extension methods provided in the project**: this project provides a set of extension methods located in the [`HealthCheckComposablesExtensions`](./Composables/HealthCheckComposablesExtensions.cs) class that will complement the functionality offered by the default composable classes with default composition building for different use cases.
+
+In each cases, having that the Health Check will have to be registered with the DI, the implementor will have to create a class implementing the `IHealthCheck` interface directly, composing the different health checks in the constructor and storing the reference in a private field in the class.
```
K-Cully, line 51:
```suggestion
In each case, having that the Health Check will have to be registered with the DI, the implementor will have to create a class implementing the `IHealthCheck` interface directly, composing the different health checks in the constructor and storing the reference in a private field in the class.
```

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -0,0 +1,105 @@
...
+- `IHttpClientFactory`: it will be the responsibility of the implementor to provide a way to create the `HttpClient` instance. This will be addressed in the [Implementation examples](#implementation-examples).
+- `ActivitySource`: this Health Check will create an `Activity` instance. If a wrapper implementation in the composition classes chain created one already, it should be linked to the newly created one inside this health check.
+
+The `HttpEndpointHealthCheck` offers the possibility to "mark" the created `Activity` with custom markers by defining the `activityMarker` function in the constructor.
+
+### Implementation examples
+
+Implementing an Health Check using the composable classes provided by default can be done following two patterns:
+
+- **Creating the classes manually**: this approach consists on creating the classes manually inside the constructor of the Health Check that will be registered for the application. The resulting Health Check will have to declare the join set of dependencies needed by all the default classes used, along with its own.
```
K-Cully, line 48: I'd suggest swapping the order of these - in the examples below too.
It seems like we'd generally want folks to use the extensions, unless they need the extra flexibility; so that approach should come first.

K-Cully (changes requested):

src/Diagnostics.HealthChecks.AspNetCore/Readme.md
MaxDac: It does, it expresses the sense more clearly than what I wrote

src/Diagnostics.HealthChecks/Readme.md
MaxDac: Thanks for spotting that, solved.

K-Cully (approved):

PR #601, opened 2023-08-22
ActivityMetricSender Extra Dimension fix

Gnol-VN:
## Why?
In previous design, Metric dimensions are taken from `Activity.Baggage` and `Activity.TagObjects`. However, consumers may assign baggage item to indeterministic values such as random ID, Timestamp, duration, random address.
Assigning indeterministic values will lead to unique KeyValue pair explosion that cannot render the metric in any dashboards. 

For example, 
+ If we assign Location dimension to a fixed set of 10 Azure regions, they are fine to display in dashboard.
+ If we assign Location dimension to a random User address, they lead to dashboard explosion.

## What I have done?
1. Create DI functionality to allow consumers to register extra dimensions which are meaningful to ingest.
2. Introduce TagList to replace ArrayPool
3. Create more tests to verify the extra dimensions properly

### DIFF
--- a/src/Activities/ExtraActivityDimensions/ExtraActivityBaggageDimensions.cs
+++ b/src/Activities/ExtraActivityDimensions/ExtraActivityBaggageDimensions.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Omex.Extensions.Activities
 		/// <summary>
 		/// Override the default constructor to specify extra dimension set.
 		/// </summary>
-		/// <param name="overrideDefaultExtraDimensions"></param>
+		/// <param name="overrideDefaultExtraDimensions">Set of extra dimensions which are added to metric point</param>
 		public ExtraActivityBaggageDimensions(HashSet<string> overrideDefaultExtraDimensions) => ExtraDimensions = overrideDefaultExtraDimensions;
 
 		/// <inheritdoc/>
--- a/src/Activities/ExtraActivityDimensions/ExtraActivityBaggageDimensions.cs
+++ b/src/Activities/ExtraActivityDimensions/ExtraActivityBaggageDimensions.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Omex.Extensions.Activities
+{
+	/// <inheritdoc/>
+	public class ExtraActivityBaggageDimensions : IExtraActivityBaggageDimensions
+	{
+		/// <summary>
+		/// Default constructor which has no extra baggage dimensions
+		/// </summary>
+		public ExtraActivityBaggageDimensions() => ExtraDimensions = new HashSet<string>();
+
+		/// <summary>
+		/// Override the default constructor to specify extra dimension set.
+		/// </summary>
+		/// <param name="overrideDefaultExtraDimensions"></param>
+		public ExtraActivityBaggageDimensions(HashSet<string> overrideDefaultExtraDimensions) => ExtraDimensions = overrideDefaultExtraDimensions;
+
+		/// <inheritdoc/>
+		public HashSet<string> ExtraDimensions { get; }
+	}
+}
--- a/src/Activities/Internal/ActivityMetricsSender.cs
+++ b/src/Activities/Internal/ActivityMetricsSender.cs
@@ -56,17 +56,19 @@ namespace Microsoft.Omex.Extensions.Activities
 
 			foreach (string dimension in m_extraBaggageDimension)
 			{
-				if (!string.IsNullOrWhiteSpace(activity.GetBaggageItem(dimension)))
+				string? baggageItem = activity.GetBaggageItem(dimension);
+				if (!string.IsNullOrWhiteSpace(baggageItem))
 				{
-					tagList.Add(dimension, activity.GetBaggageItem(dimension));
+					tagList.Add(dimension, baggageItem);
 				}
 			}
 
 			foreach (string dimension in m_extraTagObjectsDimension)
 			{
-				if (activity.GetTagItem(dimension) != null)
+				object? tagItem = activity.GetTagItem(dimension);
+				if (tagItem != null)
 				{
-					tagList.Add(dimension, activity.GetTagItem(dimension));
+					tagList.Add(dimension, tagItem);
 				}
 			}
 
--- a/src/Activities/Internal/ActivityMetricsSender.cs
+++ b/src/Activities/Internal/ActivityMetricsSender.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
 using System.Linq;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
 using Microsoft.Omex.Extensions.Abstractions.Activities;
--- a/src/Activities/Internal/ActivityMetricsSender.cs
+++ b/src/Activities/Internal/ActivityMetricsSender.cs
@@ -2,16 +2,12 @@
 // Licensed under the MIT license.
 
 using System;
-using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
-using System.Linq;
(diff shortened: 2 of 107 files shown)


src/Activities/ServiceCollectionExtensions.cs
```diff
@@ -39,6 +39,8 @@ public static IServiceCollection AddOmexActivitySource(this IServiceCollection s
 			serviceCollection.TryAddSingleton<ActivityMetricsSender>();
 			serviceCollection.TryAddSingleton<ActivityEventSender>();
 			serviceCollection.TryAddSingleton<IActivitiesEventSender, AggregatedActivitiesEventSender>();
+			serviceCollection.TryAddSingleton<IExtraActivityBaggageDimensions>(_ => new ExtraActivityBaggageDimensions());
```
Gnol-VN, line 42: Pay attention this these two lines. By default, `Result` `MetaData` `SubType` dimensions are collected.
Consumers can override the default dimensions by introducing another dimension set.

For example in their startup.cs, they can do:
> services.AddSingleton<IExtraActivityTagObjectsDimensions>(_ => new ExtraActivityTagObjectsDimensions(new HashSet<string> { "City", "Office" }));

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -21,70 +18,61 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
+			m_extraTagObjectsDimension = extraActivityTagObjectsDimensions.ExtraDimensions;
 		}
 
 		public void SendActivityMetric(Activity activity)
 		{
-			long durationMs = Convert.ToInt64(activity.Duration.TotalMilliseconds);
-
-			int tagsCount = s_customTags.Length + activity.TagObjects.Count() + activity.Baggage.Count();
-
-			KeyValuePair<string, object?>[] tags = m_arrayPool.Rent(tagsCount);
```
Gnol-VN, line 42: You maybe asking why I use `TagList` instead of ArrayPool.
1. This document explains the better performance of TagList https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics
2. The previous implementation used `int tagsCount` which take the length of Baggage before considering they match ExtraDimenion items. This design overcomes the problem of calculating length of array to rent

src/Activities/ServiceCollectionExtensions.cs
daiiniel: Are we loosing the ability to emit custom runtime dimensions? As some of the values of the metrics we will only know them at runtime. For example imagine a task that is executing something after a pipeline runs, where the name of the pipeline is passed and you want to have a dimension with the name of the pipeline to aggregate the metrics after, how would you do it?

artempushkin: The dimension should be called something like "PipelineName" and the set of values must be finite, ideally only a handful of possible values.

Gnol-VN: @daiiniel  No, we are not loosing it. We can register the dimension's name in through `ExtraActivityTagObjectsDimensions`. Our example of "PipelineName" can be seen as "ActivityName". The only difference is you need to register "PipelineName" to the above hashset. "ActivityName" is picked up by default

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -21,70 +18,61 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
+					{ "BuildVersion", m_context.BuildVersion },
+					{ "Environment", m_hostEnvironment.EnvironmentName },
+					{ "Cluster", m_context.Cluster },
+					{ "ApplicationName", m_context.ApplicationName },
+					{ "NodeName", m_context.NodeName },
+					{ "MachineId", m_context.MachineId },
+					{ "DeploymentSlice", m_context.DeploymentSlice },
+					{ "IsCanary", m_context.IsCanary },
+					{ "IsPrivateDeployment", m_context.IsPrivateDeployment }
+				};
```
artempushkin, line 55: From documentation:
When emitting metrics with more than eight tags, the SDK allocates memory on the hot-path. You SHOULD try to keep the number of tags less than or equal to eight. Check if you can extract any shared tags such as MachineName, Environment etc. into Resource attributes.

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -21,70 +18,61 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
 		private readonly Histogram<long> m_healthCheckActivityHistogram;
 		private readonly IExecutionContext m_context;
 		private readonly IHostEnvironment m_hostEnvironment;
-		private readonly ArrayPool<KeyValuePair<string, object?>> m_arrayPool;
+		private readonly HashSet<string> m_extraBaggageDimension;
+		private readonly HashSet<string> m_extraTagObjectsDimension;
 
-		public ActivityMetricsSender(IExecutionContext executionContext, IHostEnvironment hostEnvironment)
+		public ActivityMetricsSender(IExecutionContext executionContext, IHostEnvironment hostEnvironment, IExtraActivityBaggageDimensions extraActivityBaggageDimensions, IExtraActivityTagObjectsDimensions extraActivityTagObjectsDimensions)
```
artempushkin, line 24: Why not use IOptions pattern and get extra dimensions from configuration?

Gnol-VN: We can use IOption as well. but DI here look nice to me

src/Activities/Internal/ActivityMetricsSender.cs
Gnol-VN: I have been keeping this in mind. I can do another PR to reduce unnecessary dimensions such as Environment

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -21,70 +18,61 @@ internal sealed class ActivityMetricsSender : IActivitiesEventSender, IDisposabl
...
 		{
 			m_context = executionContext;
 			m_hostEnvironment = hostEnvironment;
 			m_meter = new Meter("Microsoft.Omex.Activities", "1.0.0");
 			m_activityHistogram = m_meter.CreateHistogram<long>("Activities");
```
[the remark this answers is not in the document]
Gnol-VN: It is recommended to use Histogram for tracking activity. Histogram supports Min, Max, Percentile.
Using different instruments like Counter or Gauge is not applicable here.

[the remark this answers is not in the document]
Gnol-VN: Histogram is proved to work with our monitors, which we saw some monitors in OmexDiagnostics before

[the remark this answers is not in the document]
artempushkin: Don't forget this is a public repo, avoid discussing internal details.

neilr81 (approved):

src/Activities/ExtraActivityDimensions/ExtraActivityBaggageDimensions.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Omex.Extensions.Activities
+{
+	/// <inheritdoc/>
+	public class ExtraActivityBaggageDimensions : IExtraActivityBaggageDimensions
```
K-Cully, line 9: This is excessively verbose, and kind of misleading.
They are not "extra", since the consuming service needs them.
They are just dimensions that are specific/custom to the use-case.
Suggest renaming to CustomBaggageDimensions or similar.

src/Activities/ServiceCollectionExtensions.cs
K-Cully, line 42: Given the use of TryAddSingleton here, is the assumption that anyone registering a IExtraActivityBaggageDimensions implementation does so before AddOmexActivitySource gets called, (the default only gets registered where another implementation doesn't already exist in the DI container)?
If so the documentation should be clear about the hard requirement of DI registration ordering.

artempushkin: That's why I suggested using the IOptions pattern for configuring custom dimensions.

PR #603, opened 2023-08-25
Rename "Extra" to "Custom" for dimension set

Gnol-VN:
Rename "Extra" to "Custom" for dimension set

> Extra has implications of being superflous. It also indicates that adding additional baggage items through other shared libraries would be valid (since this would still constitute "extra" baggage).
However, the real use case for the functionality is to add service/business specific baggage. Since this would be custom to the domain, and required (not superfluous), naming the types etc. with Custom rather than Extra makes more sense.
In summary; it is self-documenting the intended usage by giving library consumers more context.


src/Activities/CustomActivityDimensions/ICustomTagObjectsDimensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.Omex.Extensions.Activities
+{
+	/// <summary>
+	/// Define custom metric dimensions which are extracted from Activity TagObjects  by <see cref="ActivityMetricsSender.SendActivityMetric"/>
+	/// Make sure that the value of any custom dimensions are not random values such as ID, timestamp because it will lead to metric explosion.
+	/// </summary>
+	public interface ICustomTagObjectsDimensions
```
K-Cully, line 12: Would it make sense to rename CustomTagObjectsDimensions to CustomTagDimensions?
It seems more readable to me.
If not don't worry about it.

K-Cully (approved):

src/Activities/CustomActivityDimensions/ICustomTagObjectsDimensions.cs
Gnol-VN: We use Activity.TagsObjects , so I want to keep them uniform in variable name and object field name

PR #611, opened 2023-09-29
Option to disable OmexLogger

Gnol-VN:
We introduce a setting `Monitoring:OmexLoggingEnabled` to allow consumers to enable or disable default OmexLogger (via Microsoft EventSource).


src/Logging/InitializationLogger.cs
```diff
@@ -29,7 +29,7 @@ private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder bui
 				builder.AddConsole();
 			}
 
-			builder.AddOmexLogging();
+			builder.AddOmexLogging(null!); // Because this method cannot access HostBuilder's configuration.
```
K-Cully, line 32: Given that this change is breaking legacy logging for service initialization, it doesn't meet its own goal.

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static class HostBuilderExtensions
 		/// <summary>
 		/// Registering Dependency Injection classes that will provide Service Fabric specific information for logging
 		/// </summary>
-		public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this IServiceCollection collection)
+		public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this IServiceCollection collection, HostBuilderContext? hostBuilderContext)
```
K-Cully, line 44: I don't like this approach.
It is messy.
Given that it is a breaking change for Initialization Logger, I'd be in flavor of making it a fully breaking change - updating the package version would be disabling the old logger. Since both approaches would require a code change and deployment to release, there is no release operational difference.
The main difference is that downstream consumers would need to be aware of the breaking change in the package update.

K-Cully: Alternatively, define extra extension methods where one adds the new dependencies and the other adds the old dependencies, then move the configuration handling back to the service and out of the dependency registration extensions.

src/Logging/ServiceCollectionExtensions.cs
```diff
@@ -29,23 +31,32 @@ public static IServiceCollection AddOmexServiceContext<TServiceContext>(this ISe
...
+		public static IServiceCollection AddOmexLogging(this IServiceCollection serviceCollection, HostBuilderContext? context)
 		{
-			serviceCollection.AddLogging();
-
 			serviceCollection.TryAddTransient<IServiceContext, EmptyServiceContext>();
 			serviceCollection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
+			serviceCollection.AddLogging();
+
+			const string settingName = "Monitoring:OmexLoggingEnabled";
+			bool isEventSourceLoggingEnabled = bool.Parse(context?.Configuration.GetSection(settingName).Get<string>() ?? "true");
```
AndreyTretyak, line 54: Not sure if this works exactly like this, but I would expect something simular to this code should be possible, could you please check?
```suggestion
			bool isEventSourceLoggingEnabled = context?.Configuration.GetSection(settingName).Get<bool>(true);
```

src/Hosting.Services/HostBuilderExtensions.cs
AndreyTretyak: I would be in faivour of doing breaking change and making `AddOmexServiceFabricDependencies` an extension on `HostBuilderContext`, since pattern of optional parameter super unusial and easy to miss for user. We can still keep old extension method if absolutly needed but mark it as obsolete with an explanation of why change needed.

PR #614, opened 2023-10-18
Enabled options to disable Omex telemetry via config

artempushkin:
This opens the way to disable Omex telemetry implementations.

OmexLogger can be controlled by standard logging configuration, usually found in appsettings.json. Option class to disable generation of TimedScope metrics from Activities already existed, binding was improved with validation on start.

To disable both logs and activity metrics, put the following in appsettings.json:
```
{
  "Monitoring": {
    "ActivityEventSenderEnabled": false
  },
  "Logging": {
    "Omex": {
      "LogLevel": {
        "Default": "None"
      }
    }
  }
}
```

Initialization logger for now remains enabled, its usage should be reduced first.

### DIFF
--- a/src/Hosting.Services/HostBuilderExtensions.cs
+++ b/src/Hosting.Services/HostBuilderExtensions.cs
@@ -85,7 +85,6 @@ namespace Microsoft.Omex.Extensions.Hosting.Services
 				builderAction(new ServiceFabricHostBuilder<TService, TContext>(builder));
 
 				IHost host = builder
-					.ConfigureLogging(builder => builder.AddOmexLogging())
 					.ConfigureServices((context, collection) =>
 					{
 						collection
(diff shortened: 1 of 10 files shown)


src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -101,6 +101,7 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 						options.ValidateOnBuild = true;
 						options.ValidateScopes = true;
 					})
+					.ConfigureLogging(builder => builder.AddOmexLogging())
```
AndreyTretyak, line 104: This registration suppose to happen couple lines above https://github.com/microsoft/Omex/blob/712c919c1c5f8370c5105279bbd6b67bd59b9401/src/Hosting.Services/HostBuilderExtensions.cs#L95C9-L95C41

Could we try to make `AddOmexServiceFabricDependencies` an extensions method on `HostBuilder`, and same for `AddOmexServices`, Ideally `AddOmexServices` should not allow be called on ServiceCollection (or at least marked as depricated) as I've mentioned on previous PR that introduced configuration.

artempushkin: This last line basically just enables OmexLogger to read configuration, the logger itself is already added to services at that point. It would be correct to call it as early as possible, but that messes up execution context registration and breaks unit tests. There's no point trying to fix things now, we'll be adopting OTel and deleting old logging and activities code.

neilr81 (approved):

PR #629, opened 2023-11-06
Enable Data Contract Serialization for Remoting Exceptions

Pr1ynka:
We want to discontinue usage of binary formatter based serialization because of security implications. With this change we will switch to data contract based based serialization for remoting exceptions.
More information: https://learn.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-services-exception-serialization

### DIFF
--- a/src/Hosting.Services.Remoting/HostBuilderExtensions.cs
+++ b/src/Hosting.Services.Remoting/HostBuilderExtensions.cs
@@ -22,10 +22,17 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Remoting
 			this ServiceFabricHostBuilder<OmexStatefulService, StatefulServiceContext> builder,
 			string name,
 			FabricTransportRemotingListenerSettings? settings = null)
-				where TService : class, IService =>
-					builder
-						.ConfigureServices((_, services) => services.TryAddTransient<TService>())
-						.AddRemotingListener(name, (provider, _) => provider.GetRequiredService<TService>(), settings);
+				where TService : class, IService
+		{
+			if (settings == null)
+			{
+				settings = new FabricTransportRemotingListenerSettings();
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
+			return builder
+				.ConfigureServices((_, services) => services.TryAddTransient<TService>())
+				.AddRemotingListener(name, (provider, _) => provider.GetRequiredService<TService>(), settings);
+		}
 
 		/// <summary>
 		/// Adds remote listener to stateless service
@@ -34,10 +41,17 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Remoting
 			this ServiceFabricHostBuilder<OmexStatelessService, StatelessServiceContext> builder,
 			string name,
 			FabricTransportRemotingListenerSettings? settings = null)
-				where TService : class, IService =>
-					builder
-						.ConfigureServices((_, services) => services.TryAddTransient<TService>())
-						.AddRemotingListener(name, (provider, _) => provider.GetRequiredService<TService>(), settings);
+				where TService : class, IService
+		{
+			if (settings == null)
+			{
+				settings = new FabricTransportRemotingListenerSettings();
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
+			return builder
+				.ConfigureServices((_, services) => services.TryAddTransient<TService>())
+				.AddRemotingListener(name, (provider, _) => provider.GetRequiredService<TService>(), settings);
+		}
 
 		/// <summary>
 		/// Adds remote listener to stateful service
@@ -64,8 +78,15 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Remoting
 				Func<IServiceProvider, TService, IService> createService,
 				FabricTransportRemotingListenerSettings? settings = null)
 					where TService : IServiceFabricService<TContext>
-					where TContext : ServiceContext =>
-						builder.AddServiceListener(p =>
-							new GenericRemotingListenerBuilder<TService>(name, p, createService, settings));
+					where TContext : ServiceContext
+		{
+			if (settings == null)
+			{
+				settings = new FabricTransportRemotingListenerSettings();
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
+			return builder.AddServiceListener(p =>
+				new GenericRemotingListenerBuilder<TService>(name, p, createService, settings));
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/Hosting.Services.Remoting/HostBuilderExtensions.cs
```diff
@@ -34,10 +41,17 @@ public static class HostBuilderExtensions
...
+		{
+			if (settings == null)
+			{
+				settings = new FabricTransportRemotingListenerSettings();
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
+			return builder
+				.ConfigureServices((_, services) => services.TryAddTransient<TService>())
+				.AddRemotingListener(name, (provider, _) => provider.GetRequiredService<TService>(), settings);
+		}
```
K-Cully, line 54: Why is the remoting listener creation on lines 59 and 67 unaffected?

K-Cully (commented):
I believe there should be changes to the proxy factory creation as well https://github.com/microsoft/Omex/blob/main/src/Services.Remoting/Client/OmexServiceProxyFactory.cs

src/Hosting.Services.Remoting/HostBuilderExtensions.cs
Pr1ynka: I am not sure how to add it for them because it is different type of listener function. Can you give me an example for one of them?

K-Cully: https://github.com/microsoft/Omex/blob/main/src/Hosting.Services.Remoting/RemotingListenerBuilder.cs could be updated to just set ExceptionSerializationTechnique by default, then no change would be required here.

K-Cully: If I am reading the rest of the code here correctly the rest of the calls to .AddRemotingListener in this class end up calling those two extensions. If so, making the change in RemotingListenerBuilder would negate the need to make any changes in this class.

src/Hosting.Services.Remoting/HostBuilderExtensions.cs
```diff
@@ -64,8 +78,15 @@ public static class HostBuilderExtensions
...
-						builder.AddServiceListener(p =>
-							new GenericRemotingListenerBuilder<TService>(name, p, createService, settings));
+					where TContext : ServiceContext
+		{
+			if (settings == null)
+			{
+				settings = new FabricTransportRemotingListenerSettings();
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
+			return builder.AddServiceListener(p =>
```
K-Cully, line 88: This one doesn't call the other AddRemotingListener extensions, so it will need the explicit setting override anyway.

Pr1ynka: This one calls GenericRemotingListenerBuilder which in turn is child of RemotingListenerBuilder. I think we can remove change from HostBuilderExtensions entirely.

src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
```diff
@@ -29,6 +29,11 @@ public abstract class RemotingListenerBuilder<TService> : IListenerBuilder<TServ
 			string name,
 			FabricTransportRemotingListenerSettings? settings = null)
 		{
+			if (settings == null)
```
artempushkin, line 32: From a quick look into SF code, passing settings from outside may interfere with secure remoting, we need to check that carefully.

K-Cully: This is exactly what the SF docs say to do to enable data contract serialization - https://learn.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-services-exception-serialization

src/Services.Remoting/Client/OmexServiceProxyFactory.cs
```diff
@@ -30,6 +30,7 @@ public static class OmexServiceProxyFactory
 		/// </summary>
 		public static void WithCustomTransportSettings(FabricTransportRemotingSettings transportSettings)
 		{
+			transportSettings.ExceptionDeserializationTechnique = FabricTransportRemotingSettings.ExceptionDeserialization.Default;
```
K-Cully, line 33: s_serviceProxyFactory initialization above needs to be updated with transport settings setting FabricTransportRemotingSettings.ExceptionDeserialization.Default too.

K-Cully (changes requested):

src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
artempushkin: Yes, but secure remoting depends on the same settings class and we are not populating the relevant fields. It might work, but I would triple check.

src/Services.Remoting/Client/OmexServiceProxyFactory.cs
Pr1ynka: yes, added it now!

src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
```diff
@@ -29,6 +29,14 @@ public abstract class RemotingListenerBuilder<TService> : IListenerBuilder<TServ
...
+			{
+				if (!FabricTransportRemotingListenerSettings.TryLoadFrom("TransportSettings", out settings))
+				{
+					settings = new FabricTransportRemotingListenerSettings();
+				}
+			}
+			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
```
K-Cully, line 39:
```suggestion
			}
			
			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
```

src/Services.Remoting/Client/OmexServiceProxyFactory.cs
```diff
@@ -20,16 +21,26 @@ namespace Microsoft.Omex.Extensions.Services.Remoting.Client
...
 	{
 		private static ServiceProxyFactory s_serviceProxyFactory = new(handler =>
-			new OmexServiceRemotingClientFactory(
+		{
+			FabricTransportRemotingSettings remotingSettings;
+			if (!FabricTransportRemotingSettings.TryLoadFrom("TransportSettings", out remotingSettings))
+			{
+				remotingSettings = new FabricTransportRemotingSettings();
+			}
+			remotingSettings.ExceptionDeserializationTechnique = FabricTransportRemotingSettings.ExceptionDeserialization.Default;
```
K-Cully, line 30:
```suggestion
			}
			
			remotingSettings.ExceptionDeserializationTechnique = FabricTransportRemotingSettings.ExceptionDeserialization.Default;
```

K-Cully (approved):
The change looks fine, but I'd suggest adding some additional tests to the public/internal APIs that consume these to validate the behaviour.

artempushkin (approved):

PR #630, opened 2023-11-09
Update and register email hash providers

rabbit0v0:
Updated `StaticSaltProvider` to get salt from IOptions;
Update corresponding tests;
Remove unused logger in `StaticSaltProvider`;
Update `ServiceCollectionExtensions` to register correct services.


artempushkin (changes requested):
As discussed, we probably have an internal implementation of this functionality that can be used instead. Once confirmed, this PR should be closed.

PR #639, opened 2023-11-21
[Breaking Change] Make transport settings required

K-Cully:
The current listener and proxy creation logic falls-back on creating insecure listeners and proxies when transport settings are not defined.
This change updates the behaviour to throw when no transport settings are provider or transport settings cannot be loaded from the TransportSettings section of the hosting service's settings.xml file.

Since transport settings can still be defined explicitly, the implementation still allows for insecure connection creation, if the user understands well how to initialise transport settings in an insecure way.
However, most users will use defaults, or follow the Service Fabric documentation which details using LoadFrom to set up secure remoting, so this is not a concern.

FabricTransportSettings.LoadFrom could have been used instead but I think having a descriptive exception message is valuable, so kept TryLoadFrom.

### DIFF
--- a/src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
+++ b/src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-using System;
 using System.Fabric;
+using Microsoft.Omex.Extensions.Services.Remoting;
 using Microsoft.Omex.Extensions.Services.Remoting.Runtime;
 using Microsoft.ServiceFabric.Services.Communication.Runtime;
 using Microsoft.ServiceFabric.Services.Remoting;
@@ -33,7 +33,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Remoting
 		{
 			if (settings == null && !FabricTransportRemotingListenerSettings.TryLoadFrom("TransportSettings", out settings))
 			{
-				throw new InvalidOperationException("Transport security is required for Service Fabric Remoting connections. TransportSettings must be defined in settings.xml.");
+				throw new InsecureRemotingUnsupportedException();
 			}
 
 			settings.ExceptionSerializationTechnique = FabricTransportRemotingListenerSettings.ExceptionSerialization.Default;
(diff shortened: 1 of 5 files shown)


src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
```diff
@@ -20,23 +21,21 @@ public abstract class RemotingListenerBuilder<TService> : IListenerBuilder<TServ
...
 		/// <summary>
 		/// Constructor
 		/// </summary>
+		/// <remarks>Transport security settings should be defined in TransportSettings of a service's settings.xml before creating a listener</remarks>
 		protected RemotingListenerBuilder(
 			string name,
 			FabricTransportRemotingListenerSettings? settings = null)
 		{
-			if (settings == null)
+			if (settings == null && !FabricTransportRemotingListenerSettings.TryLoadFrom("TransportSettings", out settings))
```
K-Cully, line 34: FabricTransportRemotingListenerSettings.LoadFrom could be used here either, but I decided that throwing a very explicit exception was more desirable.

src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
```diff
@@ -20,23 +21,21 @@ public abstract class RemotingListenerBuilder<TService> : IListenerBuilder<TServ
...
 			{
-				if (!FabricTransportRemotingListenerSettings.TryLoadFrom("TransportSettings", out settings))
-				{
-					settings = new FabricTransportRemotingListenerSettings();
-				}
+				throw new InvalidOperationException("Transport security is required for Service Fabric Remoting connections. TransportSettings must be defined in settings.xml.");
```
MaxDac, line 36: Probably not worth it, but given that this message will be used twice in the codebase, would it be better to define a constant?

MaxDac (approved):

src/Hosting.Services.Remoting/RemotingListenerBuilder.cs
K-Cully: I debated creating a custom exception type to unify it but I don't think it matters that much.
Given that you've commented on it too, maybe I should add it.

K-Cully: Added a custom exception type.

MaxDac (approved):

PR #641, opened 2023-11-27
Add CertificatesHealthCheck

Adnovac:
Add CertificatesHealthCheck that verifies whether the certificates have been loaded correctly.


src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
```diff
@@ -0,0 +1,129 @@
...
+	private readonly IHealthCheck m_healthCheck;
+
+	/// <summary>
+	/// The constructor.
+	/// </summary>
+	/// <param name="parameters"></param>
+	/// <param name="logger"></param>
+	/// <param name="activitySource"></param>
+	/// <param name="certificateReader"></param>
+	/// <param name="options"></param>
```
K-Cully, line 35: This is an OSS repo. Please detail public classes, etc. correctly.
Also, please add some detail to the summary.

src/Diagnostics.HealthChecks/CertificatesHealthCheckOptions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Omex.Extensions.Diagnostics.HealthChecks
+{
+
```
K-Cully, line 9: nit: remove blank line

src/Diagnostics.HealthChecks/Readme.md
```diff
@@ -105,3 +105,23 @@ public class CustomHealthCheck : IHealthCheck
 ```
 
 This approach offers more flexibility and customisation, but it introduces more complexity and coupling with this project.
+
+### Certificates Health Check
+
+The Certificates HealthCheck verifies whether certificates have been loaded correctly. To use it, the `CertificatesHealthCheck` class needs to be registered. Here's an example of how to register it using the `AddTypeActivatedCheck` method:
```
K-Cully, line 111: Please add more detail regarding what it verifies.
It also validates that the cert's validity period and private key inclusion.

K-Cully (changes requested):

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
Adnovac: I updated the summary description of the constructor and the class

src/Diagnostics.HealthChecks/Readme.md
Adnovac: I updated the readme and included the description of what exactly it verifies

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
```diff
@@ -0,0 +1,131 @@
...
+		IOptions<CertificatesHealthCheckOptions> options)
+	{
+		HashSet<string> uniqueCertificateSubjectNames = new(options.Value.CertSubjectNames);
+
+		m_healthCheck = new CertificatesHealthCheckInternal(
+			parameters,
+			uniqueCertificateSubjectNames,
+			logger,
+			certificateReader)
+				.AsObservableHealthCheck(activitySource, logger, parameters: parameters);
```
artempushkin, line 52: This health check should have only one responsibility which is ensuring certificates validity. Observable wrapper will be added by the consumer if desired.

Adnovac: The AsObservableHealthCheck extension method only adds activity tracking and exception handling, which are not defined in the HealthCheck logic itself. If we remove the wrapper, the consumer side will be forced to add exact same part into their code, which could lead to unnecessary code duplication across projects. I suggest keeping it as it is to provide a ready-to-go solution.

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
```diff
@@ -0,0 +1,131 @@
...
+			foreach (string certName in m_uniqueCertificateSubjectNames)
+			{
+				X509Certificate2? x509Certificate2 = m_certificateReader.GetCertificateByCommonName(certName);
+				ValidateCertificate(x509Certificate2, certName, message, ref healthStatus);
+			}
+
+			return Task.FromResult(new HealthCheckResult(healthStatus, message.ToString(), data: m_parameters.ReportData));
+		}
+
+		private void ValidateCertificate(X509Certificate2? certificate, string certName, StringBuilder message, ref HealthStatus healthStatus)
```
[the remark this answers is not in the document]
Adnovac: I added unit tests

MaxDac (dismissed):

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
K-Cully: I have to agree regarding the registration pattern being wrong here. However, I don't think we should publicly expose the internal health check class, as that corrupts any effort to unify on the Observable health check pattern.
In my mind the CertificatesHealthCheckInternal class is the actual health check. A health check is a unit of logic that evaluates some aspect of system/service health.
I'd suggest following the extension-led pattern that is already defined for AddEndpointHttpHealthCheck, etc. (https://github.com/microsoft/Omex/blob/main/src/Diagnostics.HealthChecks.AspNetCore/HealthCheckComposablesExtensions.cs) where the health check is internal and the addition of the health check is done through publically exposed health check builder extensions.

artempushkin: Http health check is something of a special case, it's important to set it up correctly, so we hid the complexity behind convenience methods. I don't really have a strong preference, one approach could be similar to what you suggest Keith, in line with multiple examples here: https://github.com/Xabaril/AspNetCore.Diagnostics.HealthChecks/tree/master/src.

artempushkin: In my mind our preferred approach was <Custom health check in the service> -> <Observable> -> <Startup> -> <Library health check>. That would go against .AddXXX pattern.

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
```diff
@@ -0,0 +1,131 @@
...
+			m_parameters = parameters;
+		}
+
+		public Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken token = default)
+		{
+			if (m_uniqueCertificateSubjectNames.Count == 0)
+			{
+				string emptyCertNames = "No certificate names are configured to be checked.";
+				m_logger.LogError(Tag.Create(), emptyCertNames);
+				return Task.FromResult(new HealthCheckResult(HealthStatus.Unhealthy, emptyCertNames, data: m_parameters.ReportData));
```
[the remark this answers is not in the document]
K-Cully: I take it the implication here is that the configuration is unhealthy.
The only reason I can think of to have a cert health check with no certs to check would be in a development environment. How we should configure such a case is debatable.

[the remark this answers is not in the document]
artempushkin: I think it just doesn't merit special handling. We shouldn't be making assumptions on behalf of consumer, if they don't provide a list of certs, health check becomes no-op.

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
artempushkin: After chatting with Anna, these approaches are not exclusive, we can have a convenience .Add method for typical usage and also expose the class for custom handling on consumer side.

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
```diff
@@ -0,0 +1,131 @@
...
+			else if (!certificate.HasPrivateKey)
+			{
+				errorMessage = $"Certificate with subject name '{certName}' does not have a private key.";
+			}
+
+			if (!string.IsNullOrWhiteSpace(errorMessage))
+			{
+				message.AppendLine(errorMessage);
+				m_logger.LogError(Tag.Create(), errorMessage);
+				healthStatus = HealthStatus.Unhealthy;
```
artempushkin, line 121: Non-success should return context.Registration.FailureStatus.

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
Adnovac: I added .Add method and removed wrapping class

src/Diagnostics.HealthChecks/CertificatesHealthCheck.cs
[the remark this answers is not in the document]
Adnovac: Similar to Keith, I would assume that if a consumer intentionally adds a Certificates health check to the code, then having at least one certificate is expected. If no certificates were detected, it is visible at first glance that the health check was not configured correctly and its unhealthy. I left it for now, but if you strongly feel against it, I can remove it.

K-Cully (approved):

MaxDac (approved):

PR #644, opened 2023-12-13
Add legacy correlationid param

marc-hayek:
Adding the parameter "correlationId" to the list of legacy correlation Id parameters

### DIFF
--- a/src/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
+++ b/src/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
@@ -113,6 +113,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Services.Web.Middlewares
 
 		private static readonly string[] s_transactionsNames = {
 			"corrtid",					// Correlation transaction query parameter
+   			"correlationId",
 			"X-TransactionId" };
 
 		private static readonly string[] s_correlationIdNames = {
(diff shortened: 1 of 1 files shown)


src/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -113,6 +113,7 @@ private void ExtractCorrelationFromRequest(HttpRequest request)
 
 		private static readonly string[] s_transactionsNames = {
 			"corrtid",					// Correlation transaction query parameter
+   			"correlationId",
```
Gnol-VN, line 116: Do you think you should have it in the array below on line 121 instead? This is for transaction

marc-hayek: for the use case we have it's just parameter being sent, not in the headers

marc-hayek: fixed

src/Hosting.Services.Web/Middlewares/ObsoleteCorrelationHeadersMiddleware.cs
```diff
@@ -120,6 +120,7 @@ private void ExtractCorrelationFromRequest(HttpRequest request)
 			"X-MS-Correlationid",
 			"MS-CorrelationId",			// Correlation Id header used by other Microsoft services
 			"corr",						// Correlation query parameter
+   			"correlationId",
```
K-Cully, line 123: Why is there a need to add another parameter here?
Why can't the corr parameter just be used?

marc-hayek: rex is getting passed correlationId from partners and it would cause downtime to change it to corr on their side and our side while the current experiments are running. We need this logged to track and issue

K-Cully: Approving on the premise that this is temporary and is removed in quick order.

K-Cully (approved):

PR #646, opened 2024-01-08
Changed BaseExecutionContext.BuildVersion to numeric

artempushkin:
BuildVersion was returning FileVersionInfo.ProductVersion for the entry assembly, but with .NET 8 assembly informational version includes commit hash for SourceLink. This is a breaking change for logic that relies on well-known x.x.x.x version format.

More details here: https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link

Rather than disable that behavior completely, I prefer to change BuildVersion logic, this way the native version is still available to dev tools and any other potential use.

### DIFF
--- a/src/Abstractions/ExecutionContext/BaseExecutionContext.cs
+++ b/src/Abstractions/ExecutionContext/BaseExecutionContext.cs
@@ -6,6 +6,7 @@ using System.Diagnostics;
 using System.Net;
 using System.Net.Sockets;
 using System.Reflection;
+using System.Runtime.ConstrainedExecution;
 using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.Omex.Extensions.Abstractions.ExecutionContext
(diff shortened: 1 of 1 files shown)


src/Abstractions/ExecutionContext/BaseExecutionContext.cs
```diff
@@ -140,10 +141,16 @@ public BaseExecutionContext(IHostEnvironment? hostEnvironment = null)
 		/// </summary>
 		protected static string GetBuildVersion()
 		{
+			string buildVersion = DefaultEmptyValue;
```
K-Cully, line 144: Can a regression test be added for this?

K-Cully (approved):

src/Abstractions/ExecutionContext/BaseExecutionContext.cs
artempushkin: Tried, original implementation shows the test runner version "17.8.0.0" without the hash so unit testing wouldn't catch the breaking change, it only affects our services at runtime.

K-Cully (approved):

PR #650, opened 2024-01-17
Removing NetStandard support for OMEX

neilr81:


src/Activities/Internal/ActivityLoggerHelpers.cs
```diff
@@ -22,7 +22,7 @@ internal static class ActivityLoggerHelpers
 			builder.Append(name).Append(':');
 
 		/// <summary>
-		/// Appends collection of key value pairs to StringBuilder, by calling ToSting on key and value
+		/// Appends collection of key value pairs to StringBuilder, by calling ToString on key and value
 		/// </summary>
 		/// <remarks>
 		/// TODO: should be replaced by AppendJoin when we remove netstandard target, since KeyValuePair also overrides ToString
```
neilr81, line 28: @K-Cully what change can be made here now?

K-Cully: AppendJoin did not exist in NetStandard.
As far as I understand, this entire method is now redundant and can be dropped in favor of just using AppendJoin https://learn.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.appendjoin?view=net-8.0

Directory.Build.props
```diff
@@ -8,8 +8,7 @@
     <OldestSupportedDotNetVersion>net8.0</OldestSupportedDotNetVersion>
     <NetCoreVersions>$(LatestSupportedDotNetVersion)</NetCoreVersions>
     <NetCoreVersions Condition="'$(LatestSupportedDotNetVersion)' != '$(OldestSupportedDotNetVersion)'">$(LatestSupportedDotNetVersion);$(OldestSupportedDotNetVersion)</NetCoreVersions>
-    <NetStandardVersions>netstandard2.0</NetStandardVersions>
```
K-Cully, line 11: Please also remove the Condition Variables property group below now too.
IsNetStandard and IsNetFramework are no longer relevant and IsNetCore is the only remaining condition so it is also redundant.

K-Cully (dismissed):

K-Cully (approved):

PR #653, opened 2024-02-08
Revert "Remove .NET7 Support"

neilr81:
Adding .NET 7 for library support.

Some dependencies are slow at rolling out support for the latest .NET version, this will allow those projects to target older frameworks.

This was previously handled by having .NET Standard support but continually holds back the project code.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -25,19 +25,19 @@
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="8.0.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.1" />
   </ItemGroup>
-  <ItemGroup Label="DotNet Package Versions." Condition="'$(TargetFramework)' == '$(OldestSupportedDotNetVersion)'">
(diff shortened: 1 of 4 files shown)


Directory.Packages.props
```diff
@@ -25,6 +25,27 @@
     <PackageVersion Include="System.Diagnostics.DiagnosticSource" Version="8.0.0" />
     <PackageVersion Include="System.Text.Json" Version="8.0.1" />
   </ItemGroup>
+  <ItemGroup Label="DotNet 7 Package Versions." Condition="'$(TargetFramework)' == '$(OldestSupportedDotNetVersion)'">
```
K-Cully, line 28: This condition probably needs to be extended to ensure they are not the same.
```suggestion
  <ItemGroup Label="Previous dotnet Package Versions." Condition="'$(TargetFramework)' == '$(OldestSupportedDotNetVersion)' And '$(OldestSupportedDotNetVersion)' != '$(LatestSupportedDotNetVersion)'">

```

K-Cully (dismissed):

K-Cully (approved):

PR #671, opened 2024-04-22
Adding SetParentAsDependent Activity Extension

IsaacWalker:
The intention here is to allow us to track activity metrics with respect to the scope that depends on it.
This new setting, will set the name of the parent activity as a new dimension in the child activity.

The main driver here is for parameterized dashboards, particularly for TSGs, that can show QoS of all dependencies for a particular activity when that activity runs into issue.

### DIFF
--- a/src/Abstractions/Activities/ActivityExtensions.cs
+++ b/src/Abstractions/Activities/ActivityExtensions.cs
@@ -105,6 +105,22 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 			activity.SetTag(ActivityTagKeys.Metadata, metadata);
 
 		/// <summary>
+		/// Set's the activities parent as the dependent property.
+		/// </summary>
+		/// <returns>This property will only be set if the parent exists.</returns>
+		public static Activity SetParentAsDependent(this Activity activity)
+		{
+			Activity? parent = activity.Parent;
+
+			if (parent != null && !string.IsNullOrEmpty(parent.OperationName))
+			{
+				activity.SetTag(ActivityTagKeys.Dependent, parent.OperationName);
+			}
+
+			return activity;
+		}
+
+		/// <summary>
 		/// Get correlation guid that is used by old Omex services
 		/// </summary>
 		[Obsolete(CorrelationIdObsoleteMessage, false)]
--- a/src/Abstractions/Activities/ActivityExtensions.cs
+++ b/src/Abstractions/Activities/ActivityExtensions.cs
@@ -105,22 +105,6 @@ namespace Microsoft.Omex.Extensions.Abstractions.Activities
 			activity.SetTag(ActivityTagKeys.Metadata, metadata);
 
 		/// <summary>
-		/// Set's the activities parent as the dependent property.
-		/// </summary>
-		/// <returns>This property will only be set if the parent exists.</returns>
-		public static Activity SetParentAsDependent(this Activity activity)
-		{
-			Activity? parent = activity.Parent;
-
-			if (parent != null && !string.IsNullOrEmpty(parent.OperationName))
-			{
-				activity.SetTag(ActivityTagKeys.Dependent, parent.OperationName);
-			}
-
-			return activity;
-		}
-
-		/// <summary>
 		/// Get correlation guid that is used by old Omex services
 		/// </summary>
 		[Obsolete(CorrelationIdObsoleteMessage, false)]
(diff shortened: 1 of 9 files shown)


src/Activities/Internal/EventSource/ActivityEventSender.cs
artempushkin, line 1: This code will be deprecated soon.

src/Abstractions/Activities/ActivityExtensions.cs
```diff
@@ -104,6 +104,22 @@ public static class ActivityExtensions
 		public static Activity SetMetadata(this Activity activity, string metadata) =>
 			activity.SetTag(ActivityTagKeys.Metadata, metadata);
 
+		/// <summary>
+		/// Set's the activities parent as the dependent property.
+		/// </summary>
+		/// <returns>This property will only be set if the parent exists.</returns>
+		public static Activity SetParentAsDependent(this Activity activity)
+		{
+			Activity? parent = activity.Parent;
```
artempushkin, line 113: With OTel you already have all links between activities, there's no need for an extra dimension.

artempushkin (changes requested):

src/Abstractions/Activities/ActivityExtensions.cs
IsaacWalker: What existing dimension captures the name of the parent activity?

artempushkin: In logs you have the parent span ID. Adding freeform string dimension to a metric may result in cardinality explosion, we shouldn't do that. Do you have a particular scenario where you want this?

IsaacWalker: This is for Metrics, not for Logs or Traces.

Essentially, I want to be able to automatically find (in dashboards), all the activities (by name) that a particular activity depends on. In this context, "depends on" will be it's child activities.

i.e. Let's say I have a parameterized dashboard that takes an activity name. I put in the name "MyEndpoint" and it shows that activity and also the activities that "MyEndpoint" depends on, by setting the Dependent to "MyEndpoint".

As you've mentioned we can't use anything unique to a particular execution of an activity, such as span, because of the cardinality. Furthermore, it's just wouldn't work for this scenario - we need something unaffected by aggregation.

src/Activities/Internal/EventSource/ActivityEventSender.cs
IsaacWalker: Shall I undo the change?

AnkitaKalra18 (approved):

src/Activities/Option/ActivityOption.cs
```diff
@@ -17,5 +17,10 @@ internal class ActivityOption
 		/// Disable ActivityEventSender so Activity metric is only sent via ActivityMetricsSender
 		/// </summary>
 		public bool ActivityEventSenderEnabled { get; set; } = true;
+
+		/// <summary>
+		/// Sets each activities parent name as a dimension value.
+		/// </summary>
+		public bool SetParentNameAsDimension { get; set; } = false;
```
Gnol-VN, line 24:
```suggestion
		public bool SetParentNameAsDimensionEnabled { get; set; } = false;
```It is normally have "Enabled" suffix for boolean option

File: src/Activities/Internal/ActivityMetricsSender.cs [hunk 3]
```diff
@@ -73,6 +81,16 @@ public void SendActivityMetric(Activity activity)
 				}
 			}
 
+			bool isSetParentNameAsDimensionEnabled = m_activityOptions.Value.SetParentNameAsDimension;
```
Gnol-VN, line 84: In terms of performance, I believe that having m_SetParentNameAsDimensionEnabled as a class property would improve the performance, because we do not need to access IOption object > access value > access SetParentNameAsDimension object property. 
This will reduce 2 additional accessing steps.

src/Activities/Internal/ActivityMetricsSender.cs
```diff
@@ -73,6 +81,16 @@ public void SendActivityMetric(Activity activity)
...
+			if(isSetParentNameAsDimensionEnabled)
+			{
+				Activity? parent = activity.Parent;
+				if (parent != null && !string.IsNullOrEmpty(parent.OperationName))
```
Gnol-VN, line 88:
```suggestion
				if (!string.IsNullOrEmpty(parent.OperationName))
```I believe the first nullcheck is redundant because it is alreadyt covered by string.isNullOrEmpty

Gnol-VN (approved):

src/Activities/Internal/ActivityMetricsSender.cs
artempushkin: Differences in performance are negligible, IOptions.Value is created and cached at first access.

src/Abstractions/Activities/ActivityExtensions.cs
IsaacWalker: We discussed offline: there is no way currently in OTEL to get this information for metrics.

artempushkin (commented):
Ok for now, we should check what troubleshooting opportunities OTel provides out of the box.

artempushkin (approved):

PR #673, opened 2024-05-10
Deprecation message for legacy telemetry

Gnol-VN:


src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -41,6 +41,7 @@ public static class HostBuilderExtensions
 		/// <summary>
 		/// Registering Dependency Injection classes that will provide Service Fabric specific information for logging
 		/// </summary>
+		[Obsolete($"This method also adds Legacy OmexLogger and ActivityEventSender and they are deprecated. This legacy telemetry is pending for removal by 1 July 2024. Consider adding a different telemetry solution. Code: 8913598")]
```
artempushkin, line 44: Obsoletion only makes sense for methods directly called by consumer code. Also don't forget that this is a public repo, don't add internal details.

Gnol-VN: Hey, I mark them in some other places in case there is a contributor wants to use the code. 
Also, the code 8913598 is naive to external audience, but it helps us for better tracking in the future change.

tests/Activities.UnitTests/Internal/ActivitiesEventSourceTests.cs
```diff
@@ -17,6 +17,7 @@
 namespace Microsoft.Omex.Extensions.Activities.UnitTests
 {
 	[TestClass]
+	[Obsolete($"{nameof(ActivityEventSender)} is obsolete and pending for removal by 1 July 2024. Code: 8913598")]
```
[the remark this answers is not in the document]
Gnol-VN: Same usage as #pragma

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -41,6 +41,7 @@ public static class HostBuilderExtensions
 		/// <summary>
 		/// Registering Dependency Injection classes that will provide Service Fabric specific information for logging
 		/// </summary>
+		[Obsolete($"This method also adds Legacy OmexLogger and ActivityEventSender which are deprecated. They are pending for removal by 1 July 2024. Consider adding a different telemetry solution. Code: 8913598")]
```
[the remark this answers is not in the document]
Gnol-VN: yes, make sense, I use #pragma here instead

src/Logging/ServiceCollectionExtensions.cs
```diff
@@ -42,6 +44,7 @@ public static ILoggingBuilder AddOmexLogging(this ILoggingBuilder builder)
 		/// </summary>
 		/// <param name="serviceCollection">The extension method argument</param>
 		/// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained</returns>
+		[Obsolete($"{nameof(OmexLogger)} and {nameof(OmexLogEventSource)} are obsolete and pending for removal by 1 July 2024. Please consider using a different Logger. Code: 8913598.")]
```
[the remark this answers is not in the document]
Gnol-VN: Cool. I moved them to `AddOmexServiceFabricDependencies`

src/Activities/Internal/EventSource/ActivityEventSender.cs
```diff
@@ -15,6 +15,7 @@
 
 namespace Microsoft.Omex.Extensions.Activities
 {
+	[Obsolete($"{nameof(ActivityEventSender)} is obsolete and pending for removal by 1 July 2024. Code: 8913598")]
```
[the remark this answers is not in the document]
Gnol-VN: Cool. Changed to OMEX188

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -60,7 +61,16 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 
 			collection.TryAddSingleton<IServiceContext, OmexServiceFabricContext>();
 			collection.TryAddSingleton<IExecutionContext, ServiceFabricExecutionContext>();
+
+			collection.AddLogging();
+
+			collection.TryAddTransient<IServiceContext, EmptyServiceContext>();
```
K-Cully, line 67: Why register these types as transient after registering the interfaces with different types (and as Singleton) above?
Suggest adding a comment here.

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -101,16 +112,20 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 						options.ValidateOnBuild = true;
 						options.ValidateScopes = true;
 					})
-					.ConfigureLogging(builder => builder.AddOmexLogging())
 					.Build();
 
+#pragma warning disable OMEX188 // InitializationLogger using OmexLogger is obsolete. DiagnosticId = "OMEX188"
 				InitializationLogger.LogInitializationSucceed(serviceNameForLogging);
+#pragma warning restore OMEX188 // InitializationLogger using OmexLogger is obsolete. DiagnosticId = "OMEX188"
```
K-Cully, line 119: Where are the equivalent log events for the OpenTelemetry based logger?

src/Hosting/Certificates/InitializationCertificateReader.cs
```diff
@@ -16,6 +16,8 @@ public static class InitializationCertificateReader
 		/// <summary>
 		/// Instance of CertificateReader
 		/// </summary>
+#pragma warning disable OMEX188 // InitializationLogger using OmexLogger is obsolete and is pending for removal by 1 July 2024. DiagnosticId = "OMEX188"
```
K-Cully, line 19: It would make more sense to mark this class as Obsolete as well, since the only thing in it is this obsolete functionality.

src/Logging/Internal/EventSource/EmptyServiceContext.cs
```diff
@@ -8,10 +8,16 @@ namespace Microsoft.Omex.Extensions.Logging
 	/// <summary>
 	/// IServiceContext without any information
 	/// </summary>
-	internal class EmptyServiceContext : IServiceContext
+	public class EmptyServiceContext : IServiceContext
```
K-Cully, line 11: Why make this public?

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: this code is copied from `Logging/ServiceCollectionExtensions.cs`. Do you suggest changing to TryAddSingleton instead (which makes sense) ?

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: OpenTelemetry is not (yet) supported in this public repo.

src/Logging/Internal/EventSource/EmptyServiceContext.cs
Gnol-VN: Because we need to register this in Logging/ServiceCollectionExtensions.cs which is in a different project.

src/Hosting/Certificates/InitializationCertificateReader.cs
Gnol-VN: Updated accordingly

src/Logging/ServiceCollectionExtensions.cs
```diff
@@ -50,14 +53,20 @@ public static IServiceCollection AddOmexLogging(this IServiceCollection serviceC
 			serviceCollection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
 			serviceCollection.TryAddTransient<IExternalScopeProvider, LoggerExternalScopeProvider>();
 
+			AddOmexLegacyLogging(serviceCollection);
```
artempushkin, line 56: What are you achieving my extracting code into a separate function?

src/Logging/Internal/EventSource/EmptyServiceContext.cs
K-Cully: So in summary, the change over to Open Telemetry is resulting in this the solution in this repo becoming incomplete.
If that is the case then why keep this repo alive at all and why not just move it to the internal repo that completes the solution?

src/Hosting.Services/HostBuilderExtensions.cs
K-Cully: Okay, the old logging approach shouldn't really be removed then. The current approach is making the solution in this repo incomplete.

src/Logging/Internal/EventSource/EmptyServiceContext.cs
artempushkin: Logging/ServiceCollectionExtensions.cs is not in a different project.

src/Logging/InitializationLogger.cs
```diff
@@ -56,6 +56,7 @@ public static void LogInitializationSucceed(string serviceNameForLogging, string
 		/// <param name="serviceNameForLogging">Service name for logging</param>
 		/// <param name="ex">Exception to log</param>
 		/// <param name="message">Message to log</param>
+		[Obsolete("InitializationLogger using OmexLogger is obsolete and is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
```
artempushkin, line 59: Is it not enough to mark the class as obsolete?

src/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -17,13 +18,15 @@ public static class ServiceCollectionExtensions
...
 		/// </summary>
+		[Obsolete($"Please do not use {nameof(AddOmexServices)} if you just want to use Legacy OmexLogger and ActivityEventSender because they are deprecated. Consider using a different telemetry solution. This method is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
 		public static IHostBuilder AddOmexServices(this IHostBuilder builder) =>
 			builder
 				.ConfigureServices((context, collection) => collection.AddOmexServices());
 
 		/// <summary>
 		/// Add Omex Logging and ActivitySource dependencies
 		/// </summary>
+		[Obsolete($"Please do not use {nameof(AddOmexServices)} if you just want to use Legacy OmexLogger and ActivityEventSender because they are deprecated. Consider using a different telemetry solution. This method is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
```
artempushkin, line 29: Is OmexActivitySource getting deprecated?

src/Hosting/Certificates/InitializationCertificateReader.cs
```diff
@@ -11,6 +12,7 @@ namespace Microsoft.Omex.Extensions.Hosting.Certificates
 	/// <remarks>
 	/// Please prefer using instance resolved from DI container.
 	/// </remarks>
+	[Obsolete("InitializationLogger using OmexLogger is obsolete and is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
```
artempushkin, line 15: What is going to replace this class?

src/Hosting.Services/HostBuilderExtensions.cs
artempushkin: Repeated calls to TryAddXXX will not override the first registration.

src/Hosting.Services/HostBuilderExtensions.cs
artempushkin: Looks like logging components are coupled with other concerns. It would be better to split them. OTel configuration should not be integrated in this repo, consumers should be free to choose their own telemetry implementation.

artempushkin (changes requested):

src/Logging/Internal/EventSource/EmptyServiceContext.cs
Gnol-VN: > Logging/ServiceCollectionExtensions.cs is not in a different project.

I mean making EmptyServiceContext public so that we can register EmptyServiceContext in `Hosting.Services\HostBuilderExtensions.cs`

Gnol-VN: > So in summary, the change over to Open Telemetry is resulting in this the solution in this repo becoming incomplete. If that is the case then why keep this repo alive at all and why not just move it to the internal repo that completes the solution?

In future, we want to have another OmexOpenTelemetryLogger opensource as discussed with the team

src/Hosting/ServiceCollectionExtensions.cs
Gnol-VN: This source is deprecated [EventSource(Name = "Microsoft-OMEX-TimedScopes-Ext")]

src/Hosting/Certificates/InitializationCertificateReader.cs
Gnol-VN: Nothing for now. We can get this supported by OmexOpenTelemetry in the future. But for now it is obsolete. 
A workaround / renovation can come in another PR

src/Logging/ServiceCollectionExtensions.cs
Gnol-VN: I know some consumers are using this AddOmexLogging directly. I make this obsolete to explicitly remind them not to use this method.
If they do not use this method, there is still a need for EmptyServiceContext, BaseExecutionContext, LoggerExternalScopeProvider. That is why I copy them to [HostBuilderExtensions.cs](https://github.com/microsoft/Omex/pull/673/files#diff-4bf858f6b9cd29174b972054a6df039db4dcb0c3f43c4316e5e3e19eff5b7f2d)

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: see my explanation here: https://github.com/microsoft/Omex/pull/673/files#r1598767613 in this file src/Logging/ServiceCollectionExtensions.cs. 
If consumers do not call `AddOmexLogging()`, they will lose EmptyServiceContext, BaseExecutionContext, LoggerExternalScopeProvider. 
So this is the place to keep them so as not to break other code.

src/Logging/InitializationLogger.cs
Gnol-VN: removed

src/ServiceFabricGuest.Abstractions/ServiceFabricClientFactory.cs
```diff
@@ -12,6 +13,7 @@ namespace Microsoft.Omex.Extensions.ServiceFabricGuest.Abstractions
 	/// <remarks>
 	/// Please prefer using instance resolved from DI container.
 	/// </remarks>
+	[Obsolete("InitializationLogger using OmexLogger is obsolete and is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
```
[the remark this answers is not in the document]
Gnol-VN: resolved

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: This can be addressed in another PR for rearranging components and librarys

src/Hosting.Services/HostBuilderExtensions.cs
artempushkin: This code in its current state doesn't make sense. You have to figure out the scenarios where EmptyServiceContext and BaseExecutionContext get used instead of OmexServiceFabricContext and ServiceFabricExecutionContext, and make sure those scenarios continue working.

OR

Be less aggressive with obsoletion, mark just those classes and methods which you 100% know will get deleted, and leave the rest untouched.

src/Hosting/Certificates/InitializationCertificateReader.cs
artempushkin: OTel does not provide a certificate reader class AFAIK.

src/Logging/ServiceCollectionExtensions.cs
artempushkin: I don't get it. You extracted some code into AddOmexLegacyLogging which is a private method and cannot be called from anywhere else. Why?

Gnol-VN: Oh, I see the confusion. I revert and remove ~AddOmexLegacyLogging~. 
1. The goal 1 here is to mark AddOmexLogging [Obsolete], and delete it in the future.

2. Goal 2 is to still keep EmptyServiceContext, BaseExecutionContext, and LoggerExternalScopeProvider registered. For goal 2, I copied them to  `AddOmexServiceFabricDependencies()`.  This is a downstream place to catch-all scenario when consumers follow these call orders:
a. BuildStatelessService -> BuildServiceFabricService -> AddOmexServiceFabricDependencies
b. BuildStatefulService -> BuildServiceFabricService -> AddOmexServiceFabricDependencies

For Omex consumers, we all called BuildStatelessService / BuildStatefulService  which in turns call AddOmexServiceFabricDependencies and will get EmptyServiceContext, BaseExecutionContext, and LoggerExternalScopeProvider registered

src/Hosting/Certificates/InitializationCertificateReader.cs
Gnol-VN: Yes, but we can have OpenTelemetryInitializationLogger here to build this InitializationCertificateReader

artempushkin: If you are not going to delete this class, don't mark it obsolete.

src/Logging/ServiceCollectionExtensions.cs
Gnol-VN: The both EmptyServiceContext  and OmexServiceFabricContext both implement IServiceContext.
EmptyServiceContext is used when consumers only use AddOmexLogging() but dont use BuildServiceFabricService.
OmexServiceFabricContext is used when  consumers use BuildServiceFabricService.

Gnol-VN: Therefore, EmptyServiceContext and LoggerExternalScopeProvider dont need to be copied to AddOmexServiceFabricDependencies

src/Hosting/Certificates/InitializationCertificateReader.cs
Gnol-VN: Yes, obsolete removed. Used Pragma instead

src/Hosting/ServiceCollectionExtensions.cs
artempushkin: I'm asking about ActivitySource registered from .AddOmexActivitySource(). It looks like it will stay with some changes, so you can't obsolete this method.

Gnol-VN: oh okay. I use Pragma instead

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -60,7 +61,14 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
...
 			collection.TryAddSingleton<IServiceContext, OmexServiceFabricContext>();
 			collection.TryAddSingleton<IExecutionContext, ServiceFabricExecutionContext>();
+
+			collection.AddLogging();
+
+			collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
+
+#pragma warning disable OMEX188 // This method also adds Legacy OmexLogger and ActivityEventSender which are deprecated. They are pending for removal by 1 July 2024. Consider adding a different telemetry solution. DiagnosticId = "OMEX188"
 			return collection.AddOmexServices();
+#pragma warning restore OMEX188 // This method also adds Legacy OmexLogger and ActivityEventSender which are deprecated. They are pending for removal by 1 July 2024. Consider adding a different telemetry solution. DiagnosticId = "OMEX188"
```
artempushkin, line 71: As explained in another comment, these changes are not needed, revert. The line ```collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();``` is not doing anything anyway, it will be simply ignored.

src/Logging/Internal/EventSource/EmptyServiceContext.cs
Gnol-VN: yes

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -101,16 +110,20 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
 						options.ValidateOnBuild = true;
 						options.ValidateScopes = true;
 					})
-					.ConfigureLogging(builder => builder.AddOmexLogging())
 					.Build();
```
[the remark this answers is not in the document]
Gnol-VN: Isnt it redundant? 
AddOmexServiceFabricDependencies() will call AddOmexServices() will call AddOmexLogging()

src/Logging/InitializationLogger.cs
```diff
@@ -12,6 +11,7 @@ namespace Microsoft.Omex.Extensions.Logging
 	/// InitializationLogger is the logger to be used before the proper ILogger from DI is set.
 	/// Not to be used as main logger.
 	/// </summary>
+	[Obsolete($"{nameof(InitializationLogger)} using {nameof(OmexLogger)} is obsolete and is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
 	public static class InitializationLogger
```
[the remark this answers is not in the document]
Gnol-VN: right

src/Logging/ServiceCollectionExtensions.cs
Gnol-VN: Kudo. I see your point. If they call AddOmexLogging first, the IServiceContext will be resolved into EmptyServiceContext.  
Subsequently, if they call any of BuildStatelessService or BuildServiceFabricService or AddOmexServiceFabricDependencies, IServiceContext will **not** be resolved in OmexServiceFabricContext.

In other words, if Omex consumers are explicitly use AddOmexLogging **and** BuildStatelessService() , they will lose SFContext. 

Is my understanding correct?

By the way, this is very good to spot Artem. It should be discussed out of this PR.

tests/Hosting.Services.UnitTests/HostBuilderExtensionsTests.cs
```diff
@@ -29,8 +29,10 @@ public void AddOmexServiceFabricDependencies_TypesRegistered(Type typeToResolve,
 		{
 			void CheckTypeRegistration<TContext>() where TContext : ServiceContext
 			{
+#pragma warning disable CS0618 // AddOmexServiceFabricDependencies also adds Legacy OmexLogger and ActivityEventSender and they are deprecated. This legacy telemetry is pending for removal by 1 July 2024. Consider adding a different telemetry solution. DiagnosticId = "OMEX188"
```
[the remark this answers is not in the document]
Gnol-VN: done

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: done

src/ServiceFabricGuest.Abstractions/ServiceFabricClientFactory.cs
```diff
@@ -18,6 +19,8 @@ public class ServiceFabricClientFactory
 		/// Creates service fabric client wrapper instance and returns it.
 		/// </summary>
 		public static IServiceFabricClientWrapper Create(ServiceFabricRestClientOptions options) =>
+#pragma warning disable OMEX188 // InitializationCertificateReader using OmexLogger is obsolete. DiagnosticId = "OMEX188"
```
K-Cully, line 22: This is incorrect.
ServiceFabricClientFactory is required.
InitializationCertificateReader should be updated to use the OTelInitialisationLogger Instance instead.
*at this point, and in this case, it's better to break the old and fix the new.

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -101,16 +103,20 @@ public static IServiceCollection AddOmexServiceFabricDependencies<TContext>(this
...
 
+#pragma warning disable OMEX188 // InitializationLogger using OmexLogger is obsolete. DiagnosticId = "OMEX188"
 				InitializationLogger.LogInitializationSucceed(serviceNameForLogging);
```
K-Cully, line 109: If AddOmexLogging is no longer called, how will this InitializationLogger work?

src/ServiceFabricGuest.Abstractions/ServiceFabricClientFactory.cs
Gnol-VN: This is pragma to mark InitializationCertificateReader.Instance, not [Obsolete] to ServiceFabricCLientWrapper, so it does not really affect consumer.
OTELInitializationLogger should be available in future PR

src/Hosting.Services/HostBuilderExtensions.cs
Gnol-VN: I dont fully get your question. however InitializationLogger is a static class/method which is isolated from DI using AddOmexLogging

src/Logging/InitializationLogger.cs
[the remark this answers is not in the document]
artempushkin: You have not obsoleted it though. ServiceInitializationEventSource.

src/Hosting.Services/HostBuilderExtensions.cs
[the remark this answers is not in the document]
artempushkin: It may be redundant, but it's not obvious. There may also be functional differences, e.g. you are removing a call to ILoggingBuilder.AddConfiguration. Besides, this change is out of scope for obsoletion task.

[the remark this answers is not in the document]
Gnol-VN: Reverted

artempushkin (approved):

PR #697, opened 2024-10-10
Update Service Fabric to Service Fabric 10.1 Cumulative Update 4.0

neilr81:
Release notes:
https://github.com/microsoft/service-fabric/blob/master/release_notes/Service_Fabric_ReleaseNotes_101CU4.md

PR #698, opened 2024-10-24
Revert "Update Service Fabric to Service Fabric 10.1 Cumulative Update 4.0"

neilr81:
Reverts microsoft/Omex#697


neilr81:
Adding comment to avoid completion

marc-hayek (approved):

PR #705, opened 2024-11-29
Pinned major version for framework packages

artempushkin:
Pinning the packages which comprise a part of .NET. This way we have to manually update the major version once when moving to .NET vNext, but we don't have to downgrade the packages in every package update PR between vNext release and our upgrade.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -64,8 +64,7 @@
     <PackageVersion Include="Microsoft.ServiceFabric.Services.Remoting" Version="7.1.2448" />
     <PackageVersion Include="ServiceFabric.Mocks" Version="7.2.8" />
   </ItemGroup>
-  <ItemGroup Label="Moq">
-    <!-- Pinning moq due to [SponsorLink privacy concerns](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/moq-incident-and-remediation) -->
+  <ItemGroup Label="Moq pinned to 4.18.4 until SponsorLink is fully removed and new versions are deemed trustworty">
     <PackageVersion Include="Moq" Version="4.18.4" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -5,25 +5,23 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
   </PropertyGroup>
   <ItemGroup Label="Latest DotNet Package Versions. AutoUpdate" Condition="'$(TargetFramework)' == '$(LatestSupportedDotNetVersion)' OR '$(IsNetStandard)'">
```
K-Cully, line 7: latest shouldn't really have PreserveMajor defined if we want to follow an automated stay-current model.
Otherwise every major version bump will require manually curating every package version.

Directory.Packages.props
```diff
@@ -66,7 +64,8 @@
     <PackageVersion Include="Microsoft.ServiceFabric.Services.Remoting" Version="7.1.2448" />
     <PackageVersion Include="ServiceFabric.Mocks" Version="7.2.8" />
   </ItemGroup>
-  <ItemGroup Label="Moq pinned to 4.18.4 until SponsorLink is fully removed and new versions are deemed trustworty">
+  <ItemGroup Label="Moq">
+    <!-- Pinning moq due to [SponsorLink privacy concerns](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/moq-incident-and-remediation) -->
```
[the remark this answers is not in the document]
artempushkin: Good point, reverted to the original state.

Directory.Packages.props
artempushkin: I'd prefer manually updating once a year after vNext release over manually downgrading each package update PR between release and adoption.

artempushkin:
I'll defer this change for now.

PR #713, opened 2025-01-22
Remove omexlogger 

Gnol-VN:
This PR is to:
1. Remove OmexLogger and its relatives
2. Add an action to support InitializationLogger, so consumers can supply their own logger to know if the host is built successfully or not
3. Remove [Obsolete] attributes

## Why did I add an action to configure the under-the-hood InitializationLogger?
The `BuildState(less/ful)Service()` will call `BuildServiceFabricService()`, and this is the only place we do logging just before host is really built. 

To receive Initialization logs, the last guard is having InitializationLogger in consumer's `Program.cs`. However, this has to ensure all consumers use 
> try { } catch { OTELInitLogger.log(exception) }

I idealize, instead of assuming consumer having this try catch pattern in all of their Program.cs files, this action should be configured in a Shared library, or Shared SDK.  For handy experience, consumers within an organisation just need to import the shared library and they do not need to do the try/catch above to get logs from InitializationLogger

### DIFF
--- a/src/Abstractions/EventSources/EventSourcesEventIds.cs
+++ b/src/Abstractions/EventSources/EventSourcesEventIds.cs
@@ -12,31 +12,6 @@ namespace Microsoft.Omex.Extensions.Abstractions.EventSources
 	public enum EventSourcesEventIds
 	{
 		/// <summary>
-		/// Event Id for logging general info message
-		/// </summary>
-		LogInfo = 1,
-
-		/// <summary>
-		/// Event Id for logging error message
-		/// </summary>
-		LogError = 2,
-
-		/// <summary>
-		/// Event Id for logging warning message
-		/// </summary>
-		LogWarning = 3,
-
-		/// <summary>
-		/// Event Id for logging verbose message
-		/// </summary>
-		LogVerbose = 4,
-
-		/// <summary>
-		/// Event Id for logging spam message
-		/// </summary>
-		LogSpam = 5,
-
-		/// <summary>
 		/// Event Id for logging activities
 		/// </summary>
 		LogActivity = 6,
@@ -45,40 +20,5 @@ namespace Microsoft.Omex.Extensions.Abstractions.EventSources
 		/// Event Id for logging test activities
 		/// </summary>
 		LogActivityTestContext = 7,
-
-		/// <summary>
-		/// Event Id for service type registered
-		/// </summary>
-		ServiceTypeRegistered = 11,
-
-		/// <summary>
-		/// Event Id for service host initialization failed
-		/// </summary>
-		ServiceHostInitializationFailed = 12,
-
-		/// <summary>
-		/// Event Id for actor type registered
-		/// </summary>
-		ActorTypeRegistered = 13,
-
-		/// <summary>
-		/// Event Id for actor host initialization failed
-		/// </summary>
-		ActorHostInitializationFailed = 14,
-
-		/// <summary>
-		/// Event Id for logging Analytics information
-		/// </summary>
-		LogAnalytics = 15,
-
-		/// <summary>
-		/// Event Id for generic host build succeded
-		/// </summary>
-		GenericHostBuildSucceeded = 16,
-
-		/// <summary>
-		/// Event Id for generic host build failed
-		/// </summary>
-		GenericHostFailed = 17
 	}
 }


artempushkin (changes requested):
Blocking to get time for proper review.

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -97,29 +99,24 @@ private static IHost BuildServiceFabricService<TRunner, TService, TContext>(
 							.AddOmexServiceFabricDependencies<TContext>()
 							.AddSingleton<IOmexServiceRegistrator, TRunner>()
 							.AddHostedService<OmexHostedService>();
+						collection.TryAddTransient<IServiceContext, EmptyServiceContext>();
+						collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
```
K-Cully, line 103: suggest adding a comment before these to indicate why they are added via TryAddTransient
```suggestion
						// If specific context constructs are not registered, add defaults for telemetry purposes
						collection.TryAddTransient<IServiceContext, EmptyServiceContext>();
						collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
```

src/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -26,11 +24,7 @@ public static IHostBuilder AddOmexServices(this IHostBuilder builder) =>
 		/// Add Omex Logging and ActivitySource dependencies
 		/// </summary>
 		public static IServiceCollection AddOmexServices(this IServiceCollection collection) =>
-#pragma warning disable OMEX188 // OmexLogger and OmexLogEventSource are obsolete and pending for removal by 1 July 2024. Please consider using a different Logger.
-			collection
-				.AddOmexLogging()
-#pragma warning restore OMEX188 // OmexLogger and OmexLogEventSource are obsolete and pending for removal by 1 July 2024. Please consider using a different Logger.
-				.AddOmexActivitySource();
+			collection.AddOmexActivitySource();
```
K-Cully, line 27: Why is the Activity Source path still supported?

src/Logging/Internal/EventSource/ServiceInitializationEventSource.cs
```diff
@@ -11,7 +11,6 @@ namespace Microsoft.Omex.Extensions.Logging
 	/// Service Fabric event source
 	/// </summary>
 	[EventSource(Name = "Microsoft-OMEX-HostLogs")]
-	[Obsolete($"{nameof(ServiceInitializationEventSource)} is obsolete and is pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
 	internal sealed class ServiceInitializationEventSource : EventSource
```
K-Cully, line 14: I take it the decision was to accept keeping EventSource use for service initialisation only. Is that correct?

artempushkin:
Did you forget to actually remove OmexLogger?

src/Abstractions/ServiceContext/IServiceContext.cs
```diff
@@ -3,7 +3,7 @@
```
artempushkin, line 3: This interface is practically unused, the only reason to keep it would be for the sake of completeness. Most of the context is exposed via IExecutionContext which pretends to be platform-agnostic, the 2 fields here are SF-specific and also not very useful. If really needed, they can be accessed via ```IAccessor<ServiceContext>```.

src/Logging/InitializationLogger.cs
```diff
@@ -11,28 +11,33 @@ namespace Microsoft.Omex.Extensions.Logging
...
-		 }).CreateLogger("Initial-Logging");
-
-		private static ILoggingBuilder LoadInitializationLogger(this ILoggingBuilder builder)
+		public static ILogger Instance
 		{
-			if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") == "Development")
-			{
-				builder.AddConsole();
-			}
+			get =>
```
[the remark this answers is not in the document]
artempushkin: You are creating (and not disposing) a new ILoggerFactory and ILogger each time Instance is accessed.

src/Logging/InitializationLogger.cs
```diff
@@ -11,28 +11,33 @@ namespace Microsoft.Omex.Extensions.Logging
...
-			return builder;
+					s_loggingBuilderAction?.Invoke(builder);
+				}).CreateLogger("Initial-Logging");
 		}
 
+		/// <summary>
+		/// Action to customize the initialization logger builder
+		/// </summary>
+		/// <param name="action"></param>
+		public static void CustomizeInitializationLoggerBuilder(Action<ILoggingBuilder> action) => s_loggingBuilderAction = action;
```
artempushkin, line 37: It's not making life easier for consumers. Instead of wrapping host initialization in try/catch and explicitly logging failures, you want them to call this hook to customize the logger. And then they still have to add try/catch and logging to capture exceptions from IHost.Run[Async].

artempushkin (changes requested):

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -75,53 +77,39 @@ private static IHost BuildServiceFabricService<TRunner, TService, TContext>(
...
+						.Configure<ServiceRegistratorOptions>(options =>
+						{
+							options.ServiceTypeName = serviceName;
+						})
+						.AddOmexServiceFabricDependencies<TContext>()
+						.AddSingleton<IOmexServiceRegistrator, TRunner>()
+						.AddHostedService<OmexHostedService>();
+					// If specific context constructs are not registered, add defaults for telemetry purposes
+					collection.TryAddTransient<IServiceContext, EmptyServiceContext>();
+					collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
```
artempushkin, line 102: Both interfaces already registered in .AddOmexServiceFabricDependencies()

src/Abstractions/ServiceContext/EmptyServiceContext.cs
```diff
@@ -3,15 +3,21 @@
...
 using System;
 
-namespace Microsoft.Omex.Extensions.Logging
+namespace Microsoft.Omex.Extensions.Abstractions.ServiceContext
 {
 	/// <summary>
 	/// IServiceContext without any information
 	/// </summary>
-	internal class EmptyServiceContext : IServiceContext
+	public class EmptyServiceContext : IServiceContext
```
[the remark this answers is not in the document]
artempushkin: This class was only registered from within .AddOmexLogging, now it's not in use, see my comment in the new registration place.

EmptyServiceContext and BaseExecutionContext were meant for non-SF or platform agnostic services, but registration was not implemented properly and due to spurious use of .AddOmexLogging they often ended up preventing SF-specific implementations from being registered.

src/Abstractions/ServiceContext/IServiceContext.cs
artempushkin: If you want to keep the interface and both implementations, registration logic should be fixed.

I would either leave them where they are now or just delete.

src/Hosting.Services/HostBuilderExtensions.cs
```diff
@@ -75,53 +71,37 @@ private static IHost BuildServiceFabricService<TRunner, TService, TContext>(
...
+					collection
+						.Configure<ServiceRegistratorOptions>(options =>
+						{
+							options.ServiceTypeName = serviceName;
+						})
+						.AddOmexServiceFabricDependencies<TContext>()
+						.AddSingleton<IOmexServiceRegistrator, TRunner>()
+						.AddHostedService<OmexHostedService>();
+					// If specific context constructs are not registered, add defaults for telemetry purposes
+					collection.TryAddTransient<IExecutionContext, BaseExecutionContext>();
```
artempushkin, line 95: IExecutionContext is already registered in .AddOmexServiceFabricDependencies

src/Hosting.Services/HostBuilderExtensions.cs
artempushkin: I suppose you later noticed that .AddOmexServiceFabricDependencies is called a couple of lines higher.

src/Logging/Internal/EventSource/OmexLogEventSource.cs
```diff
@@ -1,134 +0,0 @@
...
-using Microsoft.Omex.Extensions.Abstractions.EventSources;
-
-namespace Microsoft.Omex.Extensions.Logging
-{
-	// Renamed from Microsoft-OMEX-Logs to avoid conflict with sources in other libraries
-	[Obsolete($"{nameof(OmexLogEventSource)} is obsolete and pending for removal by 1 July 2024.", DiagnosticId = "OMEX188")]
-	[EventSource(Name = "Microsoft-OMEX-Logs-Ext")]
-	internal sealed class OmexLogEventSource : EventSource
-	{
-		[Event((int)EventSourcesEventIds.LogError, Level = EventLevel.Error, Message = "{12}:{13} {16}", Version = 7)]
```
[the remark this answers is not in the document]
artempushkin: Some event ids from EventSourcesEventIds enum are no longer in use after deleting the event sources. It would be nice to clean them up. You can't delete the entire enum, other event ids are still in use I think.

src/Logging/Internal/EventSource/OmexLogEventSender.cs
```diff
@@ -1,126 +0,0 @@
...
-			long replicaId = m_serviceContext.ReplicaOrInstanceId;
-			string applicationName = m_executionContext.ApplicationName;
-			string serviceName = m_executionContext.ServiceName;
-			string buildVersion = m_executionContext.BuildVersion;
-			string machineId = m_executionContext.MachineId;
-
-			string tagName = eventId.Name ?? string.Empty;
-			// In case if tag created using Tag.Create (line number and file in description) it's better to display decimal number
-			string tagId = string.IsNullOrWhiteSpace(eventId.Name)
-				? eventId.ToTagId()
```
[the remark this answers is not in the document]
artempushkin: why another PR? it's related code

artempushkin (approved):

PR #724, opened 2025-04-01
Rifatkale/feature/replica role check

rftkl1:
Make replica role available to health checks

### DIFF
--- a/src/Hosting.Services/Internal/OmexStatefulServiceRegistrator.cs
+++ b/src/Hosting.Services/Internal/OmexStatefulServiceRegistrator.cs
@@ -19,12 +19,14 @@ namespace Microsoft.Omex.Extensions.Hosting.Services
 			IAccessorSetter<StatefulServiceContext> contextAccessor,
 			IAccessorSetter<IStatefulServicePartition> partitionAccessor,
 			IAccessorSetter<IReliableStateManager> stateAccessor,
+			IAccessorSetter<ReplicaRoleWrapper> roleAccessor,
 			IEnumerable<IListenerBuilder<OmexStatefulService>> listenerBuilders,
 			IEnumerable<IServiceAction<OmexStatefulService>> serviceActions)
 				: base(options, contextAccessor, listenerBuilders, serviceActions)
 		{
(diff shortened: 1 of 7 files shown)


src/Hosting.Services/Internal/OmexStatefulServiceRegistrator.cs
```diff
@@ -33,5 +35,14 @@ public override Task RegisterAsync(CancellationToken cancellationToken) =>
...
 		public IAccessorSetter<IStatefulServicePartition> PartitionAccessor { get; }
+
+		public IAccessorSetter<ReplicaRoleWrapper> RoleAccessor { get; }
+
+		// 'ReplicaRole' must be a reference type in order to use it as parameter 'TValue' in the generic type or method. That's why we use this wrapper class
+		public class ReplicaRoleWrapper
+		{
+			public ReplicaRole Role { get; set; }
+		}
+
```
artempushkin, line 46: We usually keep classes in their own files, unless there's a strong reason to make it nested.

src/Hosting.Services/OmexStatefulService.cs
```diff
@@ -6,41 +6,57 @@
...
+            m_serviceRegistrator.ListenerBuilders.Select(b => new ServiceReplicaListener(c => b.Build(this), b.Name));
+
+        /// <inheritdoc />
+        protected override Task RunAsync(CancellationToken cancellationToken) =>
+            Task.WhenAll(m_serviceRegistrator.ServiceActions.Select(r => r.RunAsync(this, cancellationToken)));
+
+        /// <summary>
+        /// Gets the current replica role.
+        /// </summary>
+        public ReplicaRole GetCurrentReplicaRole() => m_replicaRoleAccessor.Value?.Role ?? ReplicaRole.Unknown;
```
artempushkin, line 60: I don't think this is useful, consumers generally don't access this class directly.

src/Hosting.Services/Internal/OmexStatefulServiceRegistrator.cs
rftkl1: fixed

PR #728, opened 2025-04-29
Rifatkale/feature/replica role check

rftkl1:
image


tests/Hosting.Services.UnitTests/ServiceFabricHostBuilderTests.cs
```diff
@@ -162,6 +167,32 @@ public void ConfigureServices_PropagatesTypesRegistration()
...
 		}
 
+		[TestMethod]
+		public void OmexStatefulService_Constructor_SetsOmexStateManagerWithUnknownRole()
+		{
+			StatefulServiceContext serviceContext = MockStatefulServiceContextFactory.Default;
+
+			Mock<IAccessorSetter<StatefulServiceContext>> contextAccessorMock = new();
+			Mock<IAccessorSetter<IReliableStateManager>> stateAccessorMock = new();
+			Mock<IAccessorSetter<OmexStateManager>> stateManagerAccessorMock = new();
```
artempushkin, line 177: No need to mock where you can simply use a class.
https://github.com/microsoft/Omex/blob/main/src/Abstractions/Accessors/Accessor.cs

tests/Hosting.Services.UnitTests/ServiceFabricHostBuilderTests.cs
```diff
@@ -162,6 +167,32 @@ public void ConfigureServices_PropagatesTypesRegistration()
...
+
+			OmexStatefulServiceRegistrator serviceRegistrator = new OmexStatefulServiceRegistrator(
+				Mock.Of<IOptions<ServiceRegistratorOptions>>(),
```
artempushkin, line 180: For IOptions you can use Options.Create instead of mocking.

rftkl1: @artempushkin fixed

tests/Hosting.Services.UnitTests/ServiceFabricHostBuilderTests.cs
rftkl1: @artempushkin  fixed

